PR #20, opened 2022-06-15
Code review, minor suggestions/comment changes, questions. Looks great!

toddbaert:
I did a `git reset {sha}` back to the sha from my last review, then added comments, etc and then did a `git rebase main -X theirs` which brought back all your commits and took mine for all conflicts, leaving us with just my diff!

Anyway...

This looks really good to me, I like the choices you made (the changes haven't been massive compared to my last review) but I can't see anything that looks like a problem to me, and we appear to have made the same decisions in most places.

### DIFF
--- a/lib/src/main/java/dev/openfeature/javasdk/BaseEvaluation.java
+++ b/lib/src/main/java/dev/openfeature/javasdk/BaseEvaluation.java
@@ -1,7 +1,7 @@
 package dev.openfeature.javasdk;
 
 /**
- * We differ between the evaluation results that providers return and what is given to the end users. This is a common interface between them.
+ * This is the common interface between the evaluation results that providers return and what is given to the end users.
  * @param <T> The type of flag being evaluated.
  */
 public interface BaseEvaluation<T> {
--- a/lib/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
+++ b/lib/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
@@ -1,6 +1,6 @@
 package dev.openfeature.javasdk;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectMapper; // jackson is pretty standard, I'm glad you decided on this for object support
 import lombok.*;
 
 import java.time.ZonedDateTime;
@@ -28,6 +28,8 @@ public class EvaluationContext {
     }
 
     // TODO Not sure if I should have sneakythrows or checked exceptions here..
+    // do you mean un-checked exceptions?
+    // I had no idea about SneakyThrows, but it seems like a nicer solution than swallowing in an unchecked exception. 
     @SneakyThrows
     public <T> void addStructureAttribute(String key, T value) {
         jsonAttributes.put(key, objMapper.writeValueAsString(value));
--- a/lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
+++ b/lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
@@ -84,6 +84,7 @@ public class OpenFeatureClient implements Client {
             }
             details.value = defaultValue;
             details.reason = Reason.ERROR;
+            // We may want to extend exception and add a defined code field, instead of just using message.
             details.errorCode = e.getMessage();
             this.errorHooks(hookCtx, e, mergedHooks, hints);
         } finally {
@@ -93,6 +94,7 @@ public class OpenFeatureClient implements Client {
         return details;
     }
 
+    // errorHooks and afterAllHooks error handling seems consistent with spec, and what I implemented. I'm glad we landed here.
     private void errorHooks(HookContext hookCtx, Exception e, List<Hook> hooks, ImmutableMap<String, Object> hints) {
         for (Hook hook : hooks) {
             try {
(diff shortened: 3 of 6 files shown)


lib/src/main/java/dev/openfeature/javasdk/BaseEvaluation.java
```diff
@@ -1,7 +1,7 @@
 package dev.openfeature.javasdk;
 
 /**
- * We differ between the evaluation results that providers return and what is given to the end users. This is a common interface between them.
+ * This is the common interface between the evaluation results that providers return and what is given to the end users.
```
toddbaert, line 4: I try to avoid using "I/we" in favor of passive voice in comments, but this is a nitpick.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -84,6 +84,7 @@ <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key, T defa
             }
             details.value = defaultValue;
             details.reason = Reason.ERROR;
+            // We may want to extend exception and add a defined code field, instead of just using message.
```
toddbaert, line 87: This is the only thing I think might need changing. Should we have a code field in a custom exception that's has it's own purpose beyond the general message?

I know this isn't in the spec, but maybe it should be...

cc @beeme1mr

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -122,6 +124,7 @@ private <T> void afterHooks(HookContext hookContext, FlagEvaluationDetails<T> de
     private EvaluationContext beforeHooks(HookContext hookCtx, List<Hook> hooks, ImmutableMap<String, Object> hints) {
         // These traverse backwards from normal.
         EvaluationContext ctx = hookCtx.getCtx();
+        // This is a bit counter-intuitive to me, just want to make sure... this is the stage that gets reversed because ArrayList.addAll() does a prepend, right?
```
justinabrahms, line 127: Before order is: 1,2,3,4
after,error,finally order is: 4,3,2,1

So we maintain 4,3,2,1 order and reverse for before.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
justinabrahms: What would we hope to accomplish? The `code` is a useful tool for devs for debugging. We have strings which are useful to understand what went wrong. That's an exception message.

Can you share an example so I can better understand?

lib/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -28,6 +28,8 @@ public class EvaluationContext {
     }
 
     // TODO Not sure if I should have sneakythrows or checked exceptions here..
+    // do you mean un-checked exceptions?
```
justinabrahms, line 31: This was a TODO for https://github.com/open-feature/spec/issues/88

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
toddbaert: Right, I understand that, it's just that reading this, it seems that before is running in reverse (4,3,2,1), you are reversing the ArrayList. I'm just confirming that this is because the arraylist already exists in order (4,3,2,1) because items are PRE-pended with addAll(), right?

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
toddbaert: I only mean that a dev might not assume the `message` gets put into the `errorCode` field. If we define a custom `OpenFeatureException`, with a `code` prop, we can say that the value of the code field will be put into the `errorCode` field in the `EvaluationDetails`.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
justinabrahms: Screen Shot 2022-06-15 at 8 51 14 PM
They aren't prepended via addAll(), but they are prepended via this above.
jdoodle.com/ia/sfr is an example of what addAll does, if it helps.

--

It sounds like you might have a really good test case in your head. Mind writing that out and I'll add it to the suite?

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
justinabrahms: So if folks don't throw our special exception type, are you thinking that we just leave the error code empty?

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
toddbaert: I think I'm doing a bad job communicating about this. `addAll()` does seem to "pre-prend" if I understand the [doc](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/util/List.html#addAll(int,java.util.Collection)) right.

> Shifts the element currently at that position (if any) and any subsequent elements to the right (increases their indices).

The case I'm wondering about is if you did:

```
client.addHooks(before1, before2)
```
before1 will run before2. As the code is written, it seem like this will happen because of the way `addAll` works.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
justinabrahms: I think I'd like to leave this as is. Giving the original error provides something for the dev to work against when it comes to debugging. Just getting a "generic error" string back would be quite frustrating when something went wrong.

toddbaert:
Closing this, great work from my perspective, and looks consistent with the node implementation. I might write a specific test around the case I mentioned after confirming it isn't covered, but I think it will pass anyway if I'm reading the code correctly.

Issue #16, opened 2022-06-14
Help Wanted: Tests for logging to validate 1.4.10


justinabrahms:
I would love help testing the logging around requirement 1.4.10 (see link below). I've made a few attempts at it, but I can't quite get the slf4j incantation correct to mock the relevant pieces.

https://github.com/open-feature/java-sdk/blob/main/lib/src/test/java/dev/openfeature/javasdk/FlagEvaluationSpecTests.java#L163

justinabrahms:
Fixed by #22

PR #22, opened 2022-06-17
[16] Tests for logging to validate 1.4.10 (#1)

lopitz:
This PR addresses #16 and can be used as a blueprint for other tests asserting log messages.

-removes dependency to log4j binding for slf4j, so clients can choose a different binding
-adds dependency to slf4j-api
-adds test logging binding by adding (test) dependency to uk.org.lidalia:slf4j-test
-adds (test) dependency to assertj for better readable assertions

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -26,8 +26,8 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


justinabrahms (approved):
Love it. Thanks, Lars!

Issue #23, opened 2022-06-20
Hook type issue


justinabrahms:
@lopitz found an issue whereby hooks could fail silently due to a ClassCastException if the whole chain (hooks & values) isn't all of the same / right type.

lopitz:
Due to the nature of this issue, there are two options to mitigate it. 
1. call only these hooks, which support the type of the evaluated flag
2. call all hooks, but let the hooks be untyped, leaving it up to the hook author to check the type and apply logic.

In my PR I opted for the first option, but with some sort of expressiveness by introducing dedicated interfaces for boolean, string and integer typed hooks. If hook authors still wants their hook to get called everytime, they still can use the generic form.

PR #24, opened 2022-06-20
Fixes ClassCastExceptions when hooks are called (address #23)

lopitz:
Due to the nature of this issue, there are two options to mitigate it.

1. call only these hooks, which support the type of the evaluated flag
2. call all hooks, but let the hooks be untyped, leaving it up to the hook author to check the type and apply logic.

In my PR I opted for the first option, but with some sort of expressiveness by introducing dedicated interfaces for boolean, string and integer typed hooks. If hook authors still wants their hook to get called everytime, they still can use the generic form.

-fixes ClassCastException by only calling hooks, that support the flag value
-introduces specialized hook interfaces for supported types
-adds a fixture for hooks for tests
-moves out calling of hooks from OpenFeatureClient to dedicated class HookSupport
-introduces method in Hook showing the supported flag value type
-makes FlagEvaluationOptions immutable

### DIFF
--- a/lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
+++ b/lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
@@ -1,0 +1,69 @@
+package dev.openfeature.javasdk;
+
+import java.util.*;
+import java.util.function.Consumer;
+
+import com.google.common.collect.Lists;
+import lombok.*;
+import org.slf4j.Logger;
+
+@RequiredArgsConstructor
+class HookSupport {
+
+    private final Logger log;
+
+    public void errorHooks(FlagValueType flagValueType, HookContext hookCtx, Exception e, List<Hook> hooks, Map<String, Object> hints) {
+        executeHooks(flagValueType, hooks, "error", hook -> hook.error(hookCtx, e, hints));
+    }
+
+    public void afterAllHooks(FlagValueType flagValueType, HookContext hookCtx, List<Hook> hooks, Map<String, Object> hints) {
+        executeHooks(flagValueType, hooks, "finally", hook -> hook.finallyAfter(hookCtx, hints));
+    }
+
+    public void afterHooks(FlagValueType flagValueType, HookContext hookContext, FlagEvaluationDetails details, List<Hook> hooks, Map<String, Object> hints) {
+        executeHooksUnsafe(flagValueType, hooks, hook -> hook.after(hookContext, details, hints));
+    }
+
+    private <T> void executeHooks(
+        FlagValueType flagValueType, List<Hook> hooks,
+        String hookMethod,
+        Consumer<Hook<T>> hookCode
+    ) {
+        hooks
+            .stream()
+            .filter(hook -> hook.supportsFlagValueType() == flagValueType)
+            .forEach(hook -> executeChecked(hook, hookCode, hookMethod));
+    }
+
+    private <T> void executeHooksUnsafe(
+        FlagValueType flagValueType, List<Hook> hooks,
+        Consumer<Hook<T>> hookCode
+    ) {
+        hooks
+            .stream()
+            .filter(hook -> hook.supportsFlagValueType() == flagValueType)
+            .forEach(hookCode::accept);
+    }
+
+    private <T> void executeChecked(Hook<T> hook, Consumer<Hook<T>> hookCode, String hookMethod) {
+        try {
+            hookCode.accept(hook);
+        } catch (Exception exception) {
+            log.error("Exception when running {} hooks {}", hookMethod, hook.getClass(), exception);
+        }
+    }
+
+    public EvaluationContext beforeHooks(HookContext hookCtx, List<Hook> hooks, Map<String, Object> hints) {
+        // These traverse backwards from normal.
+        EvaluationContext ctx = hookCtx.getCtx();
+        for (Hook hook : Lists.reverse(hooks)) {
+            Optional<EvaluationContext> newCtx = hook.before(hookCtx, hints);
+            if (newCtx != null && newCtx.isPresent()) {
+                ctx = EvaluationContext.merge(ctx, newCtx.get());
+                hookCtx = hookCtx.withCtx(ctx);
+            }
+        }
+        return ctx;
+    }
+
+}
--- a/lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
+++ b/lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
@@ -2,9 +2,10 @@ package dev.openfeature.javasdk;
 
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import com.google.common.collect.Lists;
-import lombok.*;
+import lombok.RequiredArgsConstructor;
 import org.slf4j.Logger;
 
 @RequiredArgsConstructor
@@ -12,18 +13,22 @@ class HookSupport {
 
     private final Logger log;
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public void errorHooks(FlagValueType flagValueType, HookContext hookCtx, Exception e, List<Hook> hooks, Map<String, Object> hints) {
         executeHooks(flagValueType, hooks, "error", hook -> hook.error(hookCtx, e, hints));
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public void afterAllHooks(FlagValueType flagValueType, HookContext hookCtx, List<Hook> hooks, Map<String, Object> hints) {
         executeHooks(flagValueType, hooks, "finally", hook -> hook.finallyAfter(hookCtx, hints));
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public void afterHooks(FlagValueType flagValueType, HookContext hookContext, FlagEvaluationDetails details, List<Hook> hooks, Map<String, Object> hints) {
-        executeHooksUnsafe(flagValueType, hooks, hook -> hook.after(hookContext, details, hints));
+        executeHooksUnchecked(flagValueType, hooks, hook -> hook.after(hookContext, details, hints));
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
     private <T> void executeHooks(
         FlagValueType flagValueType, List<Hook> hooks,
         String hookMethod,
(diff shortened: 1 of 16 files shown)


lib/src/test/java/dev/openfeature/javasdk/DoSomethingProvider.java
```diff
@@ -4,12 +4,7 @@ public class DoSomethingProvider implements FeatureProvider {
 
     @Override
     public Metadata getMetadata() {
-        return new Metadata() {
-            @Override
-            public String getName() {
-                return "test";
-            }
-        };
+        return () -> "test";
```
justinabrahms, line 7: I'm not familiar with this language feature. Do you know the name of it?

.github/workflows/merge.yml
```diff
@@ -22,10 +22,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up JDK 8
+    - name: Set up JDK 11
```
justinabrahms, line 25: We still need to ensure that this works for java 8 given internal workloads which haven't moved over. I think I might have also fat fingered the contrib docs on this recently, which I'll go check.

lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private <T> void executeHooks(
+        FlagValueType flagValueType, List<Hook> hooks,
+        String hookMethod,
+        Consumer<Hook<T>> hookCode
+    ) {
+        hooks
+            .stream()
+            .filter(hook -> hook.supportsFlagValueType() == flagValueType)
```
justinabrahms, line 36: Should `supportsFlagValueType` take a type as an argument? Seems plausible that a hook may want to support multiple.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -1,135 +1,98 @@
...
-    @Getter private final List<Hook> clientHooks;
     private static final Logger log = LoggerFactory.getLogger(OpenFeatureClient.class);
 
+    private final OpenFeatureAPI openfeatureApi;
+    @Getter
+    private final String name;
+    @Getter
+    private final String version;
+    @Getter
+    private final List<Hook> clientHooks;
```
justinabrahms, line 20: Please revert the formatting change to these lines.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -1,135 +1,98 @@
...
+    private final HookSupport hookSupport;
+
     public OpenFeatureClient(OpenFeatureAPI openFeatureAPI, String name, String version) {
         this.openfeatureApi = openFeatureAPI;
         this.name = name;
         this.version = version;
         this.clientHooks = new ArrayList<>();
+        this.hookSupport = new HookSupport(log);
```
justinabrahms, line 28: Should this not generate it's own logger instance? My understanding of java logging (rusty!) is that generating it w/ the right class name makes for better logs.  Or are you thinking that b/c this is a private utility only for use in this context that sharing the logger is fine?

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -1,135 +1,98 @@
...
+        this.clientHooks.addAll(List.of(hooks));
     }
 
-    <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key, T defaultValue, EvaluationContext ctx, FlagEvaluationOptions options) {
-        FeatureProvider provider = this.openfeatureApi.getProvider();
-        ImmutableMap<String, Object> hints = options.getHookHints();
+    private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key, T defaultValue, EvaluationContext ctx, FlagEvaluationOptions options) {
+        var flagOptions = ObjectUtils.defaultIfNull(options, () -> FlagEvaluationOptions.builder().build());
+        FeatureProvider provider = openfeatureApi.getProvider();
+        Map<String, Object> hints = Collections.unmodifiableMap(flagOptions.getHookHints());
```
justinabrahms, line 39: Thank you!  I knew this was somewhere, but I couldn't find it. :)

justinabrahms (changes requested):

beeme1mr:
Hey @lopitz, it sounds like we could use the generic form you mentioned in your PR but I would just like to confirm that we would still be able to write hooks that work with multiple return types. [Here](https://github.com/open-feature/node-sdk-contrib/blob/alpha/libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts) is an an example OTel hook I've written for Node. It will be important to be able to do something similar in Java.

Thanks again for your contributions. Glad to have your help!

justinabrahms:
> Hey @lopitz, it sounds like we could use the generic form you mentioned in your PR but I would just like to confirm that we would still be able to write hooks that work with multiple return types. [Here](https://github.com/open-feature/node-sdk-contrib/blob/alpha/libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts) is an an example OTel hook I've written for Node. It will be important to be able to do something similar in Java.
> 
> Thanks again for your contributions. Glad to have your help!

This is possible in the provided framework. You'd just use the bare `Hook<T>` instance and deal with casting things yourself.

lib/src/test/java/dev/openfeature/javasdk/DoSomethingProvider.java
toddbaert: When a class contains a single method, it's called a functional interface. Because it contains only a single method, it can be expressed as a single lambda. Basically this class can be implemented by returning a single higher order function.

I think they were introduced in Java 8

lib/src/main/java/dev/openfeature/javasdk/IntegerHook.java
```diff
@@ -0,0 +1,9 @@
+package dev.openfeature.javasdk;
+
+public interface IntegerHook extends Hook<Integer> {
```
toddbaert, line 3: I like this solution, I think it's a helpful guardrail, and it implies that if you use the generic, you have to be careful.

.github/workflows/merge.yml
lopitz: Mmmh, I was following the README.md, where it states, the lib requires JDK11+. Hence I used language features (e.g. `List.of`, `Map.of`, `var`) that is supported by this minimum version. To make it clear. This PR would make the lib incompatible with Java 8 clients. If JDK8 needs to be supported then I will need to adapt the PR. 
From my perspective, it would be beneficial to base this lib on Java 11, even though we might loose some users. Java 11 is out for more than 3 years now and is considered outdated itself. Java 8 is 8 years old at this point and Oracle ended its commercial support in March '22. Having more libs being incompatible w/ such old versions might increase the pressure to update to new JDK versions. IMHO, this would be favorable for us, as maintenance is easier and for the whole industry due to security patches and updates.

Please let me know, which direction we want to go.

lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
lopitz: Well, then it doesn't work with typed Hooks anymore. There can either be a `Hook<String>` or a `Hook<Boolean>` but not a combined one, unless a generic `Hook<?>` is used. 
This is actually a miss on my side, as the generic hooks are currently not called on boolean, string and integer values.

lib/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
lopitz: Usually I don't pass loggers around. In this case, I wanted to make sure, that the logger stays the same, even after moving out the code, in case there is already a logging configuration, that I missed. As it doesn't seem to be that way, I will fetch a new logger in the `HookSupport` class, rather than passing it in.

.github/workflows/merge.yml
justinabrahms: I'm very much on your side with "How are folks not on Java 11 yet?!". The reality is there are several internal usecases to eBay that are still on JDK8. Given I'm authoring the library I get to (at least for now) push my pet use-case of our internal needs and support JDK 8.

I'll fix the README. I agree that it would be beneficial to be on JDK11, but I don't want to do that just yet. Sorry for the confusion!

lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
justinabrahms: What I'm suggesting is this line looks like:

```
             .filter(hook -> hook.supportsFlagValueType(flagValueType)) 
```

The `Hook<String>` implementation would look like:

```
public boolean supportsFlagValueType(FVT type) {
  return type == TYPES.String;
}
```

but this opens the possibility for generic flags to specify the types they want to work with which seems useful?

.github/workflows/merge.yml
toddbaert: I know from personal experience that there's a unfortunate amount of JDK8 out there. I won't comment on if that's the case where I work. :sweat_smile:

lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
lopitz: I changed it but I'm currently not sure, whether I like it. A pro is certainly, that the Hook now decides, whether it can handle a feature flag type.

lopitz:
I changed also the gradle config file to build with Java 8. This is important, as otherwise a Java 11 binary, which is incompatible with Java 8 code bases.

lib/src/main/java/dev/openfeature/javasdk/HookSupport.java
```diff
@@ -0,0 +1,85 @@
...
+        String hookMethod,
+        Consumer<Hook<T>> hookCode
+    ) {
+        hooks
+            .stream()
+            .filter(hook -> isHookCompatible(flagValueType, hook))
+            .forEach(hook -> executeChecked(hook, hookCode, hookMethod));
+    }
+
+    private boolean isHookCompatible(FlagValueType flagValueType, Hook hook) {
```
justinabrahms, line 39: Inline?

justinabrahms (approved):

PR #30, opened 2022-07-01
Quick hacky investigation into event hooks

agentgonzo:
OK, this registers hooks with the global singleton `OpenFeatureApi`. It would probably be better to register the hooks with the `client` instead of the singleton, but this works for the moment. Also, `OpenFeatureAPI` doesn't maintain references to the individual `client`s so it's hard to trigger a callback on the client without some work in that area.

Anyway, It  was a quick look into the mechanism. Many improvements could be made (at the moment, I just registered a hook as receiving a generic string, but we want to properly type it in future)

### DIFF
--- a/src/main/java/dev/openfeature/javasdk/EventAwareFeatureProvider.java
+++ b/src/main/java/dev/openfeature/javasdk/EventAwareFeatureProvider.java
@@ -1,0 +1,7 @@
+package dev.openfeature.javasdk;
+
+import java.util.function.Consumer;
+
+public interface EventAwareFeatureProvider extends FeatureProvider {
+    void setEventHookCallback(Consumer<String> eventHookConsumer);
+}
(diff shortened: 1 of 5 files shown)


beeme1mr:
Hey @agentgonzo, would you be willing to start an [OFEP](https://github.com/open-feature/research) for event hooks?

justinabrahms:
I'm not entirely clear which hooks we'll be firing, so I'm having a difficult time contextualizing this.  Sorry for the delay!

beeme1mr:
Relates to an [OFEP](https://github.com/open-feature/research/pull/25). Closing this PR until the proposal has been accepted.

agentgonzo:
@beeme1mr , I think I must have turned off notifications, as I didn't see the messages on this.

I dont' have any spare time to look at a research topic for this, but happy to take a look in a couple of weeks or answer questions.

beeme1mr:
@agentgonzo, no problem. Todd created a proposal before he left on vacation. When you're back from vacation, your feedback would be highly appreciated.

https://github.com/open-feature/research/pull/25

PR #34, opened 2022-07-23
feat: Add method to get double

### LINKED ISSUE #32: Get float value?
It seems like the Java SDK doesn't support floats. Is that something that should be supported? I suppose this can be worked around with strings, but I wonder if decimal values are important enough to be supported, especially fractional values like .5 could be practical.

Relevant part of the spec: https://github.com/open-feature/spec/blob/main/specification/types.md#number

thomaspoignant:
This PR adds the support of `double` value from the flags.

Resolves #32


beeme1mr (approved):

beeme1mr:
@justinabrahms any objections to merging this PR?

PR #41, opened 2022-08-02
iterators to get all integer, string, boolean, or structure map entries

rgrassian-split:
Signed-off-by: Robert Grassian <robert.grassian@split.io>

### DIFF
--- a/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
+++ b/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
@@ -6,6 +6,7 @@ import lombok.*;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 @ToString @EqualsAndHashCode
@@ -39,12 +40,20 @@ public class EvaluationContext {
         return objMapper.readValue(val, klass);
     }
 
+    public Iterator<Map.Entry<String, String>> getStructureAttributes() {
+        return jsonAttributes.entrySet().iterator();
+    }
+
     public void addStringAttribute(String key, String value) {
         stringAttributes.put(key, value);
     }
 
     public String getStringAttribute(String key) {
         return stringAttributes.get(key);
+    }
+
+    public Iterator<Map.Entry<String, String>> getStringAttributes() {
+        return stringAttributes.entrySet().iterator();
     }
 
     public void addIntegerAttribute(String key, Integer value) {
@@ -55,12 +64,20 @@ public class EvaluationContext {
         return integerAttributes.get(key);
     }
 
+    public Iterator<Map.Entry<String, Integer>> getIntegerAttributes() {
+        return integerAttributes.entrySet().iterator();
+    }
+
     public Boolean getBooleanAttribute(String key) {
         return booleanAttributes.get(key);
     }
 
     public void addBooleanAttribute(String key, Boolean b) {
         booleanAttributes.put(key, b);
+    }
+
+    public Iterator<Map.Entry<String, Boolean>> getBooleanAttributes() {
+        return booleanAttributes.entrySet().iterator();
     }
 
     public void addDatetimeAttribute(String key, ZonedDateTime value) {
--- a/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
+++ b/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
@@ -6,7 +6,6 @@ import lombok.*;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 
 @ToString @EqualsAndHashCode
@@ -40,8 +39,8 @@ public class EvaluationContext {
         return objMapper.readValue(val, klass);
     }
 
-    public Iterator<Map.Entry<String, String>> getStructureAttributes() {
-        return jsonAttributes.entrySet().iterator();
+    public Map<String, String> getStructureAttributes() {
+        return new HashMap<>(jsonAttributes);
     }
 
     public void addStringAttribute(String key, String value) {
@@ -52,8 +51,8 @@ public class EvaluationContext {
         return stringAttributes.get(key);
     }
 
-    public Iterator<Map.Entry<String, String>> getStringAttributes() {
-        return stringAttributes.entrySet().iterator();
+    public Map<String, String> getStringAttributes() {
+        return new HashMap<>(stringAttributes);
     }
 
     public void addIntegerAttribute(String key, Integer value) {
@@ -64,8 +63,8 @@ public class EvaluationContext {
         return integerAttributes.get(key);
     }
 
-    public Iterator<Map.Entry<String, Integer>> getIntegerAttributes() {
-        return integerAttributes.entrySet().iterator();
+    public Map<String, Integer> getIntegerAttributes() {
+        return new HashMap<>(integerAttributes);
     }
 
     public Boolean getBooleanAttribute(String key) {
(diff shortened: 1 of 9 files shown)


beeme1mr:
Relates to https://github.com/open-feature/spec/pull/120

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -39,6 +40,10 @@ public <T> T getStructureAttribute(String key, Class<T> klass) {
         return objMapper.readValue(val, klass);
     }
 
+    public Iterator<Map.Entry<String, String>> getStructureAttributes() {
```
justinabrahms, line 43: Mind adding tests as well? That way it stays working. :)

justinabrahms (changes requested):

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
thomaspoignant: Just to be sure, why have you decided to send a `String` and not an `Object` here?

thomaspoignant:
For the sake of simplicity will it be easier to have a method that returns everything at one rather than having to call 4 iterators to fetch all the data?

In the other SDKs you have access to everything with a single call:
- https://github.com/open-feature/golang-sdk/blob/32d6d647ae8ba8eed0c049aff70b48d6d00a35af/pkg/openfeature/evaluation_context.go#L7
- https://github.com/open-feature/python-sdk/blob/6b05df774e9e2a167b38db884d01c6ffd59acc86/open_feature/evaluation_context/evaluation_context.py#L4
- https://github.com/open-feature/node-sdk/blob/6a0dfb1f9e61149ef9cffba924751b640dd18f54/src/types.ts#L13

This question is more general in the java SDK it-self.
Should we store all the keys in a single `Map` or should we have multiple `Map` as it is right now?

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
rgrassian-split: Yup I will add some tests right now. 
I sent a string here because thats the way we store it serialized currently. If we wanted to fetch it as an object, we would need to know the type of each of the values. This could be done if we keep track of the types of each value in something like a map `{key -> class of value}`, then we would be able to deserialize each object when fetching all. However I'll leave that up to @justinabrahms

rgrassian-split:
> For the sake of simplicity will it be easier to have a method that returns everything at one rather than having to call 4 iterators to fetch all the data?
> 
> In the other SDKs you have access to everything with a single call:
> 
> * https://github.com/open-feature/golang-sdk/blob/32d6d647ae8ba8eed0c049aff70b48d6d00a35af/pkg/openfeature/evaluation_context.go#L7
> * https://github.com/open-feature/python-sdk/blob/6b05df774e9e2a167b38db884d01c6ffd59acc86/open_feature/evaluation_context/evaluation_context.py#L4
> * https://github.com/open-feature/node-sdk/blob/6a0dfb1f9e61149ef9cffba924751b640dd18f54/src/types.ts#L13
> 
> This question is more general in the java SDK it-self. Should we store all the keys in a single `Map` or should we have multiple `Map` as it is right now?

I honestly agree and think we could store everything in one map, that also gets rid of the key collision issue you had pointed out. We could keep everything in a map from string -> string, serializing each value, and maybe do something like the map from key -> class I mentioned above when deserializing. I will leave this up to @justinabrahms though

rgrassian-split:
@justinabrahms @thomaspoignant thoughts on making these return copies of the maps instead of iterator over entries? Would probably be way easier to work with

thomaspoignant:
> @justinabrahms @thomaspoignant thoughts on making these return copies of the maps instead of iterator over entries? Would probably be way easier to work with

 I like this much better, but I let @justinabrahms share his throughts

justinabrahms:
I'm fine with returning copies of maps.

I've steered away from `Map<String, Object>` due to typecasting concerns. I've previously found it much easier to work with strongly defined types.

I think this is the code that we're discussing. Whether we like

## A..
```java
       for (Map.Entry<String, Object> entry : this.attributes.entrySet()) {
            Object value = entry.getValue();
            if (value instanceof Double) {
                doSomethingWith((Double) value);
            } else if (value instanceof Integer) {
                doSomethingWith((Integer) value);
            } else if (value instanceof Boolean) {
                doSomethingWith((Boolean) value);
            } else if (value instanceof String) {
                doSomethingWith((String) value);
            } else { // structure
                doSomethingWith(value);
            }
        }
```
## ..or B
```java
        for (Map.Entry<String, String> entry : this.stringAttributes.entrySet()) {
            doSomethingWith(entry.getValue());
        }

        for (Map.Entry<String, Integer> entry : this.integerAttributes.entrySet()) {
            doSomethingWith(entry.getValue());
        }

        for (Map.Entry<String, Double> entry : this.doubleAttributes.entrySet()) {
            doSomethingWith(entry.getValue());
        }

        for (Map.Entry<String, Object> entry : this.structureAttributes.entrySet()) {
            doSomethingWith(entry.getValue());
        }
```

I would personally prefer B. If you prefer A, I'd be interested in understanding why.

rgrassian-split:
> I'm fine with returning copies of maps.
> 
> I've steered away from `Map<String, Object>` due to typecasting concerns. I've previously found it much easier to work with strongly defined types.
> 
> I think this is the code that we're discussing. Whether we like
> 
> ## A..
> ```java
>        for (Map.Entry<String, Object> entry : this.attributes.entrySet()) {
>             Object value = entry.getValue();
>             if (value instanceof Double) {
>                 doSomethingWith((Double) value);
>             } else if (value instanceof Integer) {
>                 doSomethingWith((Integer) value);
>             } else if (value instanceof Boolean) {
>                 doSomethingWith((Boolean) value);
>             } else if (value instanceof String) {
>                 doSomethingWith((String) value);
>             } else { // structure
>                 doSomethingWith(value);
>             }
>         }
> ```
> 
> ## ..or B
> ```java
>         for (Map.Entry<String, String> entry : this.stringAttributes.entrySet()) {
>             doSomethingWith(entry.getValue());
>         }
> 
>         for (Map.Entry<String, Integer> entry : this.integerAttributes.entrySet()) {
>             doSomethingWith(entry.getValue());
>         }
> 
>         for (Map.Entry<String, Double> entry : this.doubleAttributes.entrySet()) {
>             doSomethingWith(entry.getValue());
>         }
> 
>         for (Map.Entry<String, Object> entry : this.structureAttributes.entrySet()) {
>             doSomethingWith(entry.getValue());
>         }
> ```
> 
> I would personally prefer B. If you prefer A, I'd be interested in understanding why.

B does feel cleaner. I do not feel strongly one way or the other, but we can probably keep this separate from this PR to not block it. Are there any other concerns with it? Tests have been added

src/test/java/dev/openfeature/javasdk/EvalContextTest.java
```diff
@@ -59,4 +62,44 @@ public class EvalContextTest {
...
+        Node<Integer> n1 = new Node<>();
+        n1.value = 4;
+        Node<Integer> n2 = new Node<>();
+        n2.value = 2;
+        n2.left = n1;
+        ec.addStructureAttribute("obj", n2);
+
+        Map<String, String> foundStr = ec.getStringAttributes();
+        assertEquals("test", foundStr.get("str"));
+        assertEquals("test2", foundStr.get("str2"));
```
justinabrahms, line 92: This isn't testing that we support fetching fields by key. Mind updating the tests to do this?
```suggestion
        Map<String, String> foundStr = ec.getStringAttributes();
        assertEquals(ec.getStringAttribute("str"), foundStr.get("str"));
        assertEquals(ec.getStringAttribute("str2"), foundStr.get("str2"));
```

rgrassian-split: done

PR #44, opened 2022-08-10
fix merge of eval context targeting key

rgrassian-split:
The evaluation context sets the targeting key to `""` by default in the constructor. Therefore when two contexts are merged the targeting key is never null and these checks are always true. This results in setting the targeting key always to `""` if `ctx2` did not have a targeting key set, regardless if `ctx1` does have it set

### DIFF
--- a/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
+++ b/src/main/java/dev/openfeature/javasdk/EvaluationContext.java
@@ -133,11 +133,11 @@ public class EvaluationContext {
         ec.attributes.putAll(ctx1.attributes);
         ec.attributes.putAll(ctx2.attributes);
 
-        if (ctx1.getTargetingKey() != null) {
+        if (ctx1.getTargetingKey() != null && !ctx1.getTargetingKey().equals("")) {
             ec.setTargetingKey(ctx1.getTargetingKey());
         }
 
-        if (ctx2.getTargetingKey() != null) {
+        if (ctx2.getTargetingKey() != null && !ctx2.getTargetingKey().equals("")) {
             ec.setTargetingKey(ctx2.getTargetingKey());
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -133,11 +133,11 @@ public static EvaluationContext merge(EvaluationContext ctx1, EvaluationContext
         ec.attributes.putAll(ctx1.attributes);
         ec.attributes.putAll(ctx2.attributes);
 
-        if (ctx1.getTargetingKey() != null) {
+        if (ctx1.getTargetingKey() != null && !ctx1.getTargetingKey().equals("")) {
```
justinabrahms, line 136: Oof! Can you write a regression test for this?

justinabrahms (changes requested):

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
agentgonzo: Apache StringUtils has a bunch of convenience helper methods for this, eg https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNotEmpty-java.lang.CharSequence-

For larger projects, I always tend to include this because it's simpler and less prone to errors... but do we want to introduce the extra dependencies in the SDK, making it bigger?

agentgonzo: (there's also `isNotBlank` which will treat all-whitespace strings like empty strings)

rgrassian-split: Added a test. Also added a trim to the check to handle whitespace, the java sdk is set to Java 8 and isNotBlank is part of 11 so didn't use that.

justinabrahms (approved):

PR #49, opened 2022-08-12
fix: Set no-op provider by default, handle null providers

toddbaert:
- Falls back to the no-op provider if not set. This is an easy way to assure there's no errors thrown if a user has forgotten to register a provider, or due to an race condition where a flag is evaluated before a provider is registered. This is consistent with what's done in the [dotnetSDK](https://github.com/open-feature/dotnet-sdk/blob/73fd72e7190eea585ccefb1116ed778dc2cc740b/src/OpenFeature.SDK/OpenFeature.cs#L15) and the [Node.JS SDK](https://github.com/open-feature/node-sdk/blob/5051681699fd5d805bb8a3ddccd692c45ac95f9e/src/open-feature.ts#L14)
- Moves all author code into the try/catch in the evaluation logic, meaning that we should always get a default, even if authors throw in `provider.getMetadata` or something silly like that

### DIFF
--- a/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
@@ -1,12 +1,16 @@
 package dev.openfeature.javasdk;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
 import dev.openfeature.javasdk.exceptions.GeneralError;
 import dev.openfeature.javasdk.internal.ObjectUtils;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
-
-import java.util.*;
 
 @Slf4j
 @SuppressWarnings({"PMD.DataflowAnomalyAnalysis", "PMD.BeanMembersShouldSerialize", "unchecked", "rawtypes"})
@@ -48,20 +52,25 @@ public class OpenFeatureClient implements Client {
                                                       EvaluationContext ctx, FlagEvaluationOptions options) {
         FlagEvaluationOptions flagOptions = ObjectUtils.defaultIfNull(options,
             () -> FlagEvaluationOptions.builder().build());
-        FeatureProvider provider = openfeatureApi.getProvider();
         Map<String, Object> hints = Collections.unmodifiableMap(flagOptions.getHookHints());
-        if (ctx == null) {
-            ctx = new EvaluationContext();
-        }
-
-        HookContext<T> hookCtx = HookContext.from(key, type, this.getMetadata(),
-                openfeatureApi.getProvider().getMetadata(), ctx, defaultValue);
-
-        List<Hook> mergedHooks = ObjectUtils.merge(provider.getProviderHooks(), flagOptions.getHooks(), clientHooks,
-                openfeatureApi.getApiHooks());
+        ctx =  ObjectUtils.defaultIfNull(ctx, () -> new EvaluationContext());
+        FeatureProvider provider = ObjectUtils.defaultIfNull(openfeatureApi.getProvider(), () -> {
+            log.debug("No provider configured, using no-op provider.");
+            return new NoOpProvider();
+        });
 
         FlagEvaluationDetails<T> details = null;
+        List<Hook> mergedHooks = null;
+        HookContext<T> hookCtx = null;
+        
         try {
+
+            hookCtx = HookContext.from(key, type, this.getMetadata(),
+            openfeatureApi.getProvider().getMetadata(), ctx, defaultValue);
+
+            mergedHooks = ObjectUtils.merge(provider.getProviderHooks(), flagOptions.getHooks(), clientHooks,
+            openfeatureApi.getApiHooks());
+
             EvaluationContext ctxFromHook = hookSupport.beforeHooks(type, hookCtx, mergedHooks, hints);
 
             EvaluationContext invocationCtx = EvaluationContext.merge(ctx, ctxFromHook);
--- a/src/test/java/dev/openfeature/javasdk/DeveloperExperienceTest.java
+++ b/src/test/java/dev/openfeature/javasdk/DeveloperExperienceTest.java
@@ -1,18 +1,28 @@
 package dev.openfeature.javasdk;
 
-import dev.openfeature.javasdk.fixtures.HookFixtures;
-import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
-
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
justinabrahms, line 1: Please remove this file.

src/test/java/dev/openfeature/javasdk/DeveloperExperienceTest.java
```diff
@@ -6,13 +6,23 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
 
 @SuppressWarnings("unchecked")
 class DeveloperExperienceTest implements HookFixtures {
     transient String flagKey = "mykey";
+
+    @Test void noProviderSet() {
```
justinabrahms, line 18: Do you think this should be something that's written in the spec?

justinabrahms (changes requested):

toddbaert:
@justinabrahms My tests work locally, but for some reason in the workflow on github, they all fail. My assertions are always reversed in an odd way:

The resultant string is revered...

```
Error:    DeveloperExperienceTest.noProviderSet:23 expected: <no-op> but was: <po-on>
```

If I pass in a default of true, it's false, and vice versa.

src/test/java/dev/openfeature/javasdk/DeveloperExperienceTest.java
toddbaert: I think we already say to never throw or abnormally end in any case. How to accomplish that seems like a detail. Setting a no-op provider just a nice way to do it.

.vscode/settings.json
toddbaert: Generally this should be committed for shared settings. I can remove it if you feel strongly about it, but it's pretty standard.

toddbaert:
> @justinabrahms My tests work locally, but for some reason in the workflow on github, they all fail. My assertions are always reversed in an odd way:
> 
> The resultant string is reversed...
> 
> ```
> Error:    DeveloperExperienceTest.noProviderSet:23 expected: <no-op> but was: <po-on>
> ```
> 
> If I try to resolve a boolean, and I pass in a default of true, it's false, and vice versa.

Ahh, I see it, it's the `DoSomethingProvider`. It's somehow been attached at the point my test runs.

justinabrahms:
Random thought this weekend.. we might have a package-private method to unset the provider?

toddbaert:
> Random thought this weekend.. we might have a package-private method to unset the provider?

We can already set it to null with the setter. Would the method you mention do something more than that?

src/test/java/dev/openfeature/javasdk/ProviderSpecTest.java
```diff
@@ -37,7 +38,7 @@ public class ProviderSpecTest {
         assertNotNull(boolean_result.getValue());
 
         ProviderEvaluation<Node<Integer>> object_result = p.getObjectEvaluation("key", new Node<Integer>(), new EvaluationContext(), FlagEvaluationOptions.builder().build());
-        assertNotNull(boolean_result.getValue());
+        assertNotNull(object_result.getValue());
```
toddbaert, line 41: Seems like this test was broken due to a typo. This corrects it, I think.

src/test/java/dev/openfeature/javasdk/ProviderSpecTest.java
```diff
@@ -54,16 +55,6 @@ public class ProviderSpecTest {
...
-    @Specification(number="2.8", text="In cases of abnormal execution, the provider MUST indicate an " +
-            "error using the idioms of the implementation language, with an associated error code having possible " +
-            "values PROVIDER_NOT_READY, FLAG_NOT_FOUND, PARSE_ERROR, TYPE_MISMATCH, or GENERAL.")
-    @Disabled("I don't think we expect the provider to do all the exception catching.. right?")
-    @Test void error_populates_error_code() {
-        AlwaysBrokenProvider broken = new AlwaysBrokenProvider();
-        ProviderEvaluation<Boolean> result = broken.getBooleanEvaluation("key", false, new EvaluationContext(), FlagEvaluationOptions.builder().build());
-        assertEquals(ErrorCode.GENERAL, result.getErrorCode());
-    }
-
```
toddbaert, line 66: This was basically testing a provider impl, which we don't need to do. I've deleted this test.

src/test/java/dev/openfeature/javasdk/FlagEvaluationSpecTest.java
```diff
@@ -251,9 +262,10 @@ private Client _client() {
     }
 
     @Specification(number="1.3.3", text="The client SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied default value should be returned.")
+    @Test void type_system_prevents_this() {}
+
     @Specification(number="1.1.6", text="The client creation function MUST NOT throw, or otherwise abnormally terminate.")
-    @Disabled("Not sure how to test?")
-    @Test void not_sure_how_to_test() {}
+    @Test void constructor_does_not_throw() {}
```
toddbaert, line 268: Just added some explanation here as in other tests as to why these tests aren't really necessary for the Java SDK.

src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -48,20 +52,25 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
+        });
 
         FlagEvaluationDetails<T> details = null;
+        List<Hook> mergedHooks = null;
+        HookContext<T> hookCtx = null;
+        
         try {
+
+            hookCtx = HookContext.from(key, type, this.getMetadata(),
+            openfeatureApi.getProvider().getMetadata(), ctx, defaultValue);
```
toddbaert, line 69: Even calls as simple as `.getMetadata()` have been moved to this `try` block.

src/test/java/dev/openfeature/javasdk/ProviderSpecTest.java
```diff
@@ -54,16 +55,6 @@ public class ProviderSpecTest {
         assertNull(result.getErrorCode());
     }
 
-    @Specification(number="2.8", text="In cases of abnormal execution, the provider MUST indicate an " +
```
justinabrahms, line 57: We need to keep this `@Specification` line somewhere. Maybe replace the test method with reasoning for why we don't need it.

toddbaert: Done!

justinabrahms (approved):

beeme1mr (approved):

Issue #50, opened 2022-08-18
Refine type param for getObjectValue/getObjectDetails


toddbaert:
I believe `public <T> T getObjectValue(...` and `public <T> FlagEvaluationDetails<T> getObjectDetails(...` etc are too generic to be useful for parsing and returning an arbitrary JSON structure from a flag backend (see [here](https://github.com/open-feature/research/blob/main/prior-art/api-comparision.md#json) for notes on vendor support for JSON objects).

Because in Java, objects are not stored as dynamically-sized/structured key/value pairs (aka object literals), to instantiate a particular object from a JSON payload, some mapping between the JSON fields and class members must occur (this is basically the whole goal of [JAXB](https://javaee.github.io/jaxb-v2/) and similar annotations). I don't think we want to do this sort of deserialization in our SDK _or even in providers_. Instead, I think we should return a simple representation of a JSON Object (perhaps simply using `javax.json`) or a `Map<String, SomeCustomInterface>` or `Map<String, Object>` (looking for suggestions here). This will allow the application author to handle this payload as they want, and instantiate the objects they need, leaving the SDK out of it. We would restrict the generic T in the aforementioned methods to this type.

justinabrahms:
Relevant: https://launchdarkly.github.io/java-server-sdk/com/launchdarkly/sdk/LDValue.html

justinabrahms:
@toddbaert, I was chatting w/ @s-sen.  Is there a reason why something like Jackson's `ObjectMapper` can't create things from type T in the provider code? 

```java
Node y = new Node();
Node x = getObjectValue("key", y)

// -- inside getObjectValue
ObjectMapper.getValue(innerRepresentation, T.class)
```

toddbaert:
ObjectMapper is a very specific and opinionated JSON deserialization lib that's not part of the Java runtime natively, or even conformant to the javax API - it's Jackson functionality. It just uses conventions (and **non-javax** annotations) to do the deserialization. It will transitively add `jackson-core` and `jackson-annotations` to the SDK.

This is a "nice" solution from an end-user perspective, but it's really doing the type of deserialization I mention above. I'm hesitant to add this sort of dependency to the SDK for compatibility and size reasons.

justinabrahms:
@toddbaert I'm not suggesting we add it to the SDK. I'm suggesting that providers have a facility to accomplish the deserialization but the API still returns `T`, which is the best API for app developers.

toddbaert:
[...]

PR #51, opened 2022-08-22
BREAKING CHANGE: Add generic structure type, use context/getObject

toddbaert:
As described in https://github.com/open-feature/java-sdk/issues/50, I think that expected a `T` back from object flag resolvers is not sustainable. I believe it pushes too much of a burden on providers to instantiate a new T (whatever it may be) and map the attributes received from the backend into it.

Instead, I've added a more generic `Structure` (language used in the OpenFeature spec) with methods similar to the existing `EvaluationContext`. I've also made the `EvaluationContext` inherit from this to reduce duplication (the `EvaluationContext` is also referred to as a "structure" in the spec. 

After this change, providers will create a new `Structure` and add the corresponding fields from the json/xml/protobuf they are concerned with.

In summary:

- Adds `Structure` type which represents JSON data
- `EvaluationContext` ~inherits~ encapsultes this using lombok's `@Delegate`.
- Consolidated `addIntegerAttribute`, `addStringAttribute`, etc, in `EvaluationContext` to a single overloaded `add` method.
- Removed ability to get multiple attributes of a single type from `EvaluationContext` (I'm not convinced these are that valuable, but I'll take feedback).
- introduces a `Value` wrapper class that houses one of: Structure, List<Value>, Boolean, etc (representations of all the possible JSON node values).


beeme1mr:
FYI @rgrassian-split

rgrassian-split:
> FYI @rgrassian-split. I would like to get your thoughts on this. How would this change impact the provider you're working on?

@beeme1mr Hey! This is a fantastic change, I was actually running into some issues with the `T` implementation so I'm happy to see this. It shouldn't be too crazy to change our provider to support this.

toddbaert:
@rgrassian-split it would be good if you could put some feedback in https://github.com/open-feature/java-sdk/issues/50

rgrassian-split:
@toddbaert Just did, that is actually the bigger deal for us that I very much support switching it from the current approach

pom.xml
```diff
@@ -1,5 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
justinabrahms, line 1: If you would like to change whitespace in this file (or others), please check in a tool which will break the build if it's not right. I'm not tied to my particular formatting choices, but don't want us to fight back and forth about what the right answer should be.

pom.xml
```diff
@@ -246,7 +245,7 @@
       <!-- End publish to maven central -->
 
       <!-- Begin source & javadocs being generated -->
-      <plugin>
+      <!-- <plugin>
```
justinabrahms, line 248: Reminder that we'll want to uncomment these here and elsewhere.

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -1,139 +1,31 @@
 package dev.openfeature.javasdk;
 
-import dev.openfeature.javasdk.internal.Pair;
-import lombok.*;
```
justinabrahms, line 4: Similar to above, either make your editor not do these formatting changes or check in something that consistently applies it.

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -1,139 +1,31 @@
...
+@ToString
+@EqualsAndHashCode
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
-public class EvaluationContext {
-    @Setter @Getter private String targetingKey;
-    private final Map<String, Pair<FlagValueType, Object>> attributes;
+public class EvaluationContext extends Structure {
+    @Setter
+    @Getter
+    private String targetingKey;
```
justinabrahms, line 14: I'd actually prefer we not split this across 3 lines. `@Setter` isn't enough information to have it's own line, imo.

src/main/java/dev/openfeature/javasdk/Structure.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Fetch date-time relevant key.
+     * 
+     * @param key feature key
+     * @return date time object.
+     * @throws java.time.format.DateTimeParseException if it's not a datetime
+     */
+    public ZonedDateTime getDatetimeAttribute(String key) {
```
justinabrahms, line 52: Unrelated to your change, but I'm starting to think we should drop this. It feels like it should migrate to contrib, at best.

src/test/java/dev/openfeature/javasdk/EvalContextTest.java
```diff
@@ -1,131 +1,118 @@
...
     }
 
-    @Test void can_chain_attribute_addition() {
+    @Test
+    void can_chain_attribute_addition() {
         EvaluationContext ec = new EvaluationContext();
-        EvaluationContext out = ec.addStructureAttribute("str", "test")
-                .addIntegerAttribute("int", 4)
-                .addBooleanAttribute("bool", false)
-                .addStructureAttribute("str", new Node<Integer>());
```
justinabrahms, line 124: You dropped this one from your test.

justinabrahms (changes requested):
Lots of formatting churn makes the testing stuff (especially spec definitions) a bit hard to read.

src/main/java/dev/openfeature/javasdk/Structure.java
```diff
@@ -0,0 +1,118 @@
...
+        String attr = getAttributeByType(key, FlagValueType.STRING);
+        if (attr == null) {
+            return null;
+        }
+        return ZonedDateTime.parse(attr, DateTimeFormatter.ISO_ZONED_DATE_TIME);
+    }
+
+    public Structure getStructureAttribute(String key) {
```
justinabrahms, line 60: We're also missing something like an array type.

src/main/java/dev/openfeature/javasdk/Structure.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Set;
+
+import dev.openfeature.javasdk.internal.Pair;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@ToString
+@EqualsAndHashCode
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public class Structure {
```
justinabrahms, line 16: If our backend returns this:

```json
{
  "foo": null
}
```

What type will that serialize into on a structure?

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: I think I agree... but there's something like this in the spec. Do you think we need to adjust the spec?

pom.xml
toddbaert: ugh. I didn't mean to commit this.

I have an issue with this pom and Java 18. I was able to resolve it in the contribs.

toddbaert: reverted.

pom.xml
toddbaert: I will see what I can do about this.

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: I think you are right.

src/main/java/dev/openfeature/javasdk/Structure.java
justinabrahms: Provider won't know what type to use. I think the way we work around this is returning `null` for any unknown values. This means if you come across a `null`, you can just not call `add` at all.

src/test/java/dev/openfeature/javasdk/EvalContextTest.java
toddbaert: Added back, thanks!

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: Added

toddbaert:
@justinabrahms I think I've cleaned up a lot of the whitespace changes. A few are possibly improvements (missing newlines in some places, etc). I've left those.

I also added array support in `Structure`.

I'm going to test this locally in my provider and see if it makes things nicer. Let me know what you think.

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: > If we don't call add, then foo won't be in the keySet. So... let's just pick something like boolean and shove all nulls in there as a typecasted thing. Users will get nulls out when they ask for them.

OK that makes sense. That's really something for the provider to do though, right? I guess we could add tests for this, but there isn't much for us to do in the SDK to support this, right?

justinabrahms: That's right.

toddbaert: I'll add a specific test for this.

justinabrahms:
Playing around with the API. It seems like because we broke out `Structure`, the `return this` is returning `Structure` objects, not `EvaluationContext` objects. I'm not aware of a way to have it return `EvaluationContext` objects except to have EC proxy those methods to an internal representation of a `Structure`. Thoughts?

image

toddbaert:
> Playing around with the API. It seems like because we broke out `Structure`, the `return this` is returning `Structure` objects, not `EvaluationContext` objects. I'm not aware of a way to have it return `EvaluationContext` objects except to have EC proxy those methods to an internal representation of a `Structure`. Thoughts?

I noticed that too... it's sort of a shame, but I think we'd have to proxy them as well... maybe there's a clever solution but I can't think of one immediately. I'll look around.

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: I've added such a test.

.github/workflows/pullrequest.yml
```diff
@@ -27,20 +27,8 @@ jobs:
...
-        env:
-          GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
-
       - name: Build with Maven
-        run: |
-          if [ ${{ github.event.pull_request.head.repo.full_name }} = ${{ github.repository }} ]; then
-            mvn --batch-mode --update-snapshots verify
-          else
-            mvn --batch-mode --update-snapshots verify -Dgpg.skip
-          fi
```
toddbaert, line 43: Removed signing in PRs.

pom.xml
```diff
@@ -292,7 +292,8 @@
         <executions>
           <execution>
             <id>sign-artifacts</id>
-            <phase>verify</phase>
+            <!-- we only need to sign if we are deploying (publishing) -->
+            <phase>deploy</phase>
```
toddbaert, line 296: Only sign in deploy. Nice because now you can run `mvn verify` locally and in PRs without worrying about GPG setup.

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -1,139 +1,31 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
 
-@ToString @EqualsAndHashCode
+@ToString @EqualsAndHashCode(callSuper = true)
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
-public class EvaluationContext {
+public class EvaluationContext extends Structure {
```
justinabrahms, line 13: I'm thinking about whether we should be extending structure here, rather than having an instance of Structure as an internal property. I think I have a preference for the latter?

One thing that makes me nervous about this EC/Structure merging is that Array type you added to the flag value types. We don't actually support returning arrays, and so parts of those are leaking between one another.

Thoughts?

justinabrahms: Splitting out the enum will address my concern for FVT. `@Delegate` looks like exactly what I was hoping for. Lombok is the best. Thank you!

toddbaert: Lombok is incredible.

toddbaert: I believe [this commit](https://github.com/open-feature/java-sdk/pull/51/commits/a8623253547411e94fe457adbb1fdd1a973779d7) resolves this concerns.

- encapsulates instead of inherits `Structure`, using lombok's `@Delegate`.
- adds new internal enum for `Structure` field types.

src/main/java/dev/openfeature/javasdk/Structure.java
```diff
@@ -0,0 +1,137 @@
...
+        attributes.put(key, new Pair<>(StructureFieldType.ARRAY, value));
+        return this;
+    }
+
+    /**
+     * Get all attributes, regardless of type.
+     * 
+     * @return all attributes on the structure
+     */
+    public Map<String, Object> getAllAttributes() {
```
kinyoklion, line 119: Would it be possible to get something like this that includes the type annotations? So that if one is converting this structure into something using another structure, or a builder, that they can iterate it and dispatch the methods to the correct types?

justinabrahms: @kinyoklion I think you should be able to do that with `instanceof`?

justinabrahms (approved):

src/main/java/dev/openfeature/javasdk/Structure.java
toddbaert: > @kinyoklion I think you should be able to do that with instanceof?

Indeed, this is what I'm doing in my provider. Is there some reason it's better to leverage the underlying annotation in the `Pair`?

kinyoklion: Yeah, it can be done that way. Also the performance of `instanceof` seems to be good.

The information was already there, so having to immediately re-construct it, seemed a bit counter intuitive.

justinabrahms: Totally. I just don't want it to be part of a public API

toddbaert:
@justinabrahms I'd like to play with this a bit more locally before we merge... I'll let you know if I think any changes are needed. I think this should be sufficient though.

toddbaert:
@justinabrahms  I made some more changes. Our abstraction was leaky because any node in a JSON structure can be a JSON structure, an array or a primitive. That wasn't reflected the implementation.

What I've done in [this commit](https://github.com/open-feature/java-sdk/pull/51/commits/ee58654f6bc02fa5e31b0c734fb85a6068e3cda3), is introduce a `Value` wrapper class that houses one of: `Structure`, `List<Value>`, `Boolean`, etc (representations of all the possible JSON node values). The actual value can be retrieved like `value.asBoolean()` or `value.asStructure()`. I believe any JSON structure can be represented this way.

This is similar to how java-grpc represents their structure (which is very similar to JSON) and something like a simplified version of LDValue.

I've maintained most of the previous work. `EvaluationContext` still encapsulates a `Structure`, etc.

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -86,7 +86,7 @@ public EvaluationContext add(String key, Structure value) {
         return this;
     }
 
-    public <T> EvaluationContext add(String key, List<T> value) {
+    public <T extends Value> EvaluationContext add(String key, List<Value> value) {
```
toddbaert, line 89: Lists must be of type `Value`.

src/main/java/dev/openfeature/javasdk/FlagEvaluationOptions.java
```diff
@@ -1,10 +1,13 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
-import lombok.*;
+import lombok.Builder;
+import lombok.Singular;
 
-@Value
+@lombok.Value
```
justinabrahms, line 10: Not clear why this is being qualified rather than importing it as above? Is it because it's only used once?

src/main/java/dev/openfeature/javasdk/Value.java
```diff
@@ -0,0 +1,206 @@
...
+     * Check if this Value represents a List.
+     * 
+     * @return boolean
+     */
+    public boolean isList() {
+        return this.innerObject instanceof List;
+    }
+
+    /** 
+     * Check if this Value represents a ZonedDateTime.
```
kinyoklion, line 105: Because it shares a storage format with string, would it be worthwhile to have a comment that dates are also strings? If you do care about dates, and you are using this in some kind of dispatch, then you would want to check isZonedDateTime before isString, or check both, or something to disambiguate.

src/main/java/dev/openfeature/javasdk/Value.java
```diff
@@ -0,0 +1,206 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@ToString
+@EqualsAndHashCode
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public class Value {
```
justinabrahms, line 13:
```suggestion
public class StructureValue {
```
?

justinabrahms (commented):
I think I need to play around with constructing a few of these objects to figure out what the API feels like.

src/main/java/dev/openfeature/javasdk/FlagEvaluationOptions.java
justinabrahms: NM. It's now ambiguous. I get it.

src/main/java/dev/openfeature/javasdk/internal/StructureFieldType.java
```diff
@@ -0,0 +1,5 @@
+package dev.openfeature.javasdk.internal;
+
+public enum StructureFieldType {
```
kinyoklion, line 3: This isn't used now that there pairs were removed?

kinyoklion (approved):

src/main/java/dev/openfeature/javasdk/Value.java
toddbaert: Not sure I agree with this one. I would be OK with `Node`, but I think `Value` is better.

Adding "Structure" seems wrong, since the value can be a primitive.

src/main/java/dev/openfeature/javasdk/internal/StructureFieldType.java
toddbaert: Yep, great catch.

toddbaert: Removed.

toddbaert:
> I think I need to play around with constructing a few of these objects to figure out what the API feels like.

OK @justinabrahms, I'll leave it up to you to merge it then (or message with more feedback). @kinyoklion has also approved. I think we'll need to do something similar in dotnet cc @benjiro

src/main/java/dev/openfeature/javasdk/Value.java
toddbaert: I'm wondering if it isn't better to simply store a datetime... Storing them as strings only adds this complication, as well as performance issues around parsing/error handling. It was stored as a string in the previous implementation and I didn't give it much thought when I translated it.

kinyoklion: I think storing it as a datetime, or removing it, are both good options.

src/main/java/dev/openfeature/javasdk/Structure.java
```diff
@@ -0,0 +1,87 @@
...
+        return this;
+    }
+
+    /**
+     * Get all values.
+     * 
+     * @return all attributes on the structure
+     */
+    public Map<String, Value> asMap() {
+        return new HashMap<>(this.attributes);
```
s-sen, line 85: Are we OK with a shallow copy of the values? The application developer will use this to obtain all the values hence a  shallow copy is OK in that scenario. But will this structure also be used during building and mutating the EvaluationContext?

toddbaert: I think this is a good question...

I suspect that for mutating, users will be using the `merge` and `add` functions in the EvaluationContext. I don't have any strong opinions about if this should be a shallow copy, deep copy, or just expose the inner map (though I think in most cases, that's a bad choice). I guess whatever we choose it should be well-documented.

Does anyone have a strong argument for this?

@justinabrahms @kinyoklion

src/main/java/dev/openfeature/javasdk/Value.java
toddbaert: Now storing as ZonedDateTIme.

src/main/java/dev/openfeature/javasdk/Structure.java
justinabrahms: https://github.com/open-feature/java-sdk/issues/53 Moved this discussion here.

Issue #56, opened 2022-09-03
Unable to pull the latest `javasdk` dependency into a Gradle project


ajordens:
Using `implementation 'dev.openfeature:javasdk:0.1.0'` and getting:

```
   > Could not resolve dev.openfeature:javasdk:0.1.0.
     Required by:
         project :gate-package > project :gate-entitytags
      > Could not resolve dev.openfeature:javasdk:0.1.0.
         > inconsistent module metadata found. Descriptor: dev.openfeature:javasdk:0.1.0-SNAPSHOT Errors: bad version: expected='0.1.0' found='0.1.0-SNAPSHOT'
```

Going back to 0.0.2 which predates the `version.modifier` property works fine.

beeme1mr:
Hi @ajordens, thanks for reporting this issue. We'll provide a fix as quickly as possible.

beeme1mr:
@toddbaert, it looks like using an [empty version modifier](https://github.com/open-feature/java-sdk/blob/a38c2d232177a2f5705338fd0992fde4ca3d021a/.github/workflows/release.yml#L45) isn't behaving like we had hoped.

toddbaert:
@beeme1mr you are correct I think that the empty version modifier is to blame here, and was likely never fully working. I don't think we want untransformed properties in the published pom.

I will open a PR to use release please as we do in the java contrib repo, which will solve this issue and also make releases more democratic.

Cc @justinabrahms

justinabrahms:
@toddbaert I don't know what "was likely never fully working" means, but it was definitely working before our latest changes to the pom.xml

toddbaert:
Consistent with what the reporter says, this has been an issue since 0.0.3, where we introduced the empty version modifier.

Compare the pom here: https://search.maven.org/artifact/dev.openfeature/javasdk/0.0.3/jar

With here: https://search.maven.org/artifact/dev.openfeature/javasdk/0.0.2/jar

I believe if somebody has the "version.modifier" set in their environment, it's causing the issue. But 100% sure on that one. I'll add this this dependency works fine for me, but I think the reporter is using Gradle and potentially has that prop defined.

To me it seems like the published pom shouldn't contain that expression (as before 0.0.3)

toddbaert:
[...]

PR #57, opened 2022-09-04
use mvn version to remove -SNAPSHOT, bump 0.1.1

toddbaert:
An expression in the `version` attr (e.g. `<version>0.1.0${version.modifier}</version>`) is causing [issues](https://github.com/open-feature/java-sdk/issues/56#issuecomment-1236223635) for gradle.

Instead of using expressions in the pom, this PR modifies the release action to use `mvn versions:set -DremoveSnapshot` to remove the `-SNAPSHOT` modifier, which is built-in functionality of the maven version plugin.

I think long term, we want to move to [release please](https://github.com/googleapis/release-please) for a few reasons:

- it uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to keep track of when we should consider versions breaking
- it can be configured to update all instances of a version, across all files in the repo (so even our readme will be updated correctly)
- it keeps a running PR of the potential release, so we can democratically and transparently decide when to cut one.
- other OpenFeature SDKs and contribs are already using it

For now I'm just doing this, which is a quick fix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,11 +38,14 @@ jobs:
         echo -n "$GPG_SIGNING_KEY" | base64 --decode | gpg --import
       env:
         GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
+      
+    - name: Remove SNAPSHOT
+      run: mvn versions:set -DremoveSnapshot
 
     - name: Deploy
-      run: |        
+      run: |
         mvn --batch-mode \
-          --settings release/m2-settings.xml clean deploy -Dversion.modifier=''
+          --settings release/m2-settings.xml clean deploy
       env:
         OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
         OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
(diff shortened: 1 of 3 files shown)


beeme1mr (approved):

justinabrahms (approved):
It's almost like the giant compiler warnings about this meant we were doing something wrong. 😬

I'm torn for conventional commits bc I see the value and also don't want to change how I write my commit messages. I think this may be a "grumble and do it anyway" sorta thing.

beeme1mr:
We've been using the pr titles as the commit messages when merging a pr. I can show you next week if you're interested.

toddbaert:
> It's almost like the giant compiler warnings about this meant we were doing something wrong. grimacing

Oh shoot, now that you mention it, I remember seeing those too. :facepalm: 

> I'm torn for conventional commits bc I see the value and also don't want to change how I write my commit messages. I think this may be a "grumble and do it anyway" sorta thing.

I understand what you're saying. Like @beeme1mr says we force only the PR title to be semantic (validated by an action) and then github squashes all the commits in the PR down to a new commit containing the title, so you can commit how you want. I feel like it's a pretty solid solution that doesn't add much overhead. It's really easy to change the commit title and it's super transparent.

justinabrahms:
That works for me. Thanks.

Issue #59, opened 2022-09-06
Instant instead of ZonedDateTime


agentgonzo:
`Instant` is a more ideomatic way of representing times in java (akin to unix time). It's a timezone-less entity (ie, UTC).

There are a bunch of helper functions to do any conversions you may want, as well as to/from ISO-8601 zulu-time strings.

ZonedDateTime contains timezone information, which (whilst useful) can introduce confusion when dealing with distributed systems. Ideally, all backend/server code (ie, Java) would use `Instant` (ie, fancy-formatted unix time) for all datetime representations, then leave it up to the frontend to worry about formatting with specific timezones.

I think it's worth changing https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/javasdk/Value.java#L49-L51 to be using `Instant` instead of `ZonedDateTime`

toddbaert:
I'm in agreement with this suggestion, though we also had a brief discussion on whether or not we wanted to include timestamps as "first class" data in the context or structure return values at all.

I think I'd like to keep timestamps as a first class type in the context, but I'm interested if others think it's worth the burden.

PR #65, opened 2022-09-09
Collection of small, breaking changes

toddbaert:
This is a collection of small (but breaking) changes. I hope these are the last breaking changes before a 1.0rc. I've broken it into 3 commits for easy understanding:

- add object constructor and accessor to `Value`: https://github.com/open-feature/java-sdk/commit/0152a1eef93ea1b5253ddae78718a9805c98aaf7
- use `Instant` instead of `ZonedDateTime` (https://github.com/open-feature/java-sdk/issues/59): https://github.com/open-feature/java-sdk/commit/3e6241422266825f267043e4acd116803c4939b0
- use `Value` instead of `Structure` in object resolver (https://github.com/open-feature/spec/issues/138#issuecomment-1241202093): https://github.com/open-feature/java-sdk/commit/5d262470e8ec47d2af35f0aabe55e8c969e992ac

### DIFF
--- a/src/main/java/dev/openfeature/javasdk/Value.java
+++ b/src/main/java/dev/openfeature/javasdk/Value.java
@@ -1,6 +1,6 @@
 package dev.openfeature.javasdk;
 
-import java.time.ZonedDateTime;
+import java.time.Instant;
 import java.util.List;
 
 import lombok.EqualsAndHashCode;
@@ -54,7 +54,7 @@ public class Value {
         this.innerObject = value; 
     }
 
-    public Value(ZonedDateTime value) {
+    public Value(Instant value) {
         this.innerObject = value;
     }
 
@@ -113,12 +113,12 @@ public class Value {
     }
 
     /** 
-     * Check if this Value represents a ZonedDateTime.
+     * Check if this Value represents an Instant.
      * 
      * @return boolean
      */
-    public boolean isZonedDateTime() {
-        return this.innerObject instanceof ZonedDateTime;
+    public boolean isInstant() {
+        return this.innerObject instanceof Instant;
     }
     
     /** 
@@ -207,13 +207,13 @@ public class Value {
     }
 
     /** 
-     * Retrieve the underlying ZonedDateTime value, or null.
+     * Retrieve the underlying Instant value, or null.
      * 
-     * @return ZonedDateTime
+     * @return Instant
      */
-    public ZonedDateTime asZonedDateTime() {
-        if (this.isZonedDateTime()) {
-            return (ZonedDateTime)this.innerObject;
+    public Instant asInstant() {
+        if (this.isInstant()) {
+            return (Instant)this.innerObject;
         }
         return null;
     }
--- a/src/main/java/dev/openfeature/javasdk/Value.java
+++ b/src/main/java/dev/openfeature/javasdk/Value.java
@@ -22,6 +22,10 @@ public class Value {
         this.innerObject = null; 
     }
 
+    public Value(Object value) {
+        this.innerObject = value; 
+    }
+
     public Value(Value value) {
         this.innerObject = value.innerObject; 
     }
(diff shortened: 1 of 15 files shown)


justinabrahms:
Can you also delete the Structure object since we're not using it anymore?

toddbaert:
> Can you also delete the Structure object since we're not using it anymore?

We are still using it, in fact, it's quite important. A value can BE a structure (that would be a JSON node that contains a JSON object, in JSON context). You can see it used in `Value.asStructure()`.

src/main/java/dev/openfeature/javasdk/Value.java
```diff
@@ -22,6 +22,10 @@ public Value() {
         this.innerObject = null; 
     }
 
+    public Value(Object value) {
+        this.innerObject = value;
```
kinyoklion, line 26: I am a  bit weary of this.

It effectively removes the constraints on the values in lists. Because that is an external way a value can be accepted directly. I suppose provider wise I could just ignore values that don't comply to any of the `is` checks.

toddbaert: > I am a bit weary of this.
> 
> It effectively removes the constraints on the values in lists. Because that is an external way a value can be accepted directly. I suppose provider wise I could just ignore values that don't comply to any of the `is` checks.

I think you mean "wary", but point taken!

We could throw if the object is not one of the valid types (but I'm a little worried about the consequences of that) - or we could make the inner value null in that case. I'm not sure either of these is a great solution.

@rgrassian-split specifically requested this to make his provider simpler... but he's doing the type of check you're suggesting. Maybe this change removes some guardrails we actually want. I'm open to further discussion and input from @rgrassian-split as well.

justinabrahms: It could maybe complain loudly if it doesn't pass one of the `is` checks.

justinabrahms:
@toddbaert Gotcha. So structure = json collection, Value = "holder of some json node"

src/main/java/dev/openfeature/javasdk/Value.java
toddbaert: > The constructor could maybe complain loudly if it doesn't pass one of the is checks.

If we throw a checked exception if the object is not a valid type, we will have safety and flexibility, I think. It will just require the called to use a try/catch construct. I personally wouldn't use this in my provider, but I think that might solve @rgrassian-split 's use case without causing issues down the line as @kinyoklion is validly concerned about.

What do you guys think?

kinyoklion: I think that makes sense.

toddbaert: OK. I've done so: https://github.com/open-feature/java-sdk/pull/65/commits/96623191d05937dfabe1e6cef9e4c03494ddca8b

src/main/java/dev/openfeature/javasdk/Value.java
```diff
@@ -19,7 +19,27 @@ public class Value {
...
+    public Value(Object value) throws InstantiationException {
+        // integer is a special case, convert those.
+        this.innerObject = value instanceof Integer ? ((Integer)value).doubleValue() : value;
+        if (!(this.isNull()
+            || this.isBoolean()
+            || this.isString()
+            || this.isNumber()
+            || this.isStructure()
+            || this.isList()
+            || this.isInstant())) {
```
toddbaert, line 40: We can apply DeMorgan's law here if anyone prefers the other way around. I never know which is more re-able.

src/main/java/dev/openfeature/javasdk/Value.java
rgrassian-split: Yeah I agree the type checking should still happen and am totally fine handling a case where the type doesn't match

src/main/java/dev/openfeature/javasdk/Value.java
rgrassian-split: I think its slightly more readable with the negation distributed since it's a longer list. But  not a strong preference.

toddbaert: done!

justinabrahms (approved):

PR #70, opened 2022-09-13
fix: isList check in Value checks type of list

rgrassian-split:
A list of any type would be able to be created due to the new object constructor. This makes it so that the isList check only returns true if the list contains objects of type Value (or is empty).


rgrassian-split:
@justinabrahms @toddbaert please check out this PR, as I was updating the Split Provider I noticed it would be possible to set a list of any time due to the new constructor. This change should make it so that only lists containing Values are allowed to be created through that constructor

justinabrahms (approved):

toddbaert:
Thanks for adding this check. I think we'll have a similar issue in dotnet too. I will address it there.

toddbaert:
Thinking about it more, would it be better to instead check that the list type is sane when we create it? @rgrassian-split

rgrassian-split:
> Thinking about it more, would it be better to instead check that the list type is sane when we instantiate the value? The Object constructor already throws. We could add this same test on the object constructor and throw if the List is not of type value. That way, it's impossible to ever create such a value. @rgrassian-split

@toddbaert Can you elaborate, I'm a bit confused. Right now the object constructor calls the isList method which would fail so it's impossible to create a value that is a list of things other than values.

Also as an fyi, it turned out that this object constructor was less useful than we thought it would be and I'm not sure we will end up using it for our Split Provider. The reason being if we have on our end something like a List<String> and want it as a value we would really want to turn this into something like Value<List<Value<String>>>, but the object constructor can't do that for us so that's a case we still have to handle special.

The fetching as an object has been very useful though thanks for that

toddbaert:
@rgrassian-split oof, I misread the test, and didn't recall that `isList()` is called from the object constructor - so the net behavior you've implemented is basically what I was hoping for.

PR #75, opened 2022-09-16
feat: Add asObjectMap to get the EvaluationContext as Map<String,Object>

thomaspoignant:
Add a new method called `asObjectMap` that convert the `EvaluationContext` into a `Map<String,Object>`.

This utility method will be used by the providers when they have to convert the `EvaluationContext` to transform them into the query they need to send to the backend system.


agentgonzo (approved):
Fine be me from a provider's standpoint. But I can remember that @justinabrahms wanted to keep typing around as much as possible.

toddbaert:
I think this is a good improvement. If this method is used it sacrifices some type safety, but easily allows people to use things like ObjectMapper without implementing this boilerplate.

PR #80, opened 2022-09-27
feat!: errorCode as enum, reason as string

toddbaert:
- changes `errorCode` in resolution/evaluation structures to enum
- changes `reason` in resolution/evaluation structures to string

Closes: https://github.com/open-feature/java-sdk/issues/78

:warning: the flagd integration tests WILL fail with this PR :warning: - that's because the flagd provider needs these changes (a bit of a chicken/egg situation). I have a provider update ready locally to push as soon as this is published, which will then get the integration tests working again.

Closes: https://github.com/open-feature/java-sdk/issues/78


src/main/java/dev/openfeature/javasdk/NoOpProvider.java
```diff
@@ -25,7 +25,7 @@ public ProviderEvaluation<Boolean> getBooleanEvaluation(String key, Boolean defa
         return ProviderEvaluation.<Boolean>builder()
                 .value(defaultValue)
                 .variant(PASSED_IN_DEFAULT)
-                .reason(Reason.DEFAULT)
+                .reason(Reason.DEFAULT.toString())
```
toddbaert, line 28: I think it's still good to have an enum of default reasons. We just convert them to strings were applicable. We can instead use string constants... anyone have opinions on this?

src/main/java/dev/openfeature/javasdk/FlagEvaluationDetails.java
```diff
@@ -14,8 +14,9 @@
     private String flagKey;
     private T value;
     @Nullable private String variant;
-    private Reason reason;
-    @Nullable private String errorCode;
+    @Nullable private String reason;
+    private ErrorCode errorCode;
+    @Nullable private String message;
```
toddbaert, line 19: Spec change.

src/main/java/dev/openfeature/javasdk/ErrorCode.java
```diff
@@ -1,5 +1,5 @@
 package dev.openfeature.javasdk;
 
 public enum ErrorCode {
-    PROVIDER_NOT_READY, FLAG_NOT_FOUND, PARSE_ERROR, TYPE_MISMATCH, GENERAL
+    PROVIDER_NOT_READY, FLAG_NOT_FOUND, PARSE_ERROR, TYPE_MISMATCH, TARGETING_KEY_MISSING, INVALID_CONTEXT, GENERAL
```
toddbaert, line 4: Spec change.

.github/workflows/pullrequest.yml
```diff
@@ -34,7 +34,7 @@ jobs:
             ${{ runner.os }}-maven-
 
       - name: Build with Maven
-        run: mvn --batch-mode --update-snapshots verify -P integration-test
+        run: mvn --batch-mode --update-snapshots verify # -P integration-test - add this back once we have a compatible flagd
```
toddbaert, line 37: Removing the `integration-test` profile disables the integration test suite as discussed in the description.

toddbaert:
FYI @thomaspoignant  I have updates to the contribs pending locally, including the flagd and the go-feature-flag provider for this.

src/main/java/dev/openfeature/javasdk/exceptions/FlagNotFoundError.java
```diff
@@ -7,5 +7,5 @@
 @StandardException
 public class FlagNotFoundError extends OpenFeatureError {
     private static final long serialVersionUID = 1L;
-    @Getter private final ErrorCode errorCode = ErrorCode.GENERAL;
+    @Getter private final ErrorCode errorCode = ErrorCode.FLAG_NOT_FOUND;
```
toddbaert, line 10: I think this was an existing bug, unless I'm missing something.

src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -94,9 +95,14 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
+            if (e instanceof OpenFeatureError) {
+                details.setErrorCode(((OpenFeatureError)e).getErrorCode());
+            } else {
+                details.setErrorCode(ErrorCode.GENERAL);
+            }
+            details.setMessage(e.getMessage());
             details.setValue(defaultValue);
-            details.setReason(Reason.ERROR);
-            details.setErrorCode(e.getMessage());
+            details.setReason(Reason.ERROR.toString());
```
toddbaert, line 105: This is the only behavioral change - if we can get an errorCode, us it. Otherwise use `GENERAL`.

Also, get the message from the message. Previously, the code was coming from the message.

src/main/java/dev/openfeature/javasdk/ProviderEvaluation.java
```diff
@@ -9,6 +9,7 @@
 public class ProviderEvaluation<T> implements BaseEvaluation<T> {
```
kinyoklion, line 9: Would it be useful to have an overload for setting the reason based on a Reason? I am not really sure how these lombok classes work, aside from magic.

src/test/java/dev/openfeature/javasdk/ProviderSpecTest.java
```diff
@@ -45,7 +45,7 @@ public class ProviderSpecTest {
             "reason field with a string indicating the semantic reason for the returned flag value.")
```
kinyoklion, line 45: The 2.6 spec fails the `spec_finder.py` check. I think because it was updated with a list of common reasons.

src/main/java/dev/openfeature/javasdk/FlagEvaluationDetails.java
justinabrahms: I think I missed this. We have both `reason` and `message`? Why would you want both?

toddbaert: Reason is always present. It's the semantic reason for a flag evaluation. `message` is literally the `message` property of a thrown exception. You can see we populate it with `e.message` in the handler.

I think it was one of the more popular parts of this change because actual error messages were being swallowed. It's quite useful for debugging.

toddbaert: I've updated this to `errorMessage`

toddbaert:
Similar PR in c# for any interested: https://github.com/open-feature/dotnet-sdk/pull/72

src/main/java/dev/openfeature/javasdk/ProviderEvaluation.java
toddbaert: I'm hesitant to add it b/c it's not something lombok can do for us. I'm lazy.

src/main/java/dev/openfeature/javasdk/OpenFeatureClient.java
```diff
@@ -94,9 +95,14 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
+            if (e instanceof OpenFeatureError) {
+                details.setErrorCode(((OpenFeatureError)e).getErrorCode());
+            } else {
+                details.setErrorCode(ErrorCode.GENERAL);
+            }
+            details.setErrorMessage(e.getMessage());
             details.setValue(defaultValue);
-            details.setReason(Reason.ERROR);
-            details.setErrorCode(e.getMessage());
+            details.setReason(Reason.ERROR.toString());
```
skyerus, line 105: Should we consider whether a provider might want to different reason to `ERROR`?

toddbaert: The spec says that reason must indicate an error in error cases: https://github.com/open-feature/spec/blob/f74a8d98763bcac5d9a5c67a40ca9d7c2d445180/specification/sections/01-flag-evaluation.md#requirement-148

I suppose the provider could provide a more specific error reason, but I think additional information can be conveyed by the `errorMessage` and `errorCode` fields.

src/main/java/dev/openfeature/javasdk/NoOpProvider.java
thomaspoignant: Yes this a good idea to keep the enum 👍

justinabrahms (approved):

PR #111, opened 2022-10-03
refactor!: Change the package name. Everyone knows it's java (or it doesn't matter)

### LINKED ISSUE #82: Change package name?
On the Sept 29 community call, an attendee suggested we remove "java" from the package names in the SDK (their argument was "java" is implied). I see their point, though I also don't think it's needed.

Basically: `dev.openfeature.javasdk` would become: `dev.openfeature.sdk`.

I'm just bringing this up because "now is the time" before we start marking things as stable. Mostly interested in @justinabrahms 's opinion on this one.

**UPDATE: We will likely do this at 1.0**

justinabrahms:
Fixes #82


justinabrahms:
I believe this is blocked on a circular dependency w/ FlagD provider.

toddbaert:
> I believe this is blocked on a circular dependency w/ FlagD provider.

Yep. The integration tests are already disabled for a similar reason. I've pushed an update that temporarily sidesteps this issue since the tests are disabled now anyways. I will restore it all once we've absorbed the changes in flagd.

toddbaert (dismissed):

toddbaert:
@justinabrahms looks like your 2nd commit is missing a signoff. I'm in favor of merging after that.

toddbaert:
I re-applied my last commit on top. I'm assuming you removed it accidentally when you sign your old commits.

toddbaert (approved):

PR #112, opened 2022-10-04
feat!: use evaluation context interface

toddbaert:
This attempts to defer our immutability issues. It does this by:

- Extracting the read operations of `EvaluationContext` into and interface called `EvaluationContext`
- renaming the old `EvaluationContext` to `MutableContext`, and making it implement `EvaluationContext`
- Extracting the read operations of `Structure` into and interface called `Structure`
- renaming the old `Structure` to `MutableStructure`, and making it implement `Structure`
- all SDK methods now work with the new `EvaluationContext` interface

We can now implement whatever context we like - context that lock with mutexes, immutable contexts, whatever.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/Client.java
+++ b/src/main/java/dev/openfeature/sdk/Client.java
@@ -10,7 +10,7 @@ public interface Client extends Features {
 
     /**
      * Return an optional client-level evaluation context.
-     * @return {@link MutableContext}
+     * @return {@link EvaluationContext}
      */
     EvaluationContext getEvaluationContext();
 
--- a/src/main/java/dev/openfeature/sdk/Client.java
+++ b/src/main/java/dev/openfeature/sdk/Client.java
@@ -10,7 +10,7 @@ public interface Client extends Features {
 
     /**
      * Return an optional client-level evaluation context.
-     * @return {@link EvaluationContext}
+     * @return {@link MutableContext}
      */
     EvaluationContext getEvaluationContext();
 
--- a/src/main/java/dev/openfeature/sdk/EvaluationContext.java
+++ b/src/main/java/dev/openfeature/sdk/EvaluationContext.java
@@ -1,29 +1,13 @@
 package dev.openfeature.sdk;
 
-import java.time.Instant;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 
-import lombok.Getter;
-import lombok.Setter;
-import lombok.ToString;
-import lombok.experimental.Delegate;
-
-@ToString
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
-public class EvaluationContext {
-
-    @Setter @Getter private String targetingKey;
-    @Delegate(excludes = HideDelegateAddMethods.class) private final Structure structure = new Structure();
-
-    public EvaluationContext() {
-        super();
-        this.targetingKey = "";
-    }
-
-    public EvaluationContext(String targetingKey) {
-        this();
-        this.targetingKey = targetingKey;
-    }
+public interface EvaluationContext extends Structure {
+    String getTargetingKey();
+    
+    void setTargetingKey(String targetingKey);
 
     /**
      * Merges two EvaluationContext objects with the second overriding the first in
@@ -33,16 +17,18 @@ public class EvaluationContext {
      * @param ctx2 overriding context
      * @return resulting merged context
      */
-    public static EvaluationContext merge(EvaluationContext ctx1, EvaluationContext ctx2) {
-        EvaluationContext ec = new EvaluationContext();
+    static EvaluationContext merge(EvaluationContext ctx1, EvaluationContext ctx2) {
         if (ctx1 == null) {
             return ctx2;
         } else if (ctx2 == null) {
             return ctx1;
         }
 
-        ec.structure.attributes.putAll(ctx1.structure.attributes);
-        ec.structure.attributes.putAll(ctx2.structure.attributes);
+        Map<String, Value> merged = new HashMap<String, Value>();
+
+        merged.putAll(ctx1.asMap());
+        merged.putAll(ctx2.asMap());
+        EvaluationContext ec = new MutableContext(merged);
 
         if (ctx1.getTargetingKey() != null && !ctx1.getTargetingKey().trim().equals("")) {
             ec.setTargetingKey(ctx1.getTargetingKey());
@@ -53,78 +39,5 @@ public class EvaluationContext {
         }
 
         return ec;
-    }
-
-    // override @Delegate methods so that we can use "add" methods and still return EvaluationContext, not Structure
-    public EvaluationContext add(String key, Boolean value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, String value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, Integer value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, Double value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, Instant value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, Structure value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    public EvaluationContext add(String key, List<Value> value) {
-        this.structure.add(key, value);
-        return this;
-    }
-
-    /**
-     * Hidden class to tell Lombok not to copy these methods over via delegation.
-     */
-    private static class HideDelegateAddMethods {
-        public Structure add(String ignoredKey, Boolean ignoredValue) {
-            return null;
-        }
-        
-        public Structure add(String ignoredKey, Double ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, String ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, Value ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, Integer ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, List<Value> ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, Structure ignoredValue) {
-            return null;
-        }
-
-        public Structure add(String ignoredKey, Instant ignoredValue) {
-            return null;
-        }
     }
 }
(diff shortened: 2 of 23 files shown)


src/main/java/dev/openfeature/javasdk/HashMapStructure.java
```diff
@@ -0,0 +1,161 @@
+package dev.openfeature.javasdk;
```
toddbaert, line 1: This is just a rename. I had to extract `Strcture` to an interface too.

src/main/java/dev/openfeature/javasdk/EvaluationContext.java
```diff
@@ -33,16 +17,18 @@ public EvaluationContext(String targetingKey) {
...
             return ctx1;
         }
 
-        ec.structure.attributes.putAll(ctx1.structure.attributes);
-        ec.structure.attributes.putAll(ctx2.structure.attributes);
+        Map<String, Value> merged = new HashMap<String, Value>();
+
+        merged.putAll(ctx1.asMap());
+        merged.putAll(ctx2.asMap());
+        EvaluationContext ec = new MutableContext(merged);
```
skyerus, line 31: My java knowledge is extremely limited but it's surprising to me that there's a concrete type being returned in the interface definition

skyerus: I went for the approach of including the merge as a method in the `EvaluationContext` interface, thereby leaving it to each implementation:
https://github.com/open-feature/go-sdk/pull/92/files#diff-d0bdacefeb6c765ab5d7987bdfcd09508b4f76e591bc7720f1a484c836625140R15

toddbaert: I've addressed this as above. Thanks! (now you have to help write Java)

toddbaert:
@justinabrahms I've marked this ready. I've tested this locally in the flagd providers and it works as expected. Please review the names and interfaces, especially.

src/main/java/dev/openfeature/sdk/Client.java
```diff
@@ -10,7 +10,7 @@ public interface Client extends Features {
 
     /**
      * Return an optional client-level evaluation context.
-     * @return {@link EvaluationContext}
+     * @return {@link MutableContext}
      */
     EvaluationContext getEvaluationContext();
```
justinabrahms, line 15: These docs say it should return a MutableContext? I'm thinking the docs are wrong?

src/main/java/dev/openfeature/sdk/EvaluationContext.java
```diff
@@ -1,130 +1,17 @@
...
-        this.targetingKey = targetingKey;
-    }
+public interface EvaluationContext extends Structure {
+    String getTargetingKey();
+    
+    void setTargetingKey(String targetingKey);
 
     /**
-     * Merges two EvaluationContext objects with the second overriding the first in
+     * Merges this EvaluationContext objects with the second overriding the this in
```
justinabrahms, line 10:
```suggestion
     * Merges this EvaluationContext object with the second overriding the this in
```

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Map;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+import lombok.experimental.Delegate;
+
+@ToString
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public class MutableContext implements EvaluationContext {
```
justinabrahms, line 15: Can I get a java doc to understand what this really is and how it differs from EvaluationContext?

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -0,0 +1,142 @@
...
+        this.structure = new MutableStructure(attributes);
+        this.targetingKey = "";
+    }
+
+    public MutableContext(String targetingKey, Map<String, Value> attributes) {
+        this(attributes);
+        this.targetingKey = targetingKey;
+    }
+
+    // override @Delegate methods so that we can use "add" methods and still return EvaluationContext, not Structure
```
justinabrahms, line 40:
```suggestion
    // override @Delegate methods so that we can use "add" methods and still return MutableContext, not Structure
```

src/main/java/dev/openfeature/sdk/MutableStructure.java
```diff
@@ -0,0 +1,162 @@
+package dev.openfeature.sdk;
```
justinabrahms, line 1: I'm not following how/why this is a new file?

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -67,23 +69,23 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
-                    ),
-                    invocationCtx
-            );
+            EvaluationContext apiContext = openfeatureApi.getEvaluationContext() != null
+                    ? openfeatureApi.getEvaluationContext()
+                    : new MutableContext();
+            EvaluationContext clientContext = openfeatureApi.getEvaluationContext() != null
+                    ? this.getEvaluationContext()
+                    : new MutableContext();
+            EvaluationContext mergedCtx = apiContext.merge(clientContext.merge(invocationCtx));
```
justinabrahms, line 88: Why did this change?

justinabrahms (commented):
I'm not entirely clear why we're passing around MutableContexts so much. I'd guess that it would be only used by users and providers / hooks would all use the interface?

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
toddbaert: Merge was previously a static method. I don't think it can be any longer, since we have to instantiate a new instance, or return one of the existing ones (so we need a `this`).

Previously, `merge` could easily handle if once of the context passed were null. With the new instance method, we have to make sure the the context is not null before calling it. In many of the tests, `this.getEvaluationContext()` and `openfeatureApi.getEvaluationContext()` were null.

src/main/java/dev/openfeature/sdk/MutableStructure.java
toddbaert: `EvaluationContext` is an interface that extends `Structure`. `EvaluationContext` has a superset of `Structure`'s methods. This was the case previously, except it was done via encapsulation and proxying via lombok.

I've made them both interfaces now. `MutableStructure` is just a concrete implementation of `Structure`, and as before `MutableContext` encapsulates and proxies to `MutableStructure`. 

I hope that answers your question. This wasn't entirely necessary, but I think it's better than `Structure` remaining concrete, and duplicating it's methods in `EvaluationContext`, which is an interface and therefore couldn't extend it.

src/main/java/dev/openfeature/sdk/Client.java
toddbaert: Ya, bulk replace. I will fix.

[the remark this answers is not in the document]
toddbaert: Bulk replace again.

toddbaert:
>I'm not entirely clear why we're passing around MutableContexts so much. I'd guess that it would be only used by users and providers / hooks would all use the interface?

There may be a few cases where I use it unnecessarily, but in some cases we need to create an instance within the client/API. I'll see what I can do to minimize these.

src/main/java/dev/openfeature/sdk/Client.java
toddbaert: I think this is the same as https://github.com/open-feature/java-sdk/pull/112#discussion_r988374129

kinyoklion:
It is possible that I am maybe not completely following this, but I think if the idea is to put contexts behind an interface, and then to later be able to swap them with minimal breakage, would you not want both mutable, and readonly access, to be interfaces, and for the instances to be created with factories (not builders, but just make me a new mutable instance)? 

This would be especially important if the thought is that different implementations might exist to cater to different sensitives.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -50,11 +52,11 @@ public void addHooks(Hook... hooks) {
...
 
     private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key, T defaultValue,
-                                                      EvaluationContext ctx, FlagEvaluationOptions options) {
+            EvaluationContext ctx, FlagEvaluationOptions options) {
         FlagEvaluationOptions flagOptions = ObjectUtils.defaultIfNull(options,
-            () -> FlagEvaluationOptions.builder().build());
+                () -> FlagEvaluationOptions.builder().build());
         Map<String, Object> hints = Collections.unmodifiableMap(flagOptions.getHookHints());
-        ctx = ObjectUtils.defaultIfNull(ctx, () -> new EvaluationContext());
+        ctx = ObjectUtils.defaultIfNull(ctx, () -> new MutableContext());
```
toddbaert, line 59: @justinabrahms I think this is a case we need to instantiate something. We need to pass some kind of context to the hooks (I don't think we want it to be null if none happens to have been set).

I don't _think_ this is a problem, since in the hook it will be an `EvaluationContext` interface which (unless unsafely casted) can only be used for reads.

kinyoklion:
> > It is possible that I am maybe not completely following this, but I think if the idea is to put contexts behind an interface, and then to later be able to swap them with minimal breakage, would you not want both mutable, and readonly access, to be interfaces, and for the instances to be created with factories (not builders, but just make me a new mutable instance)?
> > This would be especially important if the thought is that different implementations might exist to cater to different sensitives.
> 
> I could be missing something, but the SDK itself never needs to mutate a context. That only occurs in `application author` code. For this reason, I think only the read operations are relevant to the interface. I think with the current scheme, we could implement an `ImmutableContext` that takes a map in it's constructor, or can be built with a builder, but is immutable after instantiation. It could be used everywhere a `MutableContext` is.

I guess I mostly concerned with before hooks. They could some from a number of places and would need to be able to create some level of mutable context. A provider hook specifically. I guess they could make anything  that complies with the interface, and the considerations are more constrained in that case. (If I were implementing one though, I would just want to make whatever instance type was canonical, or whatever mutable type the application is using.)

toddbaert:
>  I guess they could make anything that complies with the interface, and the considerations are more constrained in that case.

The considerations more constrained, and additionally, the before hook also allows authors to return an instance of a context, which is merged with the ambient context as per the spec. This is the idiomatic way to alter the context from a before hook.

I think I see your point though, that hook authors have to make a choice here, and it's somewhat disconnected from whatever instances the `application author` is choosing to instantiate. I will think about that some more. I'm not sure how much of a problem it is.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
justinabrahms: Yep. I'm all for this idea. We should be able to swap this in the future in a backwards compatible way, provided that the underlying context is of the EvaluationContext type.

justinabrahms (approved):

Issue #83, opened 2022-09-30
Fix security posture for clomonitor


justinabrahms:
```

CLOMonitor linter results

Repository information

╭────────────┬─────────────────────────────────────────────────────╮
│ Local path ┆ /Users/jabrahms/src/github.com/open-feature/javasdk │
├╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ Remote url ┆ https://github.com/open-feature/java-sdk            │
├╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ Check sets ┆ [Code]                                              │
╰────────────┴─────────────────────────────────────────────────────╯

Score summary

╭────────────────┬───────╮
│     Section    ┆ Score │
╞════════════════╪═══════╡
│ Global         ┆   80  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ Documentation  ┆   78  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ License        ┆  100  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ Best practices ┆   94  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ Security       ┆   50  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ Legal          ┆  n/a  │
╰────────────────┴───────╯

Checks summary

╭──────────────────────────────────────┬────────────╮
│                 Check                ┆   Passed   │
╞══════════════════════════════════════╪════════════╡
│ Documentation / Adopters             ┆     n/a    │
[...]

Issue #55, opened 2022-08-30
generate SBOMs


justinabrahms:
We should generate sboms in cyclonedx and spdx format which are uploaded into maven central.

beeme1mr:
Creation of SBOM was added in #119

PR #119, opened 2022-10-05
feat: Support for generating CycloneDX sboms

justinabrahms:
Ref #83

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,32 @@
     <plugins>
 
       <plugin>
+        <groupId>org.cyclonedx</groupId>
+        <artifactId>cyclonedx-maven-plugin</artifactId>
+        <version>2.7.0</version>
+        <configuration>
+          <projectType>library</projectType>
+          <schemaVersion>1.3</schemaVersion>
+          <includeBomSerialNumber>true</includeBomSerialNumber>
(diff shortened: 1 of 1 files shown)


justinabrahms:
Refs #55 (still need to do spdx to close it out, I think)

pom.xml
```diff
@@ -164,6 +164,32 @@
...
+        <artifactId>cyclonedx-maven-plugin</artifactId>
+        <version>2.7.0</version>
+        <configuration>
+          <projectType>library</projectType>
+          <schemaVersion>1.3</schemaVersion>
+          <includeBomSerialNumber>true</includeBomSerialNumber>
+          <includeCompileScope>true</includeCompileScope>
+          <includeProvidedScope>true</includeProvidedScope>
+          <includeRuntimeScope>true</includeRuntimeScope>
+          <includeSystemScope>true</includeSystemScope>
```
toddbaert, line 178: I guess it doesn't really matter, since more information is better, but I'm trying to imagine why putting dependencies with these scopes in could matter for an SBOM.

Would a consumer care that we use lombok (provided scope), or transitive dependencies coming from lombok at compile time? Maybe? More of a question that a critique.

toddbaert (approved):

PR #123, opened 2022-10-06
chore: I don't think we use that permission?

justinabrahms:
Signed-off-by: Justin Abrahms <jabrahms@ebay.com>

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -14,8 +14,6 @@ jobs:
         ports:
           - 8013:8013
 
-    permissions:
-      packages: write
     steps:
       - name: Check out the code
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b


.github/workflows/pullrequest.yml
```diff
@@ -14,8 +14,6 @@ jobs:
         ports:
           - 8013:8013
 
-    permissions:
-      packages: write
```
toddbaert, line 18: Do we need this [here](https://github.com/open-feature/java-sdk/blob/73cc784491f6dc0a371a2bc44d780fa5e3d39eaa/.github/workflows/merge.yml#L21-L22)?

toddbaert (approved):

.github/workflows/pullrequest.yml
justinabrahms: I don't think so. Added https://github.com/open-feature/java-sdk/pull/130

Issue #64, opened 2022-09-08
Review thread safety of state in API and client


toddbaert:
I'd like to review the thread safety of the SDK to make sure we don't have potential concurrency issues, particularly around state maintained in the global API object or clients. I'd like to confirm we have the required safeguards around the context, hooks, or any other state on these objects.

We may need some constraints or synchronization around setting context, hooks, etc.

Let's use this issue to investigate and discuss some of these.

**UPDATE**:  after lots of discussion [here](https://github.com/open-feature/go-sdk/issues/90) and the recent changes [here](https://github.com/open-feature/java-sdk/pull/112) to make Context an interface, I think the basic outline of what needs to be done to resolve thread safety issues is:

- create an immutable implementation of `EvaluationContext`
- create a read/write lock on the global provider, global hooks, and the client hooks that allows _n_ threads to read and 1 to write
- document thread safety in general, particularly use of the `ImmutableContext`.

cc @justinabrahms @kinyoklion @thomaspoignant

PR #131, opened 2022-10-07
fix!: add rw locks to client/api, hook accessor name

toddbaert:
This PR adds multi-read/single-write locks to both the global singleton and the client. Particularly it addresses problems relating to the providers/hooks being mutated while an evaluation is occurring. This is especially important since some providers have provider hooks, and if we dont lock, it's possible hooks from proivder1 run on provider2 if an interleaving thread switched the provider.

To make things cleaner and less error-prone, I've created a little wrapper so we can use `try-with-resources` blocks.

This way, locking can be done like:

```
    public void addHooks(Hook... hooks) {
        try (AutoCloseableLock __ = rwLock.writeLockAutoCloseable()) {
            this.apiHooks.addAll(Arrays.asList(hooks));
        }
    }
```

...with no need to explicitly close the lock.

I'm going to write some tests in the morning for this before I mark it as ready.

Closes: https://github.com/open-feature/java-sdk/issues/64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -246,7 +246,7 @@
           <argLine>
             ${surefireArgLine}
           </argLine>
-          <!-- fork a new JVM to isolate test suites, especially import with singletons -->
+          <!-- fork a new JVM to isolate test suites, especially important with singletons -->
           <forkCount>${cpu.count}</forkCount>
           <reuseForks>false</reuseForks>
           <excludes>
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,21 @@
     <plugins>
 
       <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>3.3.0</version>
+        <executions>
+          <execution>
+            <phase>validate</phase>
+            <id>get-cpu-count</id>
+            <goals>
+              <goal>cpu-count</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
         <groupId>org.cyclonedx</groupId>
         <artifactId>cyclonedx-maven-plugin</artifactId>
         <version>2.7.2</version>
@@ -231,6 +246,9 @@
           <argLine>
             ${surefireArgLine}
           </argLine>
+          <!-- fork a new JVM to isolate test suites, especially import with singletons -->
+          <forkCount>${cpu.count}</forkCount>
+          <reuseForks>false</reuseForks>
           <excludes>
             <!-- tests to exclude -->
             <exclude>${testExclusions}</exclude>
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
@@ -8,7 +8,6 @@ import javax.annotation.Nullable;
 
 import dev.openfeature.sdk.internal.AutoCloseableLock;
 import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
-import lombok.Getter;
 
 /**
  * A global singleton which holds base configuration for the OpenFeature library.
@@ -16,12 +15,11 @@ import lombok.Getter;
  */
 public class OpenFeatureAPI {
     // package-private multi-read/single-write lock
-    static AutoCloseableReentrantReadWriteLock rwLock = new AutoCloseableReentrantReadWriteLock();
-    @Getter
+    static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
+    static AutoCloseableReentrantReadWriteLock providerLock = new AutoCloseableReentrantReadWriteLock();
+    static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
     private FeatureProvider provider;
-    @Getter
     private EvaluationContext evaluationContext;
-    @Getter
     private List<Hook> apiHooks;
 
     private OpenFeatureAPI() {
@@ -60,7 +58,7 @@ public class OpenFeatureAPI {
      * {@inheritDoc}
      */
     public void setEvaluationContext(EvaluationContext evaluationContext) {
-        try (AutoCloseableLock __ = rwLock.writeLockAutoCloseable()) {
+        try (AutoCloseableLock __ = contextLock.writeLockAutoCloseable()) {
             this.evaluationContext = evaluationContext;
         }
     }
@@ -68,8 +66,17 @@ public class OpenFeatureAPI {
     /**
      * {@inheritDoc}
      */
+    public EvaluationContext getEvaluationContext() {
+        try (AutoCloseableLock __ = contextLock.readLockAutoCloseable()) {
+            return this.evaluationContext;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public void setProvider(FeatureProvider provider) {
-        try (AutoCloseableLock __ = rwLock.writeLockAutoCloseable()) {
+        try (AutoCloseableLock __ = providerLock.writeLockAutoCloseable()) {
             this.provider = provider;
         }
     }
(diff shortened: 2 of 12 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -32,12 +37,16 @@ public OpenFeatureAPI() {
      * @return The singleton instance.
      */
     public static OpenFeatureAPI getInstance() {
+        if (api != null) {
+            return api;
+        }
+
```
toddbaert, line 43: This change improves performance by only leveraging `synchronized` when we actually create the API instance, instead of EVERY call. Note that we have to check if the api is null again once we've entered the block, since another thread may have already entered and exited the block.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -57,16 +62,19 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
-        });
+
 
         FlagEvaluationDetails<T> details = null;
         List<Hook> mergedHooks = null;
         HookContext<T> hookCtx = null;
 
-        try {
+        try (AutoCloseableLock __ = OpenFeatureAPI.rwLock.readLockAutoCloseable();
+            AutoCloseableLock ___ = this.rwLock.readLockAutoCloseable()) {
```
toddbaert, line 72: We double read-lock here, bot the API and the client.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -1,43 +1,47 @@
...
      */
     public static OpenFeatureAPI getInstance() {
-        synchronized (OpenFeatureAPI.class) {
-            if (api == null) {
-                api = new OpenFeatureAPI();
-            }
-        }
-        return api;
+        return SingletonHolder.INSTANCE;
+    }
```
toddbaert, line 41: This is a more performant singleton. The `synchronized` block in the previous implementation involved inherent locking on _each call_. The "double check" singleton implementation can help here, but it also has it's downfalls.

I've used the `Bill Pugh" approach here.

This takes advantage of basic Java spec requirements that guarantee classes are loaded only once. There's no synchronization concerns here and no performance penalty.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -1,43 +1,45 @@
...
      * @return The singleton instance.
      */
     public static OpenFeatureAPI getInstance() {
-        synchronized (OpenFeatureAPI.class) {
-            if (api == null) {
-                api = new OpenFeatureAPI();
-            }
-        }
-        return api;
+        return SingletonHolder.INSTANCE;
```
toddbaert, line 42: This is a more performant singleton. The `synchronized` block in the previous implementation involved inherent locking on _each call_. The "double check" singleton implementation can help here, but it also has it's downfalls.

I've used the `Bill Pugh" approach here.

This takes advantage of basic Java spec requirements that guarantee classes are loaded only once. There's no synchronization concerns here and no performance penalty.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -1,43 +1,45 @@
...
-    @Setter
     private FeatureProvider provider;
     @Getter
     @Setter
     private EvaluationContext evaluationContext;
     @Getter
     private List<Hook> apiHooks;
 
-    public OpenFeatureAPI() {
+    private OpenFeatureAPI() {
```
toddbaert, line 29: I don't think we want anyone constructing this?

justinabrahms:
Looks good so far. Interested to see the tests.

toddbaert:
I _almost_ have some tests working - but it's Canadian Thanksgiving and I have other obligations! I will pickup next week.

pom.xml
```diff
@@ -231,6 +246,9 @@
           <argLine>
             ${surefireArgLine}
           </argLine>
+          <!-- fork a new JVM to isolate test suites, especially import with singletons -->
+          <forkCount>${cpu.count}</forkCount>
+          <reuseForks>false</reuseForks>
```
toddbaert, line 251: This isn't directly related to my change, but just addresses a nasty testing issue.

Because we have a global singleton, many tests could interfere with eachother if they interleave. For example, 1 test might run `api.clearHooks()` immediately after another one does a setup that adds a hook.

This change runs every suite in a forked JVM, which adds a bit of startup time, but gives them each their own classloader to prevent interference via the singletons.

src/test/java/dev/openfeature/sdk/LockingTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
+
+class LockingTest {
```
toddbaert, line 16: There wasn't a reasonable means of fully testing the locking with blackbox testing, so I've added this whitebox-style testing.

toddbaert:
@justinabrahms I've added tests.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -57,34 +73,45 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
-
-            mergedHooks = ObjectUtils.merge(provider.getProviderHooks(), flagOptions.getHooks(), clientHooks,
-                    openfeatureApi.getApiHooks());
+            EvaluationContext apiContext;
+            EvaluationContext clientContext;
+
+            // lock while getting the provider and hooks
+            // the retrieval of any mutable state on client/API MUST be done in this block.
+            try (AutoCloseableLock __ = OpenFeatureAPI.rwLock.readLockAutoCloseable();
+                    AutoCloseableLock ___ = this.rwLock.readLockAutoCloseable()) {
```
toddbaert, line 90: I hope this comment makes sense.

skyerus (dismissed):

skyerus (dismissed):

toddbaert:
I've added a bit more granularity to the locking.

toddbaert:
@justinabrahms @agentgonzo 

I think I'm finished with this now. It's analogous to https://github.com/open-feature/dotnet-sdk/pull/79 with a few java differences changes.

Initially I was locking in the evaluation, but I don't think that's necessary or beneficial. The current locks mainly protect from errors due to out of order issues and collection mutation during enumeration.

Additionally, I've made changes to access the provider just once in the evaluation logic. This prevents things from becoming incoherent if the provider is charged mid evaluation.

src/test/java/dev/openfeature/sdk/LockingTest.java
```diff
@@ -0,0 +1,141 @@
...
+        OpenFeatureAPI.providerLock = apiProviderLock;
+        OpenFeatureAPI.hooksLock = apiHooksLock;
+
+        clientContextLock = setupLock(clientContextLock, mockInnerReadLock(), mockInnerWriteLock());
+        clientHooksLock = setupLock(clientHooksLock, mockInnerReadLock(), mockInnerWriteLock());
+        client.contextLock = clientContextLock;
+        client.hooksLock = clientHooksLock;
+    }
+
+    // @Test
```
justinabrahms, line 48: remove?

justinabrahms (approved):

src/test/java/dev/openfeature/sdk/LockingTest.java
toddbaert: Oof, yep

toddbaert: done

toddbaert:
@justinabrahms one other thing worth pointing out ... I've changed the names of the accessors for hooks on the client and API, in order to be more consistent with addhooks, which is the same for both the API and client.

If you disagree with that, I can revert.

justinabrahms:
@toddbaert Seems fine to me, but I changed your PR title to denote the breaking change.

justinabrahms (approved):

PR #142, opened 2022-10-10
Remove dependabot. I like renovate better

justinabrahms:


toddbaert:
I'm certainly OK with this, seems like we don't need both and I agree with your preference.

Looks like the build is failing due to some invalid substitution.

toddbaert (dismissed):

beeme1mr:
Yeah, I agree. The killer feature is being able to bundle multiple, related dependencies together in a single PR>

beeme1mr (approved):

toddbaert (approved):

PR #178, opened 2022-11-15
chore: fix issues with compilation on later JDKs

toddbaert:
This just fixes an issue that prevents JDK 11+ from working with our build and adds some associated missing javadoc. This is troublesome if you happen to be working with other projects that require a later version of the JDK.

Specifically, there were changes in later versions of `javac` that applied more rules, mostly around javadoc. These caused errors we could not customize, so I've disabled the lint checks in the `maven-javadoc-plugin` and I instead enforce them with `checkstyle`, which is more customizable.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -34,6 +34,7 @@
         <property name="optional" value="true"/>
     </module>
 
+
     <!-- Checks for whitespace                               -->
     <!-- See http://checkstyle.org/config_whitespace.html -->
     <module name="FileTabCharacter">
@@ -46,7 +47,23 @@
         <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
     </module>
 
+    <module name="SuppressWarningsFilter" />
+
     <module name="TreeWalker">
+        <!-- needed for SuppressWarningsFilter -->
+        <module name="SuppressWarningsHolder" />
+        
+        <module name="SuppressWarnings">
+            <property name="id" value="checkstyle:suppresswarnings"/>
+        </module>
+
+        <!-- https://checkstyle.org/config_filters.html#SuppressionXpathFilter -->
+        <module name="SuppressionXpathFilter">
+            <property name="file" value="${org.checkstyle.google.suppressionxpathfilter.config}"
+                        default="checkstyle-xpath-suppressions.xml" />
+            <property name="optional" value="true"/>
+        </module>
+    
         <module name="OuterTypeFilename"/>
         <module name="IllegalTokenText">
             <property name="tokens" value="STRING_LITERAL, CHAR_LITERAL"/>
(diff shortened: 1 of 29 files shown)


beeme1mr (dismissed):

src/main/java/dev/openfeature/sdk/MutableStructure.java
```diff
@@ -34,55 +37,103 @@ public Set<String> keySet() {
...
-    // getters
+
     @Override
     public Value getValue(String key) {
         return this.attributes.get(key);
     }
 
-    // adders
+    /**
+     * Adds the specified value at key.
```
justinabrahms, line 47: These docs feel kinda useless?

justinabrahms (commented):
If you're somehow blocked on jdk versions.. I'd like to separate that from the documentation changes. I think the documentation changes are of low value (no offense intended) and I don't really want to accept them. No one needs to know that `collection.add(thing)` adds the thing to the collection. It's apparent on it's face.

src/main/java/dev/openfeature/sdk/MutableStructure.java
toddbaert: > If you're somehow blocked on jdk versions.. I'd like to separate that from the documentation changes. I think the documentation changes are of low value (no offense intended) and I don't really want to accept them. No one needs to know that collection.add(thing) adds the thing to the collection. It's apparent on it's face.

I see your point. While these ones in particular aren't super valuable, even in this PR there are some that are (I think some of the exceptions are nice, for instance).

I guess the questions is, would you rather see us add this rule, and ignore it where we think it's silly, or not add the rule, and be careful to add doc where users might be confused.

justinabrahms: My preference would be to enable the rule and explicitly say "skip this one" on the things that are very obvious to keep the tool happy.

toddbaert: ACK!

toddbaert: Hey @justinabrahms I've done this.

I removed the lowest value javadocs (like the ones for the adders) and added the capability to suppress checkstyle warnings with annotations for those instances.

checkstyle.xml
```diff
@@ -46,7 +47,23 @@
...
 
+    <module name="SuppressWarningsFilter" />
+
     <module name="TreeWalker">
+        <!-- needed for SuppressWarningsFilter -->
+        <module name="SuppressWarningsHolder" />
+        
+        <module name="SuppressWarnings">
+            <property name="id" value="checkstyle:suppresswarnings"/>
+        </module>
```
toddbaert, line 58: New stuff to support suppression via annotations.

src/main/java/dev/openfeature/sdk/Reason.java
```diff
@@ -1,5 +1,8 @@
 package dev.openfeature.sdk;
 
+/**
+ * Predefined resolution reasons.
+ */
```
toddbaert, line 5: This one it debatable but I wanted to make it clear that you're not limited to just these reasons, these are just some we pre-defined.

justinabrahms (approved):

PR #210, opened 2022-12-25
feat:  added implementation of immutable evaluation context

### LINKED ISSUE #161: Create immutable EvaluationContext implementation
Currently, we have a single implementation of `EvaluationContext`: `MutableContext`. An immutable implementation would be ideal and should probably be used for the few cases we instantiate an EvaluationContext internally (within the SDK code). 

- create new `ImmutableContext` (take a look at [the context in C#](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Model/EvaluationContext.cs))
- use this immutable implementation anywhere the SDK currently does `new MutableContext()`.

thiyagu06:
Signed-off-by: thiyagu06 <thiyagu103@gmail.com>

## This PR

added implementation of immutable evaluation context

### Related Issues

Fixes #161

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -1,0 +1,112 @@
+package dev.openfeature.sdk;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+import lombok.experimental.Delegate;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The EvaluationContext is a container for arbitrary contextual data
+ * that can be used as a basis for dynamic evaluation.
+ * The MutableContext is an EvaluationContext implementation which is not threadsafe, and whose attributes can
+ * be modified after instantiation.
+ */
+@ToString
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public class ImmutableContext implements EvaluationContext {
+
+    @Getter
+    private String targetingKey;
+    @Delegate(excludes = HideDelegateAddMethods.class)
+    private final MutableStructure structure;
+
+    public ImmutableContext() {
+        this.structure = new MutableStructure();
+        this.targetingKey = "";
+    }
+
+    public ImmutableContext(Map<String, Value> attributes) {
+        HashMap<String, Value> copy = new HashMap<>(attributes);
+        this.structure = new MutableStructure(copy);
+        this.targetingKey = "";
+    }
+
+    public ImmutableContext(String targetingKey, Map<String, Value> attributes) {
+        this(attributes);
+        this.targetingKey = targetingKey;
+    }
+
+    @Override
+    public void setTargetingKey(String targetingKey) {
+        throw new UnsupportedOperationException("changing of targeting key is not allowed");
+    }
+
+    /**
+     * Merges this EvaluationContext objects with the second overriding the this in
+     * case of conflict.
+     *
+     * @param overridingContext overriding context
+     * @return resulting merged context
+     */
+    @Override
+    public EvaluationContext merge(EvaluationContext overridingContext) {
+        if (overridingContext == null) {
+            return new ImmutableContext(this.asMap());
+        }
+        String targetingKey = "";
+        if (this.getTargetingKey() != null && !this.getTargetingKey().trim().equals("")) {
+            targetingKey = this.getTargetingKey();
+        }
+
+        if (overridingContext.getTargetingKey() != null && !overridingContext.getTargetingKey().trim().equals("")) {
+            targetingKey = overridingContext.getTargetingKey();
+        }
+        Map<String, Value> merged = new HashMap<>();
+
+        merged.putAll(this.asMap());
+        merged.putAll(overridingContext.asMap());
+        return new ImmutableContext(targetingKey, merged);
+    }
+
+    /**
+     * Hidden class to tell Lombok not to copy these methods over via delegation.
+     */
+    private static class HideDelegateAddMethods {
+        public MutableStructure add(String ignoredKey, Boolean ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, Double ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, String ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, Value ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, Integer ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, List<Value> ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, MutableStructure ignoredValue) {
+            return null;
+        }
+
+        public MutableStructure add(String ignoredKey, Instant ignoredValue) {
+            return null;
+        }
+    }
+}
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -23,14 +23,23 @@ public final class ImmutableContext implements EvaluationContext {
     private final Structure structure;
 
     /**
-     * create an immutable context with an empty targeting_key and attributes provided.
+     * Create an immutable context with an empty targeting_key and attributes provided.
      */
     public ImmutableContext() {
         this("", new HashMap<>());
     }
 
     /**
-     * create an immutable context with an attributes provided.
+     * Create an immutable context with given targeting_key provided.
+     *
+     * @param targetingKey targeting key
+     */
+    public ImmutableContext(String targetingKey) {
+        this(targetingKey, new HashMap<>());
+    }
+
+    /**
+     * Create an immutable context with an attributes provided.
      *
      * @param attributes evaluation context attributes
      */
@@ -39,7 +48,7 @@ public final class ImmutableContext implements EvaluationContext {
     }
 
     /**
-     * create an immutable context with given targetingKey and attributes provided.
+     * Create an immutable context with given targetingKey and attributes provided.
      *
      * @param targetingKey targeting key
      * @param attributes   evaluation context attributes
@@ -55,7 +64,7 @@ public final class ImmutableContext implements EvaluationContext {
     }
 
     /**
-     * Merges this EvaluationContext objects with the second overriding the this in
+     * Merges this EvaluationContext object with the passed EvaluationContext, with the passed EvaluationContext overriding in
      * case of conflict.
      *
      * @param overridingContext overriding context
@@ -64,7 +73,7 @@ public final class ImmutableContext implements EvaluationContext {
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
         if (overridingContext == null) {
-            return new ImmutableContext(this.asMap());
+            return new ImmutableContext(this.targetingKey, this.asMap());
         }
         String targetingKey = "";
         if (this.getTargetingKey() != null && !this.getTargetingKey().trim().equals("")) {
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -1,11 +1,11 @@
 package dev.openfeature.sdk;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import lombok.Getter;
 import lombok.ToString;
 import lombok.experimental.Delegate;
-
-import java.util.HashMap;
-import java.util.Map;
 
 /**
  * The EvaluationContext is a container for arbitrary contextual data
@@ -58,7 +58,11 @@ public final class ImmutableContext implements EvaluationContext {
         this.targetingKey = targetingKey;
     }
 
+    /**
+     * Mutating targeting key is not supported in ImmutableContext and will be removed.
+     */
     @Override
+    @Deprecated
     public void setTargetingKey(String targetingKey) {
         throw new UnsupportedOperationException("changing of targeting key is not allowed");
     }
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -1,33 +1,32 @@
 package dev.openfeature.sdk;
 
 import lombok.Getter;
-import lombok.Setter;
 import lombok.ToString;
 import lombok.experimental.Delegate;
 
-import java.time.Instant;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 /**
  * The EvaluationContext is a container for arbitrary contextual data
  * that can be used as a basis for dynamic evaluation.
- * The MutableContext is an EvaluationContext implementation which is not threadsafe, and whose attributes can
- * be modified after instantiation.
+ * The ImmutableContext is an EvaluationContext implementation which is threadsafe, and whose attributes can
+ * not be modified after instantiation.
  */
 @ToString
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
-public class ImmutableContext implements EvaluationContext {
+public final class ImmutableContext implements EvaluationContext {
 
     @Getter
-    private String targetingKey;
-    @Delegate(excludes = HideDelegateAddMethods.class)
-    private final MutableStructure structure;
+    private final String targetingKey;
+    @Delegate
+    private final Structure structure;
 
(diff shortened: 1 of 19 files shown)


thiyagu06:
Couple of question to be confirmed.

1. Should we replace all the reference of mutableContext to ImmutableContext in tests?
2. Should ImmutableContext support merge method?

toddbaert:
> 1. Should we replace all the reference of mutableContext to ImmutableContext in tests?

I think it doesn't matter so much in tests, but I would probably prefer it there.

What matters even more is that the (few) places where the SDK instantiates a context currently, it should be switched to the mutable context (where we currently create context in the SDK, they are not mutated)

> 2. Should ImmutableContext support merge method?

Yes, I think it needs to because the SDK calls `merge` internally. It should return a new immutable context.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,123 @@
...
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The EvaluationContext is a container for arbitrary contextual data
+ * that can be used as a basis for dynamic evaluation.
+ * The MutableContext is an EvaluationContext implementation which is not threadsafe, and whose attributes can
```
Kavindu-Dodan, line 16:
```suggestion
 * The ImmutableContext is an EvaluationContext implementation which is not threadsafe, and whose attributes cannot
```

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,123 @@
...
+ * be modified after instantiation.
+ */
+@ToString
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public class ImmutableContext implements EvaluationContext {
```
Kavindu-Dodan, line 21: ImmutableContext must be `final`.

You will need to change constructors to support this requirement.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Getter
+    private String targetingKey;
+    @Delegate(excludes = HideDelegateAddMethods.class)
+    private final MutableStructure structure;
+
+    public ImmutableContext() {
+        this.structure = new MutableStructure();
```
Kavindu-Dodan, line 29: I am not convinced on using `MutableStructure` for true immutability. For example, usage of `MutableStructure(Map<String, Value> attributes)` is still mutable as we do not perform a deep copy. 

I think we need an `ImutableStructure.java` implementation here. 

@toddbaert  any opinion on this ?

toddbaert: I agree. I think we should create both an `ImutableStructure` an extend that for `ImmutableContext`

beeme1mr:
Hey @thiyagu06, thanks for your contribution. When you have a moment, please address the comments above and mark them as resolved. It would be great to include this change in an upcoming release. Thanks!

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,83 @@
...
+@SuppressWarnings("PMD.BeanMembersShouldSerialize")
+public final class ImmutableContext implements EvaluationContext {
+
+    @Getter
+    private final String targetingKey;
+    @Delegate
+    private final Structure structure;
+
+    /**
+     * create an immutable context with an empty targeting_key and attributes provided.
```
toddbaert, line 26:
```suggestion
     * Create an immutable context with an empty targeting_key and attributes provided.
```

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public void setTargetingKey(String targetingKey) {
+        throw new UnsupportedOperationException("changing of targeting key is not allowed");
+    }
+
+    /**
+     * Merges this EvaluationContext objects with the second overriding the this in
+     * case of conflict.
```
toddbaert, line 59:
```suggestion
     * Merges this EvaluationContext object with the passed EvaluationContext, with the passed EvaluationContext overriding in
     * case of conflict.
```

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,83 @@
...
+            targetingKey = this.getTargetingKey();
+        }
+
+        if (overridingContext.getTargetingKey() != null && !overridingContext.getTargetingKey().trim().equals("")) {
+            targetingKey = overridingContext.getTargetingKey();
+        }
+        Map<String, Value> merged = new HashMap<>();
+
+        merged.putAll(this.asMap());
+        merged.putAll(overridingContext.asMap());
```
toddbaert, line 80: This merge isn't recursive (but neither is the one in the existing `MutableContext`. I wonder if we should create a separate issue to discuss that? I think I would expect it to be.

@Kavindu-Dodan @justinabrahms @beeme1mr

Kavindu-Dodan: If the overriding context is **immutable**, then this is handled correctly (`asMap` performs a deep copy ). But, if the overriding context is **mutable**, then we pass a map with mutable values (no deep copy). 

So yes, we need an improvement here. 

I think logic get easier if we override `clone` method from `Value.java` and replace it with `ImmutableStructure.copy` usages

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Perform deep copy of value object.
+     *
+     * @param value value object
+     * @return new copy of the given value object
+     */
+    @SneakyThrows
+    private Value copy(Value value) {
```
Kavindu-Dodan, line 97: Nice work 👏 

IMO, this best fits `Value.java` through its `clone` override.  And this means `Value.java` becomes clonable and clone method performs this deep copying internally.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public ImmutableContext(String targetingKey, Map<String, Value> attributes) {
+        this.structure = new ImmutableStructure(attributes);
+        this.targetingKey = targetingKey;
+    }
+
+    @Override
+    public void setTargetingKey(String targetingKey) {
+        throw new UnsupportedOperationException("changing of targeting key is not allowed");
+    }
```
toddbaert, line 55: @justinabrahms @Kavindu-Dodan 

This is unfortunate. I don' think the `EvaluationContext` interface should have `setTargetingKey`, since that makes it mutable. I would like to remove it from that interface, but that would be a breaking change (though minor). How do you all feel about that?

Technically the "evaluation context" aspects of the spec are still experimental, so we could implement this breaking change with only a minor version update, as per our documentation, but I would require changes from authors in cases they declare `EvaluationContext`s and then call this method (they would have to update to use the subclass).

@justinabrahms @beeme1mr @Kavindu-Dodan @thiyagu06

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -0,0 +1,83 @@
...
+     * create an immutable context with given targetingKey and attributes provided.
+     *
+     * @param targetingKey targeting key
+     * @param attributes   evaluation context attributes
+     */
+    public ImmutableContext(String targetingKey, Map<String, Value> attributes) {
+        this.structure = new ImmutableStructure(attributes);
+        this.targetingKey = targetingKey;
+    }
+
```
toddbaert, line 51: @thiyagu06  For consistency with `MutableContext`, we may want another constructor that accepts only a `targetingKey`.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
toddbaert: I'm not talking just about deep copying, I'm talking about recursive merge.

If there is a property on `this` called `myKey`, and it has subkeys, and there's also a property called `myKey` on `overridingContext`, then the subkeys are lost, instead of merged, I think. I think that's also true for the existing implementation, though, so it's not really an issue with this PR.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
thiyagu06: In that should we make all the properties inside `Value.java` to implement clone method? Is that correct? I

thiyagu06:
Not relevant to this PR, But I see the spec doesn't include evaluation context should support List type. But I see the support of List type in SDK. Do we need to update the spec?

https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Value.java#L42

https://docs.openfeature.dev/docs/specification/sections/evaluation-context#requirement-312

toddbaert:
> Not relevant to this PR, But I see the spec doesn't include evaluation context should support List type. But I see the support of List type in SDK. Do we need to update the spec?
> 
> https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Value.java#L42
> 
> https://docs.openfeature.dev/docs/specification/sections/evaluation-context#requirement-312

The spec references `structure`, which is an intentionally vague term that we define in our [types](https://docs.openfeature.dev/docs/specification/types#structure). The structure may support lists or not, that's up to the implementation. This is an area we want to maintain as much flexibility as possible.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -0,0 +1,115 @@
...
+        if (value.isList()) {
+            List<Value> copy = value.asList().stream().map(Value::new).collect(Collectors.toList());
```
thiyagu06, line 99: Shall we wrap this copied value into  `Collections.singletonList()`?

Kavindu-Dodan: No we can't use a singleton list here. Singleton list is for a single element list and here we need the collection. 

IMO, as long as we do not expose modifier functions, current implementation is safe enough.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
toddbaert: I think I agree with this.

toddbaert: I will open a new issue for this, then resolve this conversation.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
Kavindu-Dodan: In theory yes, all fields should have their own clone overrides. But given that `Value` only holds `Object innerObject`, you have to utilize the same logic you have right here with copy - Check Object type and perform copy (clone) operation and store it in the new Value clone. I hope you understand the suggestion :) ?

src/main/java/dev/openfeature/sdk/ImmutableContext.java
toddbaert: I've created [this issue](https://github.com/open-feature/java-sdk/issues/253#issue-1538452873).

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
thiyagu06: Yes. Thanks for the explanation.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
justinabrahms: Please file a ticket. Sounds like a bug.

Kavindu-Dodan:
@thiyagu06  I see some of the conversations were marked "resolved" (ex:- https://github.com/open-feature/java-sdk/pull/210#discussion_r1071509885). But I am not seeing a new commit in the PR. 

Could you please check whether you forgot to push your changes ☺️  ?

src/main/java/dev/openfeature/sdk/Value.java
```diff
@@ -14,7 +17,7 @@
 @ToString
 @EqualsAndHashCode
 @SuppressWarnings({"PMD.BeanMembersShouldSerialize", "checkstyle:MissingJavadocType"})
-public class Value {
+public class Value implements Cloneable{
```
Kavindu-Dodan, line 20:
```suggestion
public class Value implements Cloneable {
```

This is the reason for pr pipeline failure.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -0,0 +1,89 @@
+package dev.openfeature.sdk;
+
+import lombok.EqualsAndHashCode;
```
Kavindu-Dodan, line 3: There are some unused imports. Please remove them.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -0,0 +1,89 @@
...
+    public ImmutableStructure(Map<String, Value> attributes) {
+        Map<String, Value> copy = attributes.entrySet()
+                .stream()
+                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().clone()));
+        this.attributes = new HashMap<>(copy);
+    }
+
+    @Override
+    public Set<String> keySet() {
+        return this.attributes.keySet();
```
Kavindu-Dodan, line 48: This keyset is backed by the map and changes are reflected both ways (source - java doc on keySet())

So we need to perform a deep copy operation here.

Kavindu-Dodan:
> > @thiyagu06 I see some of the conversations were marked "resolved" (ex:- [#210 (comment)](https://github.com/open-feature/java-sdk/pull/210#discussion_r1071509885)). But I am not seeing a new commit in the PR.
> > Could you please check whether you forgot to push your changes ☺️ ?
> 
> Yes. I forgot 🤦‍♂️ . Pushed it now. thanks

No worries.

I found few more improvements and once they are done, PR should be good to go :)

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
thiyagu06: nice catch. I fixed it 👍

Kavindu-Dodan (approved):
Changes looks good to me

src/main/java/dev/openfeature/sdk/ImmutableContext.java
Kavindu-Dodan: @toddbaert  can you create a follow-up issue for this ?

toddbaert: Created https://github.com/open-feature/java-sdk/issues/270

toddbaert:
@Kavindu-Dodan @thiyagu06 The only thing I'm wondering here is if it's a good idea to throw in [setTargetingKey](https://github.com/open-feature/java-sdk/pull/210/files#diff-3b1bb0fba16899446d6f2ef99d562c9072f6344156ed7f243910bb8fa27d8f57R62) or not. We could make it simply no-op and log a warning instead.

Either way, I think we should add a deprecated annotation and say it will be removed in the future. cc @justinabrahms

Kavindu-Dodan:
> @Kavindu-Dodan @thiyagu06 The only thing I'm wondering here is if it's a good idea to throw in [setTargetingKey](https://github.com/open-feature/java-sdk/pull/210/files#diff-3b1bb0fba16899446d6f2ef99d562c9072f6344156ed7f243910bb8fa27d8f57R62) or not. We could make it simply no-op and log a warning instead.
> 
> Either way, I think we should add a deprecated annotation and say it will be removed in the future. cc @justinabrahms @agentgonzo

In my view, throwing the exception is correct for the current implementation. The log could be missed and caller could be mislead as there is no reaction.

toddbaert:
> > @Kavindu-Dodan @thiyagu06 The only thing I'm wondering here is if it's a good idea to throw in [setTargetingKey](https://github.com/open-feature/java-sdk/pull/210/files#diff-3b1bb0fba16899446d6f2ef99d562c9072f6344156ed7f243910bb8fa27d8f57R62) or not. We could make it simply no-op and log a warning instead.
> > Either way, I think we should add a deprecated annotation and say it will be removed in the future. cc @justinabrahms @agentgonzo
> 
> In my view, throwing the exception is correct for the current implementation. The log could be missed and caller could be mislead as there is no reaction.

OK I've added an `@Deprecated` annotation and a comment.

toddbaert (approved):
Thanks again for your contribution @thiyagu06 !

PR #280, opened 2023-02-06
fix: MutableContext and ImmutableContext merge are made recursive

### LINKED ISSUE #270: Evaluation context `merge` implementations should be recursive
If there is a property on this called `myKey`, and it has subkeys, and there's also a property called `myKey` on overridingContext, then the subkeys are lost, instead of merged.

This is un-intuitive and should be correct.

see: https://github.com/open-feature/java-sdk/pull/210/files#r1071367173

javicv:
Signed-off-by: Javier Collado <javicv@gmail.com>

## This PR

- Makes the evaluation context merge recursive for both MutableContext and ImmutableContext

### Related Issues

Fixes #270

### How to test

- A new test method _mergeShouldRetainItsSubkeysWhenOverridingContextHasTheSameKey_ in _dev.openfeature.sdk.ImmutableContextTest_ has been added

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -88,33 +88,10 @@ public final class ImmutableContext implements EvaluationContext {
             newTargetingKey = overridingContext.getTargetingKey();
         }
 
-        Map<String, Value> merged = this.merge(this.asMap(), overridingContext.asMap());
+        Map<String, Value> merged = this.merge(m -> new ImmutableStructure(m),
+                                               this.asMap(),
+                                               overridingContext.asMap());
 
         return new ImmutableContext(newTargetingKey, merged);
-    }
-
-    /**
-     * Recursively merges the base map from this EvaluationContext with the passed EvaluationContext.
-     * 
-     * @param base base map to merge
-     * @param overriding overriding map to merge
-     * @return resulting merged map
-     */
-    private Map<String, Value> merge(Map<String, Value> base, Map<String, Value> overriding) {
-        Map<String, Value> merged = new HashMap<>();
-
-        merged.putAll(base);
-        for (Entry<String, Value> overridingEntry : overriding.entrySet()) {
-            String key = overridingEntry.getKey();
-            if (overridingEntry.getValue().isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
-                Structure mergedValue = merged.get(key).asStructure();
-                Structure overridingValue = overridingEntry.getValue().asStructure();
-                Map<String, Value> newMap = this.merge(mergedValue.asMap(), overridingValue.asMap());
-                merged.put(key, new Value(new ImmutableContext(newMap)));
-            } else {
-                merged.put(key, overridingEntry.getValue());
-            }
-        }
-        return merged;
     }
 }
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -2,7 +2,6 @@ package dev.openfeature.sdk;
 
 import java.util.HashMap;
 import java.util.Map;
-
 import lombok.Getter;
 import lombok.ToString;
 import lombok.experimental.Delegate;
@@ -86,10 +85,33 @@ public final class ImmutableContext implements EvaluationContext {
         if (overridingContext.getTargetingKey() != null && !overridingContext.getTargetingKey().trim().equals("")) {
             newTargetingKey = overridingContext.getTargetingKey();
         }
+
+        Map<String, Value> merged = this.merge(this.asMap(), overridingContext.asMap());
+
+        return new ImmutableContext(newTargetingKey, merged);
+    }
+
+    /**
+     * Recursively merges the base map from this EvaluationContext with the passed EvaluationContext.
+     * 
+     * @param base base map to merge
+     * @param overriding overriding map to merge
+     * @return resulting merged map
+     */
+    private Map<String, Value> merge(Map<String, Value> base, Map<String, Value> overriding) {
         Map<String, Value> merged = new HashMap<>();
 
-        merged.putAll(this.asMap());
-        merged.putAll(overridingContext.asMap());
-        return new ImmutableContext(newTargetingKey, merged);
+        merged.putAll(base);
+        for (String key : overriding.keySet()) {
+            if (overriding.get(key).isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
+                Structure mergedValue = merged.get(key).asStructure();
+                Structure overridingValue = overriding.get(key).asStructure();
+                Map<String, Value> newMap = this.merge(mergedValue.asMap(), overridingValue.asMap());
+                merged.put(key, new Value(new ImmutableContext(newMap)));
+            } else {
+                merged.put(key, overriding.get(key));
+            }
+        }
+        return merged;
     }
 }
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -2,6 +2,8 @@ package dev.openfeature.sdk;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
+
 import lombok.Getter;
 import lombok.ToString;
 import lombok.experimental.Delegate;
@@ -101,15 +103,15 @@ public final class ImmutableContext implements EvaluationContext {
     private Map<String, Value> merge(Map<String, Value> base, Map<String, Value> overriding) {
         Map<String, Value> merged = new HashMap<>();
 
-        merged.putAll(base);
-        for (String key : overriding.keySet()) {
-            if (overriding.get(key).isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
+        for (Entry<String, Value> overridingEntry : overriding.entrySet()) {
+            String key = overridingEntry.getKey();
+            if (overridingEntry.getValue().isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
                 Structure mergedValue = merged.get(key).asStructure();
-                Structure overridingValue = overriding.get(key).asStructure();
+                Structure overridingValue = overridingEntry.getValue().asStructure();
                 Map<String, Value> newMap = this.merge(mergedValue.asMap(), overridingValue.asMap());
                 merged.put(key, new Value(new ImmutableContext(newMap)));
             } else {
-                merged.put(key, overriding.get(key));
+                merged.put(key, overridingEntry.getValue());
             }
         }
         return merged;
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -2,7 +2,6 @@ package dev.openfeature.sdk;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Map.Entry;
 
 import lombok.Getter;
 import lombok.ToString;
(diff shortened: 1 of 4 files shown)


toddbaert:
CI is passing now. I'll take a closer look at this today/tomorrow.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -86,10 +87,34 @@ public EvaluationContext merge(EvaluationContext overridingContext) {
...
+        Map<String, Value> merged = new HashMap<>();
+
+        merged.putAll(base);
+        for (Entry<String, Value> overridingEntry : overriding.entrySet()) {
+            String key = overridingEntry.getKey();
+            if (overridingEntry.getValue().isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
+                Structure mergedValue = merged.get(key).asStructure();
+                Structure overridingValue = overridingEntry.getValue().asStructure();
+                Map<String, Value> newMap = this.merge(mergedValue.asMap(), overridingValue.asMap());
+                merged.put(key, new Value(new ImmutableContext(newMap)));
```
toddbaert, line 113: I think we want a new `Structure` here, not a new `Context`.

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -92,22 +93,53 @@ public EvaluationContext merge(EvaluationContext overridingContext) {
...
+        Map<String, Value> merged = new HashMap<>();
+
+        merged.putAll(base);
+        for (Entry<String, Value> overridingEntry : overriding.entrySet()) {
+            String key = overridingEntry.getKey();
+            if (overridingEntry.getValue().isStructure() && merged.containsKey(key) && merged.get(key).isStructure()) {
+                Structure mergedValue = merged.get(key).asStructure();
+                Structure overridingValue = overridingEntry.getValue().asStructure();
+                Map<String, Value> newMap = this.merge(mergedValue.asMap(), overridingValue.asMap());
+                merged.put(key, new Value(new MutableContext(newMap)));
```
toddbaert, line 136: Same as [above](https://github.com/open-feature/java-sdk/pull/280/files#r1097828419), we want a new structure instance here, not an entire new context.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -86,10 +87,34 @@ public EvaluationContext merge(EvaluationContext overridingContext) {
...
     }
+
+    /**
+     * Recursively merges the base map from this EvaluationContext with the passed EvaluationContext.
+     * 
+     * @param base base map to merge
+     * @param overriding overriding map to merge
+     * @return resulting merged map
+     */
+    private Map<String, Value> merge(Map<String, Value> base, Map<String, Value> overriding) {
```
toddbaert, line 103: Both this, and the similar method in `MutableContext` seem very similar. I think you can extract them both to a common protected default implementation, and somehow extract or economize the instantiation [here](https://github.com/open-feature/java-sdk/pull/280/files#diff-3b1bb0fba16899446d6f2ef99d562c9072f6344156ed7f243910bb8fa27d8f57R113), which seems to be the only difference. You could perhaps use an abstract method for returning a new instance, or pass a lambda or something like that.

toddbaert (changes requested):
Thanks so much @javicv . I think this is good conceptually. See comments in private merge methods.

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -64,4 +65,27 @@ void mergeShouldReturnAllTheValuesFromTheContextWhenOverridingContextIsNull() {
         assertEquals("targeting_key", merge.getTargetingKey());
         assertArrayEquals(new Object[]{"key1", "key2"}, merge.keySet().toArray());
     }
+
+    @DisplayName("Merge should retain subkeys from the existing context when the overriding context has the same targeting key")
+    @Test
+    void mergeShouldRetainItsSubkeysWhenOverridingContextHasTheSameKey() {
+        HashMap<String, Value> attributes = new HashMap<>();
```
Kavindu-Dodan, line 72: Something minor about formatting. Let's have some spaces to separate initializations, method invocations and assertions. Given the length of the test case, current formating makes it little difficult to read :)

src/main/java/dev/openfeature/sdk/ImmutableContext.java
Kavindu-Dodan: Yes, this is correct & perfect suggestion.

What differ is the context type (mutable vs immutable). This can be easily handled with a functional interface [1]. And I would see this best fitting as a default function exposed from `Structure.java`

In pseudocode, 

```
// Structure.java

default <T extends Structure> void merge(Function<String, T> newFunc, ......args ){

   newStructure.apply(params);

}
```


[1] - https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html

javicv: I would say the default method with the recursive merge should be in `EvaluationContext.java` because `Structure` doesn't have `merge`.

thiyagu06: I think the recursive merge is actually merging the two structure. so I would also prefer to keeping it in the `Structure.java`.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
javicv: Changed to `ImmutableStructure`

src/main/java/dev/openfeature/sdk/ImmutableContext.java
javicv: Moved to `Structure.java` with a `Function<Map<String,Value>, T>` as first parameter.

src/main/java/dev/openfeature/sdk/Structure.java
```diff
@@ -90,4 +93,33 @@ default Object convertValue(Value value) {
         }
         throw new ValueNotConvertableError();
     }
+
+    /**
+     * Recursively merges the base map from this EvaluationContext with the passed EvaluationContext.
+     *
```
thiyagu06, line 99:
```suggestion
     * Recursively merges the base map with the overriding map.
     * 
```

toddbaert:
`Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.20.0:check (run-pmd) on project sdk: You have 3 PMD violations.`

@javicv I will re-review later today, but this looks almost perfect to me! You do have some PMD issues though. Try running `mvn verify` locally to see them (if you have issues, try using JDK 8.

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -92,18 +91,25 @@ public EvaluationContext merge(EvaluationContext overridingContext) {
...
         if (overridingContext.getTargetingKey() != null && !overridingContext.getTargetingKey().trim().equals("")) {
-            ec.setTargetingKey(overridingContext.getTargetingKey());
+            newTargetingKey = overridingContext.getTargetingKey();
+        }
+        
+        EvaluationContext ec = null;
+        if (newTargetingKey != null && !newTargetingKey.trim().equals("")) {
+            ec = new MutableContext(newTargetingKey, merged);
+        } else {
+            ec = new MutableContext(merged);
```
toddbaert, line 112: @javicv can you add one more test, similar to the other, which doesn't have a `targetingKey`, to cover this case?

thiyagu06: @toddbaert I hope we should look into this scenario or will do one more PR? https://github.com/open-feature/java-sdk/issues/270#issuecomment-1421175077

javicv: > @javicv can you add one more test, similar to the other, which doesn't have a `targetingKey`, to cover this case? I'm OK for merge after this.

New test added

> @toddbaert I hope we should look into this scenario or will do one more PR? [#270 (comment)](https://github.com/open-feature/java-sdk/issues/270#issuecomment-1421175077)

@toddbaert, as I mentioned in [this comment in #270](https://github.com/open-feature/java-sdk/issues/270#issuecomment-1422240459), the current implementation doesn't merge lists, only structures. IMHO, merging lists is a different issue, I can take care of it, but we need to clarify if we want to merge lists or simply, as it's now, override the existing list with the new one.

toddbaert: I don't think lists should be merged at all. Merging nested structures is a common use case. I would not expect lists/sets to be merged, but overwritten.

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -64,4 +65,54 @@ void mergeShouldReturnAllTheValuesFromTheContextWhenOverridingContextIsNull() {
...
+    @Test
+    void mergeShouldRetainItsSubkeysWhenOverridingContextHasNoTargetingKey() {
+        HashMap<String, Value> attributes = new HashMap<>();
+        HashMap<String, Value> key1Attributes = new HashMap<>();
+
+        key1Attributes.put("key1_1", new Value("val1_1"));
+        attributes.put("key1", new Value(new ImmutableStructure(key1Attributes)));
+        attributes.put("key2", new Value("val2"));
+        
+        EvaluationContext ctx = new ImmutableContext("targeting_key", attributes);
```
toddbaert, line 106:
```suggestion
        EvaluationContext ctx = new ImmutableContext(attributes);
```

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -64,4 +65,54 @@ void mergeShouldReturnAllTheValuesFromTheContextWhenOverridingContextIsNull() {
...
+        HashMap<String, Value> key1Attributes = new HashMap<>();
+
+        key1Attributes.put("key1_1", new Value("val1_1"));
+        attributes.put("key1", new Value(new ImmutableStructure(key1Attributes)));
+        attributes.put("key2", new Value("val2"));
+        
+        EvaluationContext ctx = new ImmutableContext(attributes);
+        EvaluationContext overriding = new ImmutableContext("");
+        EvaluationContext merge = ctx.merge(overriding);
+        assertEquals("targeting_key", merge.getTargetingKey());
```
toddbaert, line 109:
```suggestion
```

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -64,4 +65,53 @@ void mergeShouldReturnAllTheValuesFromTheContextWhenOverridingContextIsNull() {
...
+    void mergeShouldRetainItsSubkeysWhenOverridingContextHasNoTargetingKey() {
+        HashMap<String, Value> attributes = new HashMap<>();
+        HashMap<String, Value> key1Attributes = new HashMap<>();
+
+        key1Attributes.put("key1_1", new Value("val1_1"));
+        attributes.put("key1", new Value(new ImmutableStructure(key1Attributes)));
+        attributes.put("key2", new Value("val2"));
+        
+        EvaluationContext ctx = new ImmutableContext(attributes);
+        EvaluationContext overriding = new ImmutableContext("");
```
toddbaert, line 107:
```suggestion
        EvaluationContext overriding = new ImmutableContext();
```

toddbaert:
@javicv I pushed some (very) small changes to your second test that I think should resolve the coverage issue.

toddbaert (approved):
I'm a bit confused my codecov isn't seeing that one line covered, but I'm satisfied with the tests, and the delta is very small anyway.

Thanks @javicv for all your efforts! This will be really nice to have. Approved.

beeme1mr (approved):

Kavindu-Dodan (approved):

PR #289, opened 2023-02-10
chore: Add in the config for fuzzing

justinabrahms:
Not 100% sure this actually works, but it's a branch so we'll see.

Signed-off-by: Justin Abrahms <justin@abrah.ms>


beeme1mr:
@justinabrahms is this ready to be reviewed?

justinabrahms:
@beeme1mr it was blocked on corp computer shenanigans. Will try again now that those have cleaned up.

Issue #253, opened 2023-01-18
:warning: remove `setTargetingKey` from `EvaluationContext` interface?


toddbaert:
[This](https://github.com/open-feature/java-sdk/blob/d6886854176a9fc64a41bf2470bef2e869b7b917/src/main/java/dev/openfeature/sdk/EvaluationContext.java#L11) should be removed from this interface.

This would be a breaking change (though probably not a widespread one, and be simple to fix).

The problem is that even immutable implementations of `EvaluationContext` would need to implement this, which is not ideal (see an example [here](https://github.com/open-feature/java-sdk/pull/210#discussion_r1071517227)).

If we decide to do this, we also need to decide whether or not to increment the major version of the SDK. The Evaluation context is still labelled "experimental" so by the [spec support rules](https://github.com/open-feature/spec/tree/main/specification#experimental) we can have breaking changes here without incrementing the major version, though it might cause compiler errors for some authors.

Kavindu-Dodan:
@toddbaert this seems to be fixed with #290. Could we check and close this?

PR #290, opened 2023-02-10
feat: remove the deprecated setTargetingKey method in EvaluationContext.

thiyagu06:
Signed-off-by: thiyagu06 <thiyagu103@gmail.com>

## This PR

remove the deprecated setTargetingKey method in EvaluationContext.

### Related Issues

https://github.com/open-feature/java-sdk/issues/253

### DIFF
--- a/src/main/java/dev/openfeature/sdk/EvaluationContext.java
+++ b/src/main/java/dev/openfeature/sdk/EvaluationContext.java
@@ -7,12 +7,6 @@ package dev.openfeature.sdk;
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
 public interface EvaluationContext extends Structure {
     String getTargetingKey();
-    
-    /**
-     * Mutating targeting key is not supported in all implementations and will be removed.
-     */
-    @Deprecated
-    void setTargetingKey(String targetingKey);
 
     /**
      * Merges this EvaluationContext object with the second overriding the this in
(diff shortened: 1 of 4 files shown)


thiyagu06:
1. I still kept the `setTargetingKey` in the mutableContext Implementation. I feel it's appropriate for mutableContext to allow the mutation of `targetingKey`. Any thoughts?
2. Confirm that we're not going to do major release for this deprecation.

justinabrahms (approved):

justinabrahms:
@toddbaert Is it not adequate in your mind to just rely on version numbers for this?

toddbaert:
> @toddbaert Is it not adequate in your mind to just rely on version numbers for this?

We can! But I think technically, unless I'm missing something, removing this from the interface is a breaking change, so if we're being very strict it would mean this has to be a 2.0 release (we'd have to change the title to `feat! ...` to denote that, and release please would do the rest).

However, as per the spec, evaluation context is still `experimental`, and thus is subject to breaking changes on a minor release, so we have some leeway here.

It's such a small change, I don't mind either way, but I wanted to make sure we were all aware of this. I specifically wanted to get your opinion.

justinabrahms:
I'm fine with breaking this on a minor change given the experimental status.

toddbaert:
@justinabrahms cool! I'll merge then and we can release this in a bit once the deprecation warning has floated around for a while (that was released today).

toddbaert (approved):

PR #300, opened 2023-02-16
fix: Do not throw null reference exception accessing a missing item.

kinyoklion:
## This PR

I noticed when implementing a provider that accessing a missing item from an ImmutableContext throws an exception instead of returning null. I do not believe this to be the desired behavior.

### How to test

I think the added unit test should cover it.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ImmutableStructure.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableStructure.java
@@ -50,7 +50,7 @@ public final class ImmutableStructure implements Structure {
     @Override
     public Value getValue(String key) {
         Value value = this.attributes.get(key);
-        return value.clone();
+        return value != null ? value.clone() : null;
     }
 
     /**
--- a/src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
+++ b/src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
@@ -10,11 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertNotSame;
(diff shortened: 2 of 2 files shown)


src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
```diff
@@ -10,11 +10,7 @@
 import java.util.Map;
 import java.util.Set;
 
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
```
kinyoklion, line 13: IDE tidied this, but I could undo it.

toddbaert: Doesn't bother me.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -50,7 +50,7 @@ public Set<String> keySet() {
     @Override
     public Value getValue(String key) {
         Value value = this.attributes.get(key);
-        return value.clone();
+        return value != null ? value.clone() : null;
```
toddbaert, line 53: Thanks for finding/fixing/testing this. I don't think we have a similar issue in `MutableStructure` (the older implementation) because we don't clone the retrieved value.

toddbaert (approved):

src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
Kavindu-Dodan: Personally, I am not a fan of wildcard imports. And the IDE should have the option to disallow this, for example, in idea [1].

Anyway, this could be a good code quality check if agree on the standard pracitce 

[1] - https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

justinabrahms (approved):

src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
toddbaert: Feel free to add this to whatever check you please, here and in the contrib. I know some people don't like this, as long as we're consistent I don't mind.

PR #363, opened 2023-03-28
fix: added an automatic module name (#362)

### LINKED ISSUE #362: The SDK Jar does not provide a Module Name
This is a major issue for developers using both IntelliJ and Gradle due to a [Gradle bug](https://github.com/gradle/gradle/issues/17609)

But even when that bug is not present having the automatically created module name be plain `sdk` is not very descriptive and has a high potential for confusion.

This can be fixed without requiring Java 9 or higher by adding `Automatic-Module-Name` to the `MANIFEST.MF` in the published Jar.  Alternatively, the project can be updated to compile to Java 11 bytecode and be fully modularized.  Currently the contributing guidelines say `Java 11+` but the pom.xml say `1.8`.

ARobertsCollibra:
## This PR
- Adds an automatic module name that is more descriptive than `sdk`.  The new module name is `dev.openfeature.sdk`.

### Related Issues

Fixes #362  

### Notes
This is the easy fix, modularizing the entire Jar would take longer.  There might be a better way to achieve this but I am not a maven expert.

### How to test
After the change the produced Jar should have `Automatic-Module-Name: dev.openfeature.sdk` in the `META-INF/MANIFEST.MF` file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,6 +13,7 @@
     <junit.jupiter.version>5.9.2</junit.jupiter.version>
     <!-- exclusion expression for integration tests -->
     <testExclusions>**/integration/*.java</testExclusions>
+    <module-name>${groupId}.${artifactId}</module-name>
   </properties>
 
   <name>OpenFeature Java SDK</name>
@@ -356,7 +357,7 @@
         <configuration>
           <archive>
             <manifestEntries>
-              <Automatic-Module-Name>dev.openfeature.sdk</Automatic-Module-Name>
+              <Automatic-Module-Name>${module-name}</Automatic-Module-Name>
             </manifestEntries>
           </archive>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -349,6 +349,18 @@
       </plugin>
       <!-- End publish to maven central -->
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>3.3.0</version>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>dev.openfeature.sdk</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -349,6 +349,18 @@
...
       <!-- End publish to maven central -->
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>3.3.0</version>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>dev.openfeature.sdk</Automatic-Module-Name>
```
toddbaert, line 359: My understanding is that that module name should match the "highest level" package name, so this seems correct.

toddbaert (approved):

pom.xml
ARobertsCollibra: @toddbaert  Thank you for the review!  Do you know when this will get merged?  Is there a public release schedule?

beeme1mr: Hey @ARobertsCollibra, we can release right after this is merged. A new release will happen automatically after [this PR](https://github.com/open-feature/java-sdk/pull/331) is merged.

Kavindu-Dodan, line 359: Minor remark - shall we move this name to pom's properties section?  

```suggestion
              <Automatic-Module-Name>${module-name}</Automatic-Module-Name>
```

where `module-name` is in properties.

toddbaert: Ya, this is what I was describing [here](https://github.com/open-feature/java-sdk-contrib/issues/261). I think you're right that it makes sense to do it here as well.

toddbaert (approved):

pom.xml
Kavindu-Dodan: @toddbaert  @ARobertsCollibra  I added this suggestion :)

beeme1mr: This change was included in `v1.3.1`. Thanks for the help!

PR #388, opened 2023-04-11
feat: Support mapping a client to a given provider.

justinabrahms:
## This PR
This adds support for mapping a client name onto a given provider, while still providing backwards compatibility to existing clients.

### Related Issues
Refs https://github.com/open-feature/ofep/pull/56
Fixes: https://github.com/open-feature/java-sdk/issues/442

### Follow-up Tasks
I still need to clean up documentation around this to note the difference between a "default provider" and a client-specific provider.

### How to test
Not sure about testing, but I'd especially love eyes on the thread safety. I don't think ConcurrentHashMap is doing a lot of work here, given we have to use locks to support `getProviderForClientOrDefault`. If you have a better way, I'm all ears.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/MutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/MutableContext.java
@@ -4,6 +4,7 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
@@ -16,6 +17,7 @@ import lombok.experimental.Delegate;
  * be modified after instantiation.
  */
 @ToString
+@EqualsAndHashCode
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
 public class MutableContext implements EvaluationContext {
 
--- a/src/main/java/dev/openfeature/sdk/MutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/MutableContext.java
@@ -90,7 +90,7 @@ public class MutableContext implements EvaluationContext {
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
         if (overridingContext == null) {
-            return new MutableContext(this.asMap());
+            return new MutableContext(this.targetingKey, this.asMap());
         }
 
         Map<String, Value> merged = this.merge(map -> new MutableStructure(map),
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
@@ -1,9 +1,6 @@
 package dev.openfeature.sdk;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.annotation.Nullable;
@@ -18,12 +15,11 @@ import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
 public class OpenFeatureAPI {
     // package-private multi-read/single-write lock
     static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
-    static AutoCloseableReentrantReadWriteLock providerLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
     private EvaluationContext evaluationContext;
-    private List<Hook> apiHooks;
-    private Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
-    private static String DEFAULT_PROVIDER_KEY = "very-secret-string-which-you-shouldnt-use";
+    private final List<Hook> apiHooks;
+    private FeatureProvider defaultProvider = new NoOpProvider();
+    private final Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
 
     private OpenFeatureAPI() {
         this.apiHooks = new ArrayList<>();
@@ -42,13 +38,12 @@ public class OpenFeatureAPI {
     }
 
     public Metadata getProviderMetadata() {
-        return providers.get(DEFAULT_PROVIDER_KEY).getMetadata();
+        return defaultProvider.getMetadata();
     }
 
     public Metadata getProviderMetadata(String clientName) {
-        return providers.get(clientName).getMetadata();
+        return getProvider(clientName).getMetadata();
     }
-
 
     public Client getClient() {
         return getClient(null, null);
@@ -81,14 +76,13 @@ public class OpenFeatureAPI {
     }
 
     /**
-     * {@inheritDoc}
+     * Set the default provider.
      */
     public void setProvider(FeatureProvider provider) {
         if (provider == null) {
-            this.providers.remove(DEFAULT_PROVIDER_KEY);
-        } else {
-            setProvider(DEFAULT_PROVIDER_KEY, provider);
+            throw new IllegalArgumentException("Provider cannot be null");
         }
+        defaultProvider = provider;
     }
 
     /**
@@ -97,16 +91,17 @@ public class OpenFeatureAPI {
      * @param provider The provider to set.
      */
     public void setProvider(String clientName, FeatureProvider provider) {
-        try (AutoCloseableLock __ = providerLock.writeLockAutoCloseable()) {
-            this.providers.put(clientName, provider);
+        if (provider == null) {
+            throw new IllegalArgumentException("Provider cannot be null");
         }
+        this.providers.put(clientName, provider);
     }
 
     /**
-     * {@inheritDoc}
+     * Return the default provider.
      */
     public FeatureProvider getProvider() {
-        return this.providers.get(DEFAULT_PROVIDER_KEY);
+        return defaultProvider;
     }
 
     /**
@@ -115,13 +110,7 @@ public class OpenFeatureAPI {
      * @return A named {@link FeatureProvider}
      */
     public FeatureProvider getProvider(String name) {
-        try (AutoCloseableLock __ = providerLock.readLockAutoCloseable()) {
-            if (name == null) {
-                return this.providers.get(DEFAULT_PROVIDER_KEY);
-            }
-            return this.providers.getOrDefault(name, this.providers.get(DEFAULT_PROVIDER_KEY));
-        }
-        }
+        return Optional.ofNullable(name).map(this.providers::get).orElse(defaultProvider);
     }
 
 
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
@@ -87,6 +87,11 @@ public class OpenFeatureAPI {
         setProvider(DEFAULT_PROVIDER_KEY, provider);
     }
 
+    /**
+     * Add a provider for a named client.
+     * @param clientName The name of the client.
+     * @param provider The provider to set.
+     */
     public void setProvider(String clientName, FeatureProvider provider) {
         try (AutoCloseableLock __ = providerLock.writeLockAutoCloseable()) {
             this.providers.put(clientName, provider);
@@ -100,6 +105,11 @@ public class OpenFeatureAPI {
         return this.providers.get(DEFAULT_PROVIDER_KEY);
     }
 
+    /**
+     * Fetch a provider for a named client. If not found, return the default.
+     * @param name The client name to look for.
+     * @return A named {@link FeatureProvider}
+     */
     public FeatureProvider getProviderForClientOrDefault(String name) {
         try (AutoCloseableLock __ = providerLock.writeLockAutoCloseable()) {
             FeatureProvider val = this.providers.get(name);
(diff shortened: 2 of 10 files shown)


toddbaert:
> I still need to clean up documentation around this to note the difference between a "default provider" and a client-specific provider.

If you think it's worth it, it could be something we mention in the [glossary](https://openfeature.dev/specification/glossary).

justinabrahms:
Doc improvement: https://github.com/open-feature/spec/pull/186

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -88,7 +90,7 @@ public MutableContext add(String key, List<Value> value) {
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
         if (overridingContext == null) {
-            return new MutableContext(this.asMap());
+            return new MutableContext(this.targetingKey, this.asMap());
```
toddbaert, line 93: Is this a bugfix?

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -76,20 +84,49 @@ public EvaluationContext getEvaluationContext() {
...
-            return this.provider;
+        return this.providers.get(DEFAULT_PROVIDER_KEY);
+    }
+
+    /**
+     * Fetch a provider for a named client. If not found, return the default.
+     * @param name The client name to look for.
+     * @return A named {@link FeatureProvider}
+     */
+    public FeatureProvider getProviderForClientOrDefault(String name) {
```
toddbaert, line 117: Personally, I think I'd rather see an overload of `getProvider`, but that's just my preference.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -76,20 +84,49 @@ public EvaluationContext getEvaluationContext() {
      * {@inheritDoc}
      */
     public void setProvider(FeatureProvider provider) {
+        if (provider == null) {
+            this.providers.remove(DEFAULT_PROVIDER_KEY);
+        } else {
+            setProvider(DEFAULT_PROVIDER_KEY, provider);
+        }
```
toddbaert, line 91: I think we need to write lock at the start of this method. An interleaving thread could alter the provider between any of these operations.

I think this means you'll have to change the locking to be able to cal [the overload](https://github.com/open-feature/java-sdk/pull/388/files#diff-0bd159679b99d067ac5064dcff6f517c1aed1ef3d6023499260568c67dfdc035R99) from this method (since you won't be able to lock twice on the same call stack, I think, unless that's supported magically somehow).

src/main/java/dev/openfeature/sdk/MutableContext.java
justinabrahms: Yes. I was increasing test coverage and found this bug. :-/

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
justinabrahms: So my thought was that the concurrent hash map should solve that? Both remove and the underlying set should be safe, as far as I understand.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -39,9 +42,14 @@ public static OpenFeatureAPI getInstance() {
     }
 
     public Metadata getProviderMetadata() {
-        return provider.getMetadata();
+        return providers.get(DEFAULT_PROVIDER_KEY).getMetadata();
     }
 
+    public Metadata getProviderMetadata(String clientName) {
+        return providers.get(clientName).getMetadata();
+    }
```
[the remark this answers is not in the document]
justinabrahms: This seems obvious? tbh, the whole metadata thing doesn't seem super useful to me, but I assume others have a use for it. If you want to add it to the spec, I'm happy to approve.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: Ya, I think that part is safe. As written, I believe 2 threads could enter this method at the same time, and I'm worried could cause some races because the entire method is not atomic/locked. Perhaps I'm being overly cautious.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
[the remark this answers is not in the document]
toddbaert: I think it's OK as is.

src/main/java/dev/openfeature/sdk/MutableContext.java
toddbaert: Good fix. I think the only thing to be aware of is that if we fix it in this PR, it won't be captured in the release notes (release please generates the notes based on semantic commits).

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -18,9 +20,10 @@ public class OpenFeatureAPI {
     static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock providerLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
-    private FeatureProvider provider;
     private EvaluationContext evaluationContext;
     private List<Hook> apiHooks;
+    private Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
```
Kavindu-Dodan, line 25: Another alternative here is to make the default provider a non-named provider

```java
private FeatureProvider provider; // this is default provider
private Map<String, FeatureProvider> providers = new ConcurrentHashMap<>(); // these are named providers 
```
This helps to avoid `DEFAULT_PROVIDER_KEY` usage and avoid Map lookups for default usage (which is one less lookup to be performant)

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
Kavindu-Dodan: This might be easy to fix with the suggestion here [1] : to have a clear separation between default provider & named providers 

[1] - https://github.com/open-feature/java-sdk/pull/388/files#r1185472185

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -76,20 +84,49 @@ public EvaluationContext getEvaluationContext() {
      * {@inheritDoc}
      */
     public void setProvider(FeatureProvider provider) {
+        if (provider == null) {
+            this.providers.remove(DEFAULT_PROVIDER_KEY);
```
thiyagu06, line 88: any reason for removing the entry from map only for default client and not for named client when the given provider is null? If the intention is remove it from map, we should move this logic inside the `setProvider(String clientName, FeatureProvider provider)`

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -18,9 +20,10 @@ public class OpenFeatureAPI {
     static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock providerLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
-    private FeatureProvider provider;
     private EvaluationContext evaluationContext;
     private List<Hook> apiHooks;
+    private Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
+    private static String DEFAULT_PROVIDER_KEY = "very-secret-string-which-you-shouldnt-use";
```
thiyagu06, line 26: shouldn't be`DEFAULT_PROVIDER_CLIENT_NAME_KEY`? we discussed lot about this in ofep. https://github.com/open-feature/ofep/pull/56#discussion_r1142280472

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -39,9 +42,14 @@ public static OpenFeatureAPI getInstance() {
     }
 
     public Metadata getProviderMetadata() {
-        return provider.getMetadata();
+        return providers.get(DEFAULT_PROVIDER_KEY).getMetadata();
```
thiyagu06, line 45: we need a null check here

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -39,9 +42,14 @@ public static OpenFeatureAPI getInstance() {
...
     }
 
+    public Metadata getProviderMetadata(String clientName) {
```
thiyagu06, line 48: we need a null check here.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
thiyagu06: If we're going to have lock before writing into this map, `concurrentHashMap` seems to be bit overkill and we can use `HashMap` itself.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -76,20 +84,49 @@ public EvaluationContext getEvaluationContext() {
...
+    public FeatureProvider getProvider(String name) {
         try (AutoCloseableLock __ = providerLock.readLockAutoCloseable()) {
-            return this.provider;
+            if (name != null) { // this happens when a client name isn't set.
+                FeatureProvider val = this.providers.get(name);
+                if (val != null) {
+                    return val;
+                }
+            }
+            return this.providers.get(DEFAULT_PROVIDER_KEY);
```
lopitz, line 125:
```suggestion
            return Optional
                .ofNullable(name)
                .map(this.providers::get)
                .orElseGet(() -> this.providers.get(DEFAULT_PROVIDER_KEY));
```

```suggestion
            if (name == null) {
                return this.providers.get(DEFAULT_PROVIDER_KEY);
            }
            return this.providers.getOrDefault(name, this.providers.get(DEFAULT_PROVIDER_KEY));
        }
```

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -76,20 +84,49 @@ public EvaluationContext getEvaluationContext() {
...
         }
     }
 
+
```
lopitz, line 129: don't know, whether u would like to remove the `inheritDoc` lines below in this PR. as the class is neither derived (except for java.lang.Object) nor implementing an interface, it can't inherit docs (on that method)...

lopitz (changes requested):

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
justinabrahms: There is no client name for the default provider. The default provider is the default if no client name is specified.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
justinabrahms: Tracked here: https://github.com/open-feature/java-sdk/issues/424

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
justinabrahms: Alright. I've accepted @thiyagu06's great suggestion about separating out the default case. The logic is much clearer now, but I think it's possible for someone to change the default provider out from someone. That said, I don't think this is a big deal and I'm inclined to leave it. Please let me know if that concerns you.

thiyagu06: current implementation looks good for me. btw it's a suggestion from @Kavindu-Dodan 😃

thiyagu06 (approved):
Great work 👍

lopitz (approved):
LGTM, nice work! ❤️

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -16,11 +15,11 @@
 public class OpenFeatureAPI {
     // package-private multi-read/single-write lock
     static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
-    static AutoCloseableReentrantReadWriteLock providerLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
-    private FeatureProvider provider;
     private EvaluationContext evaluationContext;
-    private List<Hook> apiHooks;
+    private final List<Hook> apiHooks;
+    private FeatureProvider defaultProvider = new NoOpProvider();
```
Kavindu-Dodan, line 21: 👍  for the noop default provider. This ensures null safety where possible.

Kavindu-Dodan (approved):

beeme1mr:
@justinabrahms @toddbaert any concerns about merging this?

toddbaert:
> @justinabrahms @toddbaert any concerns about merging this?

Nope!

thomaspoignant (approved):

PR #417, opened 2023-05-02
chore: rename integration tests e2e

toddbaert:
As has been brought up a few times, "end-to-end" is probably a better term for these tests, since they startup multiple processes and do inter-process communication streams with real (not mocked) artifacts.

No functional changes. I did use git mv to try to preserve history as best as possible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
     <maven.compiler.target>${maven.compiler.source}</maven.compiler.target>
     <junit.jupiter.version>5.9.3</junit.jupiter.version>
     <!-- exclusion expression for integration tests -->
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -12,7 +12,7 @@
     <maven.compiler.target>${maven.compiler.source}</maven.compiler.target>
     <junit.jupiter.version>5.9.3</junit.jupiter.version>
     <!-- exclusion expression for integration tests -->
```
thisthat, line 14: nit
```suggestion
    <!-- exclusion expression for e2e tests -->
```

thisthat (approved):

justinabrahms (approved):

Issue #440, opened 2023-05-19
Implement events


toddbaert:
I'd like to implement `events` as mentioned [here](https://github.com/open-feature/java-sdk/issues/437).

At the moment, I'm trying to decide on the most "Java-ish" way of implementing an event listener. In the web-sdk, we add event handlers like this:

```typescript
client.addHandler(ProviderEvents.Ready, myClientOnReadyHandler);
```

Something very like this is possible in Java, but it doesn't feel very "Java-y". I think the spec gives us the flexibility to do something more like this:

```java
client.addEventObserver(new EventObserver() { // anonymous class implementing possible EventObserver interface

  @Override
  public onProviderReady(EventDetails eventDetails) {
    // do stuff
  }
  
  @Override
  public onProviderConfigurationChanged(EventDetails eventDetails) {
    // do stuff
  }
  
  @Override
  public onProviderError(EventDetails eventDetails) {
    // do stuff
  }

});
```

What are people's thoughts?

cc @open-feature/sdk-java-approvers  @open-feature/sdk-java-maintainers

kinyoklion:
I am not super engaged in the Java ecosystem, but how do people feel about SAM interfaces these days? For me it seems like the ability to provide a lambda when I just care about one aspect is nice, especially when they seem like reasonably different concerns. Though I imagine that would mean 3 registration methods.

toddbaert:
> I am not super engaged in the Java ecosystem, but how do people feel about SAM interfaces these days? For me it seems like the ability to provide a lambda when I just care about one aspect is nice, especially when they seem like reasonably different concerns. Though I imagine that would mean 3 registration methods.

I like functional interfaces, and I think that's basically how I would tackle the first example above... I fear I might be in the minority there though :sweat_smile: . I'm hoping that a discussion here can help us decide.

gRPC decided to go with something like my second proposal, FWIW: https://grpc.io/docs/languages/java/basics/#bidirectional-streaming-rpc https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html

justinabrahms:
I'll defer to others. I don't have experience with eventing within Java like this. I would naturally reach for unlined lambdas as others mention.

Kavindu-Dodan:
I am also in favor of lambda, even if that requires individual registrations. This gives the developer the freedom to register only on interested events.

toddbaert:
OK I think I'll go with the functional interface direction then!

[...]

PR #456, opened 2023-05-28
feat: add initialize and shutdown behavior

### LINKED ISSUE #449: Implement Initialize/Shutdown on provider registration
I'd like to take over the task to implement initialize and shutdown on a provider switch, like mentioned [here](https://github.com/open-feature/java-sdk/issues/437).

It currently looks to me pretty straight forward but please let me know, if you find some important points I might miss. 

One of them could be exception handling, as this is not mentioned in the spec at all. Hence, I would go with logging only for exceptions to fulfill [Requirement 1.4.9](https://github.com/open-feature/spec/blob/main/specification.json#L176).

Another one could be execution time of the calls to the provider. I see 3 options, sync without timeout, sync with timeout and async. 

## Async

**Advantages**

- setter immediately returns
- allows for long initialization phase (e.g. while startup) to run in parallel w/o impacting start up time

**Disadvantages**

- sdk code more complex as sdk needs to track initialization status of handler (esp. if handler is changed again, while the other handler is still initializing)
- expectation of client code might be to be able to immediately evaluate flags, which is only possible after initialization (see [Requirement 1.1.2.2](https://github.com/open-feature/spec/blob/main/specification.json#L18-L21))

## Sync

**Advantages**

- sdk code very easy

** Disadvantages **

- clients may expect an immediate return of the provider mutator, as it is just a "setter"
- "setter" method (`provider mutator`) potentially takes very long or even doesn't return at all, depending on the provider's initialization
- in such a case, the application wouldn't work, just because feature flags are not available - which sounds to me like being the opposite of the whole sdk's design

## Sync with Timeout

** Advantages**

- "setter" method (`provider mutator`) would definitely return in a given timeframe

** Disadvantages **

- sdk code more complex as timeout needs to be implemented
- execution of provider mutator can still take very long (depending on the defined timeout)
- what happens, if the mutator ran into the timeout? client not being informed, as no exception is thrown and hence no evaluation can be done

My initial thoughts were using the simple sync approach, but when writing up, I'd rather go with the async method as it allows for a "very long" initialization of the provider w/o impacting the client code.

cc @open-feature/sdk-java-approvers @open-feature/sdk-java-maintainers

lopitz:
## Add initialize and shutdown behavior

This PR adds the behavior of the feature provider mutator of the OpenFeatureAPI class to call `initialize()` and `shutdown()` of the set/replaced feature provider. It provides the implementation for the requirements [1.1.2.2](https://github.com/open-feature/spec/blob/975a908d8d50f11b015e7c79e163d96aff988ca1/specification.json#L18-L21) and [1.1.2.3](https://github.com/open-feature/spec/blob/main/specification.json#L25-L28).

### Related Issues

Fixes #449 

### Notes

I chose option 3 as described in #449, the asynchronous option to be implemented. This makes sure, there are no nasty delays in the application startup.
I also chose to provider empty default implementations for `initialize()` and `shutdown()` in the FeatureProvider interface, to be as backward compatible as possible.

### Follow-up Tasks
This PR doesn't implement the client events as described in #440. Depending on the implementation, the tests might need be changed. Hence, I created a helper class `FeatureProviderTestUtils` which encapsulates the new async nature of the provider mutator.

### How to test
Unit and spec tests are provided.


src/main/java/dev/openfeature/sdk/FeatureProvider.java
```diff
@@ -22,4 +22,29 @@ default List<Hook> getProviderHooks() {
...
+     * It is ok, if the method is expensive as it is executed in the background. All runtime exceptions will be
+     * caught and logged.
+     * </p>
+     */
+    default void initialize() {
+        // Intentionally left blank
+    }
+
+    /**
+     * This method is called when a new provider is about to be used to evaluate flags.
```
toddbaert, line 39:
```suggestion
     * This method is called when a new provider is about to be used to evaluate flags, or the SDK is shut down.
```

toddbaert:
I think we need a API shutdown method: as specified in [1.6.1](https://openfeature.dev/specification/sections/flag-evaluation#requirement-161). This method would basically just call `shutdown` on all providers to cleanly exit for a graceful shutdown.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
toddbaert: Related: https://github.com/open-feature/java-sdk/pull/456#issuecomment-1568938404

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -82,7 +91,8 @@ public void setProvider(FeatureProvider provider) {
         if (provider == null) {
             throw new IllegalArgumentException("Provider cannot be null");
         }
-        defaultProvider = provider;
+        shutdownProvider(this.defaultProvider);
+        initializeProvider(provider);
```
toddbaert, line 95: This is spec conformant as the spec is currently written, but I believe we may want to reword [1.1.2.2](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1122) slightly to emphasize that the "old" provider should not start until the "new" provider is done.

That would mean changing this slightly so that the shutdown of the old provider is a callback executed after initialization (or something similar). This way there is no "gap" in availability when providers are switched (though switching providers is perhaps in itself a bit of a niche use-case).

src/test/java/dev/openfeature/sdk/testutils/FeatureProviderTestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+        OpenFeatureAPI.getInstance().setProvider(provider);
+        waitForProviderInitializationComplete(OpenFeatureAPI::getProvider, provider);
+    }
+
+    private static void waitForProviderInitializationComplete(Function<OpenFeatureAPI, FeatureProvider> extractor, FeatureProvider provider) {
+        await()
+            .pollDelay(Duration.ofMillis(1))
+            .atMost(Duration.ofSeconds(1))
+            .until(() -> extractor.apply(OpenFeatureAPI.getInstance()) == provider);
+    }
```
toddbaert, line 24: We can replace this with code with code that listens for the READY event when that's complete; that's the coupling between `init/shutdown` and `events` I was talking about - it gives authors a better means of doing this sort of thing without polling, but this works just fine as well.

toddbaert:
This looks good to me. I think [this](https://github.com/open-feature/java-sdk/pull/456#issuecomment-1568938404) needs to be added. Though it could be done in a separate PR, it's related-enough to be worth implementing here. The testing looks good.

I'd like your insight on [this](https://github.com/open-feature/java-sdk/pull/456#discussion_r1210725372). The same thing came up in the JS-SDK implementation and I'm wondering if we don't need to slightly tighten the spec requirement here. Not a blocker for this PR but very interested in your opinion.

toddbaert:
@open-feature/sdk-java-approvers @open-feature/sdk-java-maintainers

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
lopitz: I was thinking the same, but wanted to be close to the requirements. I will change this to reflect the intention and potential rewording

src/test/java/dev/openfeature/sdk/testutils/FeatureProviderTestUtils.java
lopitz: exactly. that's why it is moved out to a dedicated class to just change one place and not all the test cases again.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: see https://github.com/open-feature/spec/pull/193

src/main/java/dev/openfeature/sdk/FeatureProvider.java
lopitz: done

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
lopitz: changed

lopitz:
@toddbaert , how can i retrigger the build. it failed, as it didn't find the api key for codecov for whatever reason... 🤔

Kavindu-Dodan:
> @toddbaert , how can i retrigger the build. it failed, as [it didn't find the api key for codecov](https://github.com/open-feature/java-sdk/actions/runs/5137103702/jobs/9244742999?pr=456#step:8:133) for whatever reason... 🤔

Retriggered and seems a GH glitch

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -1,25 +1,35 @@
...
     static AutoCloseableReentrantReadWriteLock hooksLock = new AutoCloseableReentrantReadWriteLock();
     static AutoCloseableReentrantReadWriteLock contextLock = new AutoCloseableReentrantReadWriteLock();
-    private EvaluationContext evaluationContext;
+
     private final List<Hook> apiHooks;
-    private FeatureProvider defaultProvider = new NoOpProvider();
     private final Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
+    private final ExecutorService taskExecutor = Executors.newCachedThreadPool();
+    private final AtomicReference<FeatureProvider> initializingDefaultProvider = new AtomicReference<>();
+    private final Map<String, FeatureProvider> initializingNamedProviders = new ConcurrentHashMap<>();
```
Kavindu-Dodan, line 29: I think we need to clean up entries from this map 🤔

Kavindu-Dodan:
Had an initial look and implementation and especially the tests are inspiring to review :) 

Left one [remark](https://github.com/open-feature/java-sdk/pull/456#discussion_r1212375126) to follow-up

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
lopitz: 🤦 regardless, of whether it would have a functional impact, this needs to be done. sorry, I missed that. will address this tonight. Thx for pointing out!

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: I'm sorry about all the back and forth on this @lopitz ! The last thing I want to do is waste your time... but it still seems like there's a bit of debate on this one. I'm starting to lean back to the previous implementation after the comments here ([particularly this one](https://github.com/open-feature/spec/pull/193#discussion_r1213344231)). I'd appreciate your opinion on this issue after working in this area.

toddbaert:
> > @toddbaert , how can i retrigger the build. it failed, as [it didn't find the api key for codecov](https://github.com/open-feature/java-sdk/actions/runs/5137103702/jobs/9244742999?pr=456#step:8:133) for whatever reason... thinking
> 
> Retriggered and seems a GH glitch

Ya it's very annoying but sometimes when codecov gets busy it throws this error, which is a total red herring.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: I've merged the spec PR above, which in the end contained mostly editorial changes. I think your original implementation (use the initialization provider immediately) is now more correct. Sorry about that again!

lopitz: don't worry. i think, writing the spec and forseeing all cases is not easy and if we gain knowledge in the implementation, we should address this in the spec - as we did. from my perspective, everything went the correct way and it wasn't annoying 😄

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
Kavindu-Dodan: Yeah, no functional impact. However, without the cleanup, we open up a memory leak. And depending on the provider's implementation details, there can be potential bugs :)

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    private boolean isOldProviderNotBoundByName(FeatureProvider oldProvider) {
+        return !this.providers.containsValue(oldProvider);
+    }
+
+    private void updateNamedProviderAfterInitialization(String clientName, FeatureProvider initializedProvider) {
+        Optional
+                .ofNullable(this.initializingNamedProviders.get(clientName))
```
Kavindu-Dodan, line 116:
```suggestion
                .ofNullable(this.initializingNamedProviders.remove(clientName))
```

This allows us to remove the dedicated remove call in `replaceNamedProviderAndShutdownOldOne` [1]. And this reduce complexity of the code and avoid future mistakes 


[1] - https://github.com/open-feature/java-sdk/blob/a6506558e0ab08ce2db9f7cae6c3440bab58066f/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L123

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private void replaceNamedProviderAndShutdownOldOne(String clientName, FeatureProvider provider) {
+        FeatureProvider oldProvider = this.providers.put(clientName, provider);
+        this.initializingNamedProviders.remove(clientName, provider);
+        if (isProviderNotBoundMultipleTimes(oldProvider)) {
+            shutdownProvider(oldProvider);
+        }
+    }
+
+    private boolean isProviderNotBoundMultipleTimes(FeatureProvider oldProvider) {
```
Kavindu-Dodan, line 129: Minor remark - Here the logic is to check whether the provider is already registred either as default or as a named provider. Given that, IMO renaming it to something like `isProviderRegistered` suits here.

toddbaert:
I think this is good to merge after [this](https://github.com/open-feature/java-sdk/pull/456#discussion_r1219956720).

src/main/java/dev/openfeature/sdk/FeatureProvider.java
lopitz: i think, my reasoning not to publish it was to not implement yet another requirement in the PR. but to be frank, i can't remember... 🤣

src/main/java/dev/openfeature/sdk/ProviderRepository.java
lopitz: we can't do this, as a new provider might be bound to the client name already, see the `filter` method in line 117. that's also the reason i chose `Map.remove(key, value)` and not simply `Map.remove(key)` in line 123.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
lopitz: i liked the not in the method name more than the `!` in the if conditions. but it's fine :)

lopitz:
@toddbaert, @Kavindu-Dodan i addressed the comments, pls re-review

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -140,4 +139,15 @@ public void clearHooks() {
...
     }
+
+    public void shutdown() {
+        providerRepository.shutdown();
+    }
+
+    /**
+     * This method is only here for testing as otherwise all tests after the API shutdown test would fail.
+     */
+    final void resetProviderRepository() {
```
toddbaert, line 150: fine with me since it's package private.

toddbaert (approved):

Issue #437, opened 2023-05-16
Changes to support spec `v0.6.0` compliance


toddbaert:
## Implement specification v0.6.0 changes

See [changes since last release](https://github.com/open-feature/spec/compare/v0.5.2...v0.6.0).

Please note there's no need to complete all these tasks in a single PR (in fact, that's not recommended).

### Functional requirements:

- [x] implement flag metadata: https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a
- [x] #442
- [x] #449 : https://github.com/open-feature/spec/commit/a4ffec3d4416b8ad6d12d1ebfa2faf81ad153f65
- [x] #440
  - the spec is intentionally loose here to allow for an implementation that feels "java-ish", consider:
    - [java.util.EventListener](https://docs.oracle.com/javase/8/docs/api/java/util/EventListener.html) and its subclasses
    - a custom interface
    - other options?

### Non-functional requirements:

- [x] add relevant tests
- [x] add examples and documentation, see template README
- [x] update specification badge to `v0.6.0`: [![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)

### Keep in mind:

- backwards compatibility is _highly encouraged_
  - take every precaution to ensure existing code written by _application authors_ and existing providers implementations, do not break
- it's recommended these features are released together

toddbaert:
named client support was implemented by @justinabrahms

toddbaert:
I'm interested in taking on the events part of this. I'm going to open a new issue specifically about that if nobody objects.

Kavindu-Dodan:
I am actively working on metadata requirement and will open a PR soon for the review

toddbaert:
Done with https://github.com/open-feature/java-sdk/pull/364

PR #459, opened 2023-05-31
feat: add flag metadata

Kavindu-Dodan:
## This PR

Is connected to #437 & completes the implementation of flag metadata spec implementation [1]

The implementation is backed by immutable `FlagMetadata`. If present, metadata gets propagated from `ProviderEvaluation` to `FlagEvaluationDetails`

Tests were added to validate the construction as well as the propagation 

[1] - https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a

### DIFF
--- a/src/main/java/dev/openfeature/sdk/FlagMetadata.java
+++ b/src/main/java/dev/openfeature/sdk/FlagMetadata.java
@@ -1,7 +1,6 @@
 package dev.openfeature.sdk;
 
-import dev.openfeature.sdk.exceptions.GeneralError;
-import dev.openfeature.sdk.exceptions.ParseError;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -10,6 +9,7 @@ import java.util.Map;
  * Immutable Flag Metadata representation. Implementation is backed by a {@link Map} and immutability is provided
  * through builder and accessors.
  */
+@Slf4j
 public class FlagMetadata {
     private final Map<String, Object> metadata;
 
@@ -18,8 +18,8 @@ public class FlagMetadata {
     }
 
     /**
-     * Retrieve a {@link String} value for the given key. If a value is not found, {@link GeneralError} will be thrown.
-     * If value exist but of another type, {@link ParseError} will be thrown.
+     * Retrieve a {@link String} value for the given key. A {@code null} value is returned if the key does not exist
+     * or if the value is of a different type.
      *
      * @param key flag metadata key to retrieve
      */
@@ -28,9 +28,8 @@ public class FlagMetadata {
     }
 
     /**
-     * Retrieve an {@link Integer} value for the given key.
-     * If a value is not found, {@link GeneralError} will be thrown.
-     * If value exist but of another type, {@link ParseError} will be thrown.
+     * Retrieve a {@link Integer} value for the given key. A {@code null} value is returned if the key does not exist
+     * or if the value is of a different type.
      *
      * @param key flag metadata key to retrieve
      */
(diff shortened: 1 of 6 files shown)


src/main/java/dev/openfeature/sdk/FlagMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public Boolean getBoolean(final String key) {
+        return getValue(key, Boolean.class);
+    }
+
+    private <T> T getValue(final String key, final Class<T> type) {
+        final Object o = metadata.get(key);
+
+        if (o == null) {
+            throw new GeneralError("key " + key + " does not exist in metadata");
```
toddbaert, line 77: I _think_ we want to return null here, not throw. That's consistent with the context and structure abstractions in the SDK as well.

I could be wrong about this, open to hear why you think throwing is better.

toddbaert:
@Kavindu-Dodan this all looks good to me, except that I'm not sure about [throwing on failed retrieval](https://github.com/open-feature/java-sdk/pull/459#discussion_r1212183917).

This seems like a typical eventuality, not an exceptional one.

src/main/java/dev/openfeature/sdk/FlagMetadata.java
Kavindu-Dodan: I went with the exception to match with the go-sdk implemenation [1]. I can go ahead with null if we agree to that behavior 

[1] - https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/client.go#L179

thisthat: I am not familiar with the control- and data-flow of the Java SDK, but I am also in favor of avoiding exceptions. The SDK should never fail at runtime IMHO. If exception are the best way to communicate to the caller, then I suggest explicitly marking the methods with JavaDoc annotations for the exception and having the `throws` annotation in the method signature

toddbaert: > I went with the exception to match with the go-sdk implemenation [1]. I can go ahead with null if we agree to that behavior

Ya, I think the thing is that practically, returning an error is fundamentally different than exceptions. I can choose to check the error field or not - I _must_ handle exceptions (unless I like crashes a lot).

Kavindu-Dodan: @thisthat @toddbaert  updated the implementation to avoid exceptions. Now it log(debug level) and return null value for both scenarios

src/main/java/dev/openfeature/sdk/FlagMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+import dev.openfeature.sdk.exceptions.ParseError;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Immutable Flag Metadata representation. Implementation is backed by a {@link Map} and immutability is provided
+ * through builder and accessors.
+ */
+public class FlagMetadata {
```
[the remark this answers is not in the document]
Kavindu-Dodan: Our current implementation focuses on setters/getters for types, where data is backed by a map. This is different from a value class as our main goal is to provide immutability and spec compliance. 

In the implementation, immutability is guaranteed with the builder, and spec compliance is provided through builder-constructed object accessors.

Kavindu-Dodan:
> @Kavindu-Dodan this all looks good to me, except that I'm not sure about [throwing on failed retrieval](https://github.com/open-feature/java-sdk/pull/459#discussion_r1212183917).
> 
> This seems like a typical eventuality, not an exceptional one.

Updated the implementation to avoid exceptions. Instead now we return null for missing key & value type differences , along with debug level log entry

toddbaert (approved):

thiyagu06 (approved):
LGTM 👍

beeme1mr (approved):

toddbaert:
@Kavindu-Dodan I rebased on main and resolved conflicts with @lopitz 's recent PR. I hope you don't mind. This looks good to me, and we have lots of approvals so I'm merging.

PR #476, opened 2023-06-14
feat: events

toddbaert:


src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    /**
+     * Propagates all events from the originatingEmitter to this one.
+     *
+     * @param originatingEmitter The emitter to forward events from.
+     * @param clientName         The client name that will be added to the events.
+     */
+    void forwardEvents(EventEmitter originatingEmitter, @Nullable String clientName) {
```
toddbaert, line 99: This method allows us to "proxy" events from the provider in question, and easily move them when the provider is updated.

src/main/java/dev/openfeature/sdk/EventHandling.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface EventHandling<T> {
+    
+    T onProviderReady(Consumer<EventDetails> handler);
+    
+    T onProviderConfigurationChanged(Consumer<EventDetails> handler);
+    
+    T onProviderError(Consumer<EventDetails> handler);
+    
+    T onProviderStale(Consumer<EventDetails> handler);
```
toddbaert, line 16: @lopitz you suggested these kinds of names... except I think you suggested "doOn..." not "on...". Do you have a preference? "doOn..." struck me as a bit redundant, but I don't feel too strongly about it.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public interface EventProvider {
+
+    /**
+     * Return the EventEmitter interface for this provider.
+     * The same instance should be returned from this method at each invocation.
+     * 
+     * @return the EventEmitter instance for this EventProvider. 
+     */
+    EventEmitter getEventEmitter();
```
toddbaert, line 16: _Provider authors_ can either implement this, or extend the abstract `SimpleEventProvider`. The thing I don't like about this interface, is we rely on the instance returned by `getEventEmitter` always being the same, or things won't work. `SimpleEventProvider` works this way, but there's some reduced flexibility involved with using abstracts... so I've left both options. Interested in feedback on this.

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Event emitter construct to be used by providers.
+ */
+@Slf4j
+public class EventEmitter {
+
+    private static final ExecutorService taskExecutor = Executors.newCachedThreadPool();
```
toddbaert, line 24: This thread pool is used to execute the handlers... I don't think it's huge deal, but it's possible that `newCachedThreadPool` could lead to sudden increased in resource consumption if lots of handlers are fired at the same time with some event. Having a fixed or single thread pool could reduce the changes of this happening, but we were already using a cached thread pool elsewhere.

pom.xml
```diff
@@ -190,6 +190,9 @@
             <goals>
               <goal>cpu-count</goal>
             </goals>
+            <configuration>
+              <factor>0.25</factor>
+            </configuration>
```
toddbaert, line 195: 1 fork per core causes a lot of thrashing and really slows things, and also seems to cause flaky tests due to timeouts. This seems more stable.

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.Data;
+import lombok.experimental.SuperBuilder;
+
+/**
+ * Interface for attaching event handlers.
+ */
+@Data @SuperBuilder(toBuilder = true)
+public class ProviderEventDetails {
+    @Nullable private List<String> flagsChanged;
```
toddbaert, line 15: Should this be an array or a list?

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Construct a new EventEmitter.
+     */
+    public EventEmitter() {
+        handlerMap = new ConcurrentHashMap<ProviderEvent, List<Consumer<EventDetails>>>();
+        handlerMap.put(ProviderEvent.PROVIDER_READY, new ArrayList<>());
+        handlerMap.put(ProviderEvent.PROVIDER_CONFIGURATION_CHANGED, new ArrayList<>());
+        handlerMap.put(ProviderEvent.PROVIDER_ERROR, new ArrayList<>());
+        handlerMap.put(ProviderEvent.PROVIDER_STALE, new ArrayList<>());
```
toddbaert, line 36: We could construct this map lazily if we wanted.

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+        });
+    }
+
+    void addHandler(ProviderEvent event, Consumer<EventDetails> handler) {
+        try (AutoCloseableLock __ = handlersLock.writeLockAutoCloseable()) {
+            this.handlerMap.get(event).add(handler);
+        }
+    }
+
+    void removeHandler(ProviderEvent event, Consumer<EventDetails> handler) {
```
toddbaert, line 80: I see now I should add tests specifically for removal of handlers. I will do that.

src/main/java/dev/openfeature/sdk/EventHandling.java
lopitz: `doOn...` was just an example. i'm totally fine with these names.

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
```diff
@@ -0,0 +1,18 @@
...
+    @Nullable private String message;
+    @Nullable private FlagMetadata eventMetadata; // TODO: this should use a different type.
```
toddbaert, line 17: @Kavindu-Dodan this field essentially has to be the same thing as your recently implemented `FlagMetadata`... however, the name `FlagMetadata` doesn't make sense.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -28,24 +30,42 @@ public class OpenFeatureClient implements Client {
...
+     *             Use the OpenFeatureAPI's getClient factory method instead.
      */
+    @Deprecated()
     public OpenFeatureClient(OpenFeatureAPI openFeatureAPI, String name, String version) {
         this.openfeatureApi = openFeatureAPI;
         this.name = name;
         this.version = version;
         this.clientHooks = new ArrayList<>();
         this.hookSupport = new HookSupport();
+        log.info(
```
thisthat, line 56: Maybe we should use a warn level?

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
thisthat: I would prefer having this handle by lombok and stay on the `List`

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
toddbaert: Will rebase on https://github.com/open-feature/java-sdk/pull/478

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
toddbaert: Ya, I agree. Fixed.

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
toddbaert: > I would prefer having this handle by lombok and stay on the List

Just to clarify... this is already handled by lombok builder. Is there some other lombok tooling you want added? Or are you just saying  you want it kept as a list.

pom.xml
Kavindu-Dodan: We should also consider GH Action & rate limiting. So, may be it's a good idea to define a general test timeout  through a constant (for parallel ones) and increase it a bit along with this change :)

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+                eventDetails.setClientName(((EventDetails) details).getClientName());
+            }
+
+            this.handlerMap.get(event).stream().forEach(handler -> {
+                runHander(handler, eventDetails);
+            });
+        }
+    }
+
+    void runHander(Consumer<EventDetails> handler, EventDetails eventDetails) {
```
Kavindu-Dodan, line 64:
```suggestion
    void runHandler(Consumer<EventDetails> handler, EventDetails eventDetails) {
```

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+    public void emit(ProviderEvent event, ProviderEventDetails details) {
+        try (AutoCloseableLock __ = handlersLock.readLockAutoCloseable()) {
+            EventDetails eventDetails = EventDetails.builder()
+                    .eventMetadata(details.getEventMetadata())
+                    .flagsChanged(details.getFlagsChanged())
+                    .message(details.getMessage())
+                    .build();
+
+            // we may be forwarding this event, preserve the name if so.
+            if (EventDetails.class.isInstance(details)) {
```
Kavindu-Dodan, line 54:
```suggestion
             if (details instanceof EventDetails) {
```

IMO, instanceof is easy to read and understand

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+                eventDetails.setClientName(((EventDetails) details).getClientName());
+            }
+
+            this.handlerMap.get(event).stream().forEach(handler -> {
+                runHander(handler, eventDetails);
+            });
```
Kavindu-Dodan, line 60:
```suggestion
this.handlerMap.get(event).forEach(handler -> runHander(handler, eventDetails));
```

Minor - List expose `forEach` so stream is not needed

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param originatingEmitter The emitter to forward events from.
+     * @param clientName         The client name that will be added to the events.
+     */
+    void forwardEvents(EventEmitter originatingEmitter, @Nullable String clientName) {
+        Arrays.asList(ProviderEvent.values()).stream().forEach(eventType -> {
+            originatingEmitter.addHandler(eventType, details -> {
+                // set the client name when we proxy the events through.
+                details.setClientName(clientName);
+                this.emit(eventType, details);
```
Kavindu-Dodan, line 104: Any reason to emit event when forwarding handlers 🤔  ?

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -1,24 +1,25 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
 @Slf4j
 class ProviderRepository {
 
     private final Map<String, FeatureProvider> providers = new ConcurrentHashMap<>();
-    private final ExecutorService taskExecutor = Executors.newCachedThreadPool();
-    private final Map<String, FeatureProvider> initializingNamedProviders = new ConcurrentHashMap<>();
+    private final Map<String, EventEmitter> clientEmitters = new ConcurrentHashMap<>();
```
Kavindu-Dodan, line 19: This map get populated for client names, but we never clean (`Map.remove()`) entries from it 🤔  I think we need a way to clean up this

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: :sweat_smile:

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: agreed.

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: Events are forwarded _by_ emitting them again.

Basically, forwarding amounts to listening to events from the original emitter, and re-emitting them.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
toddbaert: hmmm I guess that would be done if we "unset" a client? I'll think about that.

Kavindu-Dodan: Yeah, I think we must not define the `unset` behavior, but detect garbage collection of a client and then perform cleanup, or any other mechanism.

If we are tying this to GC, then we should be able to rely on Object.finalize() [1]

[1] - https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#finalize--

pom.xml
toddbaert: I haven't seen issues in GH, and this change is already big.

I think what you're proposing is a good idea, but it should be done separately to not make this PR even bigger.

Kavindu-Dodan: Agree, let's focus it through a dedicated issue :)

src/main/java/dev/openfeature/sdk/EventEmitter.java
Kavindu-Dodan: I am still confuesed on this. For example, let's take the following scenario - setting the default provider. Current call stack looks like below,

API.setProvider(provider) - > API.propagateEventsIfSupported(provider, clientName) -> emitter.forwardEvents(eventProvider, clientName) -> emitter.emit(eventType, details)

This flow suggests adding known API-level events to the new provider and then running them. Or am I getting it wrong here.

src/main/java/dev/openfeature/sdk/EventHandling.java
```diff
@@ -0,0 +1,19 @@
...
+    
+    T removeHandler(ProviderEvent event, Consumer<EventDetails> handler);
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is fine by me

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -63,66 +64,70 @@ public void setProvider(String clientName, FeatureProvider provider) {
...
+    }
+
+    private void initializeProvider(@Nullable String clientName, FeatureProvider newProvider) {
+        forwardProviderEvents(clientName, newProvider);
         taskExecutor.submit(() -> {
             try {
-                if (!isProviderRegistered(provider)) {
-                    provider.initialize();
+                if (!isProviderRegistered(newProvider)) {
+                    newProvider.initialize(OpenFeatureAPI.getInstance().getEvaluationContext());
```
s-sen, line 95: Is the Global EvaluationContext set in the OpenFeature API common to all providers? Is there any way a specific provider can be initialized with context specific to that provider?

toddbaert: > Is the Global EvaluationContext set in the OpenFeature API common to all providers? Is there any way a specific provider can be initialized with context specific to that provider?

This is an interesting though, but not something we have spec'd. It's the same global context for each provider at init. The context set at the `client` level would be specific to the provider - but that wouldn't be available at init.

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: I think you're right, it's worth doing. I'll make things as lazy as possible to save resources. Until handlers are actually defined we can use almost no space.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
Kavindu-Dodan: Thanks to our discussion on this subject, I am convinced that this is not a blocker nor requires further investigation. What we should do is enhance readme on the point that client event handlers (scoped event handlers) require manual handler removal if they are intended to be dynamically created (ex:- creating a unique client per each request)

src/main/java/dev/openfeature/sdk/ProviderEventDetails.java
thisthat: Sorry, I didn't explain my reasoning well. I suggest staying on List because of Lombok. AFAIK, it wouldn't work with arrays :)

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: I've made the initialization of the store lazy, for each client, but not each event type, meaning that I make space for all event types for a client if a single one is added. I think the tiny memory saved here is not work the complexity.

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: I've completely reworked this.

Now all handlers are stored in a single map, instead of doing this complex proxying.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
toddbaert: I've made changes to only allocate memory for clients that actually add handlers.

pom.xml
```diff
@@ -261,9 +246,6 @@
           <argLine>
             ${surefireArgLine}
           </argLine>
-          <!-- fork a new JVM to isolate test suites, especially important with singletons -->
-          <forkCount>${cpu.count}</forkCount>
-          <reuseForks>false</reuseForks>
```
toddbaert, line 266: As witnessed by @Kavindu-Dodan the test suite is actually 3x faster without any forking at all :shrug:

src/main/java/dev/openfeature/sdk/EventHandling.java
```diff
@@ -0,0 +1,64 @@
+package dev.openfeature.sdk;
```
toddbaert, line 1: This is the public interface for both the API and client events.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -0,0 +1,83 @@
...
+ * note that the SDK will automatically emit
+ * {@link ProviderEvent#PROVIDER_READY } or
+ * {@link ProviderEvent#PROVIDER_ERROR } accordingly when
+ * {@link FeatureProvider#initialize(EvaluationContext)} completes successfully
+ * or with error, so these events need not be emitted manually during
+ * initialization.
+ *
+ * @see FeatureProvider
+ */
+public abstract class EventProvider implements FeatureProvider {
```
toddbaert, line 17: This is the class _application authors_ must implement to use events. See javadoc above.

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Emit an event.
+     *
+     * @param event   Event type to emit.
+     * @param details Event details.
+     */
+    public void emit(ProviderEvent event, ProviderEventDetails details) {
+        try (AutoCloseableLock __ = handlersLock.readLockAutoCloseable()) {
```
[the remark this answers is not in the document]
toddbaert: Things have significantly changed here. This line doesn't exist anymore. I hope you don't mind re-reviewing (especially the locking).

src/main/java/dev/openfeature/sdk/EventEmitter.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Event emitter construct to be used by providers.
+ */
+@Slf4j
+public class EventEmitter {
+
+    private static final ExecutorService taskExecutor = Executors.newCachedThreadPool();
+    private final Map<ProviderEvent, List<Consumer<EventDetails>>> handlerMap;
+    private AutoCloseableReentrantReadWriteLock handlersLock = new AutoCloseableReentrantReadWriteLock();
```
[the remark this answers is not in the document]
toddbaert: Things have significantly changed here. This line doesn't exist anymore. I hope you don't mind re-reviewing (especially the locking).

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -135,19 +185,123 @@ public List<Hook> getHooks() {
...
+    void removeHandler(String clientName, ProviderEvent event, Consumer<EventDetails> handler) {
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
+            eventSupport.removeClientHandler(clientName, event, handler);
+        }
+    }
+
+    void addHandler(String clientName, ProviderEvent event, Consumer<EventDetails> handler) {
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
+            // if the provider is READY, run immediately
+            if (event.equals(ProviderEvent.PROVIDER_READY)
```
thisthat, line 259:
```suggestion
            if (ProviderEvent.PROVIDER_READY.equals(event)
```

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -135,19 +185,123 @@ public List<Hook> getHooks() {
...
+                    && ProviderState.READY.equals(this.providerRepository.getProvider(clientName).getState())) {
+                eventSupport.runHandler(handler, EventDetails.builder().clientName(clientName).build());
```
thisthat, line 261: I think we should run the handler outside of the write lock

src/main/java/dev/openfeature/sdk/EventEmitter.java
toddbaert: I did this.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: agree

src/main/java/dev/openfeature/sdk/EventSupport.java
```diff
@@ -0,0 +1,112 @@
...
+        final String name = Optional.ofNullable(clientName)
+                .orElse(defaultClientUuid);
+
+        HandlerStore store = Optional.ofNullable(this.handlerStores.get(name))
+                .orElseGet(() -> {
+                    HandlerStore newStore = new HandlerStore();
+                    this.handlerStores.put(name, newStore);
+                    return newStore;
+                });
+        if (this.handlerStores.get(name) == null) {
```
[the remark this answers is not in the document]
toddbaert: lol nice catch. This was a refactor.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: I think, in effect, we are.

The handler is run in a new thread by the executor. This thread doesn't block when calling `eventSupport.runHandler`.

src/main/java/dev/openfeature/sdk/EventSupport.java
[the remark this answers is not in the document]
toddbaert: fixed.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
thisthat: Gotcha, thanks for the explanation 👍

liran2000:
\* Not sure whether it fits to this PR, spec, or a discussion, otherwise I can open a new thread

# Provider Initialization Fallback

## Problem effect
With this scenario occurs, service applications can have unexpected results, since all flags evaluations returns default value.

## Problem description
Scenario: **provider service is temporary down during provider initialization**. Since provider feature flags configuration cannot be fetched, **all flags evaluations returns default value** until provider service is up.
Different scenario such as provider service is temporary down during service run is less problematic, since flags evaluation can be done via local feature flags configuration cache.

## Provider Initialization Fallback Mechanism - High level solution
* Configuration changed/fetched -> save configuration to cache, and also to file / other storage.
* Provider initialization Error event -> load configuration from the file / storage.

## OpenFeature treatment for this scenario
Provider initialization has an Error event.  
Can a solution be adopted via this PR ? If so, can a unit test be written to simulate it ? Otherwise, does it requires a spec update ?

Thoughts ?

@beeme1mr

toddbaert:
> * Not sure whether it fits to this PR, spec, or a discussion, otherwise I can open a new thread
> 
> # Provider Initialization Fallback
> ## Problem effect
> With this scenario occurs, service applications can have unexpected results, since all flags evaluations returns default value.
> 
> ## Problem description
> Scenario: **provider service is temporary down during provider initialization**. Since provider feature flags configuration cannot be fetched, **all flags evaluations returns default value** until provider service is up. Different scenario such as provider service is temporary down during service run is less problematic, since flags evaluation can be done via local feature flags configuration cache.
> 
> ## Provider Initialization Fallback Mechanism - High level solution
> * Configuration changed/fetched -> save configuration to cache, and also to file / other storage.
> * Provider initialization Error event -> load configuration from the file / storage.
> 
> ## OpenFeature treatment for this scenario
> Provider initialization has an Error event. Can a solution be adopted via this PR ? If so, can a unit test be written to simulate it ? Otherwise, does it requires a spec update ?
> 
> Thoughts ?
> 
> @beeme1mr

@liran2000 this is the expected result for this situation. This PR indirectly helps with this situation, but allowing for authors to attach event handlers (such as error events) for the case you describe.

In any case, if you'd like to discuss this behavior further, please open a separate issue for that.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -83,7 +96,14 @@ public EvaluationContext getEvaluationContext() {
...
      */
     public void setProvider(FeatureProvider provider) {
-        providerRepository.setProvider(provider);
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
+            providerRepository.setProvider(
+                    provider,
+                    (p) -> attachEventProvider(p),
+                    (p) -> emitReady(p),
+                    (p) -> detachEventProvider(p),
+                    (p, message) -> emitError(p, message));
```
Kavindu-Dodan, line 105: for readability, 

```suggestion
                    this::attachEventProvider,
                    this::emitReady,
                    this::detachEventProvider,
                    this::emitError)
```

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -93,7 +113,37 @@ public void setProvider(FeatureProvider provider) {
...
      */
     public void setProvider(String clientName, FeatureProvider provider) {
-        providerRepository.setProvider(clientName, provider);
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
+            providerRepository.setProvider(clientName,
+                    provider,
+                    (p) -> attachEventProvider(p),
+                    (p) -> emitReady(p),
+                    (p) -> detachEventProvider(p),
+                    (p, message) -> emitError(p, message));
```
Kavindu-Dodan, line 122: for readability, 

```suggestion
                    this::attachEventProvider,
                    this::emitReady,
                    this::detachEventProvider,
                    this::emitError)
```

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -135,19 +185,123 @@ public List<Hook> getHooks() {
...
     public void clearHooks() {
-        try (AutoCloseableLock __ = hooksLock.writeLockAutoCloseable()) {
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
             this.apiHooks.clear();
         }
     }
 
     public void shutdown() {
         providerRepository.shutdown();
+        // TODO: shutdown events
```
Kavindu-Dodan, line 195: needs to remove the todo

Kavindu-Dodan (approved):
Overll, I do not see any issues except for the minor remarks I left :)

src/main/java/dev/openfeature/sdk/EventEmitter.java
Kavindu-Dodan: This should be fine as long as we do not have a lot of callbacks as we execute them concurrently.

However, one thing we should consider is to introduce a timeout for callbacks. I am introducing this in go-sdk and I think this is needed to avoid any performant impacts.

pom.xml
Kavindu-Dodan: This observation can be explained with the surefire documentation [1]

> The parameter reuseForks is used to define whether to terminate the spawned process after one test class and to create a new process for the next test in line (reuseForks=false), or whether to reuse the processes to execute the next tests (reuseForks=true).

`forkCount` helped with creating N parallel forks but, given that `reuseForks` was false, tests did not reuse them. This means, while tests ran in parallel forks, new tests require a new process, adding a delay between executions. 

[1] - https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html#forked-test-execution

src/main/java/dev/openfeature/sdk/EventHandling.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.function.Consumer;
+
+/**
+ * Interface for attaching event handlers.
+ */
+public interface EventHandling<T> {
```
Kavindu-Dodan, line 8: Remark - We can also think about a single `addEventHandler()` with `ProviderEvent` enum value as a parameter. But this is just a suggestion and not a requirement.

toddbaert: If I'm understanding you correctly, that's the point of "on":

```java
    /**
     * Add a handler for the specified {@link ProviderEvent}.
     * 
     * @param event event type
     * @param handler behavior to add with this event
     * @return this
     */
    T on(ProviderEvent event, Consumer<EventDetails> handler);
```

Kavindu-Dodan: Ohh, I didn't see this clearly. This is done then :)

pom.xml
toddbaert: Thanks!

I guessed it was something like that. It makes sense that starting up a bunch of JVMs would slow things down.

I did just try `<forkCount>1C</forkCount>` (which means 1 x number-of-cores), while allowing fork re-use, and the speed of the suite is unchanged from the default... so it doesn't slow things down but doesn't really help much.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: Absolutely. I always forget about this.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: https://github.com/open-feature/java-sdk/pull/476#discussion_r1244194878

src/main/java/dev/openfeature/sdk/NoOpProvider.java
```diff
@@ -10,6 +10,12 @@ public class NoOpProvider implements FeatureProvider {
     @Getter
     private final String name = "No-op Provider";
 
+    // The Noop provider is ALWAYS NOT_READY, otherwise READY handlers would run immediately when attached.
```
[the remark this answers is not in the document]
toddbaert: No, the reason for the `NOT_READY` is so the SDK knows whether or not to initialize the provider and fire `READY` events, as dictated by the spec. It doesn't impact evaluation (the SDK still calls all the resolver methods).

I know this is a bit weird, but the Noop provider is really a special case - it only exists as a simple way to maintain the "no op" behavior when no "real" provider is set. An SDK could neglect to implement a no-op provider entirely and it would be compliant.

What this avoids is `READY` events firing immediately when no "real" provider has been set, which would be very confusing. I'd rather not have to add some hack to accomplish that.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private TriConsumer<EventProvider, ProviderEvent, ProviderEventDetails> onEmit = null;
+
+    void detach() {
+        this.onEmit = null;
+    }
+
+    void attach(TriConsumer<EventProvider, ProviderEvent, ProviderEventDetails> onEmit) {
+        if (this.onEmit == null) {
```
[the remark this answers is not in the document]
toddbaert: :thinking: I'll think more about the error scenarios here, and maybe add some comments.

src/main/java/dev/openfeature/sdk/EventSupport.java
```diff
@@ -0,0 +1,115 @@
...
+class EventSupport {
+
+    // we use a v4 uuid as a "placeholder" for anonymous clients, since
+    // ConcurrentHashMap doesn't support nulls
+    private static final String defaultClientUuid = UUID.randomUUID().toString();
+    private static final ExecutorService taskExecutor = Executors.newCachedThreadPool();
+    private final Map<String, HandlerStore> handlerStores = new ConcurrentHashMap<>();
+    private final HandlerStore globalHandlerStore = new HandlerStore();
+
+    public void runClientHandlers(@Nullable String clientName, ProviderEvent event, EventDetails eventDetails) {
```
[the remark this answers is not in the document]
toddbaert: Sure I'll try to add more javadoc and inline comments.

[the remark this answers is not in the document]
toddbaert: I've documented all public methods, and added additional in-line comments.

Note this is an entirely internal class, but I agree with the idea that exposed methods should still be documented if they are called from outside the class.

src/main/java/dev/openfeature/sdk/EventProvider.java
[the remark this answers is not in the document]
toddbaert: I've added some logic for edge cases here, and specific unit tests for this.

src/main/java/dev/openfeature/sdk/EventHandling.java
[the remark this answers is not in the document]
toddbaert: I've gone with `EventBus`, which seems consistent with similar constructs in other frameworks/languages/patters (https://dzone.com/articles/design-patterns-event-bus, https://github.com/greenrobot/EventBus)

Alternatives could include `EventNode` or `EventHub`.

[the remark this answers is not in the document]
beeme1mr: 🚌

justinabrahms (approved):

PR #478, opened 2023-06-15
chore: rename flag metadata

toddbaert:
Renames the `FlagMetadata` class (it can be used for event metadata as well). Talked with @Kavindu-Dodan (the implementer) about this. This is not yet released, so there's no breaking change to worry about.


Kavindu-Dodan (approved):
As discussed, this change allows to use `ImmutableMetadata` in many other SDK use cases.

beeme1mr (approved):

PR #487, opened 2023-06-22
feat: ProviderEvaluation json de/serialization support

### LINKED ISSUE #483: ProviderEvaluation is hard to serialize / deserialize
We're trying to build a custom REST backend using spring that returns ProviderEvaluation objects serialized to json. 

The current `ProviderEvaluation` class makes it very hard to deserialize it using jackson, because it uses lombok `Builder` and does not expose an empty default constructor.

Would it be possible to add the [Jacksonized](https://projectlombok.org/features/experimental/Jacksonized) lombok annotation, so deserialization of these instances is possible?

Kavindu-Dodan:
## This PR

Fixes #483 by adding annotations and accessors to required classes.

### Testing

Unit tests were added to validate each class's compliance on json de/serialization support

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,18 @@
       <version>2.0.7</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+      <version>2.15.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.15.2</version>
+    </dependency>
+
     <!-- test -->
     <dependency>
       <groupId>org.mockito</groupId>
@@ -150,6 +162,13 @@
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <version>4.2.0</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+      <version>2.15.2</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/src/main/java/dev/openfeature/sdk/ImmutableMetadata.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableMetadata.java
@@ -1,5 +1,7 @@
 package dev.openfeature.sdk;
 
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
@@ -13,8 +15,15 @@ import java.util.Map;
 public class ImmutableMetadata {
     private final Map<String, Object> metadata;
 
-    private ImmutableMetadata(Map<String, Object> metadata) {
+    private ImmutableMetadata(@JsonProperty("metadata") Map<String, Object> metadata) {
         this.metadata = metadata;
+    }
+
+    // This is required for serialization support
+    @SuppressWarnings({"PMD.UnusedPrivateMethod"})
+    @JsonGetter("metadata")
+    private Map<String, Object> getMetadata() {
+        return this.metadata;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -63,6 +63,18 @@
...
+      <artifactId>jackson-annotations</artifactId>
+      <version>2.15.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.15.2</version>
+    </dependency>
+
```
toddbaert, line 77: I'm not sure how I feel about adding these jackson dependencies.

Kavindu-Dodan: I haven't personally seen or used jaxb for json de/serialization. Besides, built-in support for Java SE for jaxb is removed from Java 9+ [1]

If we need to support de/serialization, I believe we will have some sort of dependency to a library. Here, my observation was that`com.fasterxml.jackson.core: jackson-databind` is anyway needed for `@Jacksonized` annotation. 

[1] - https://openjdk.org/jeps/320

toddbaert: > Besides, built-in support for Java SE for jaxb is removed from Java 9+

That's a bummer.

I have to think about this a bit more. As I mentioned in the original issue, my enthusiasm for supporting this was predicated on it not adding more dependencies. Right now we have only `slf4j` as an actual dependency. This is significantly more.

> Here, my observation was thatcom.fasterxml.jackson.core: jackson-databind is anyway needed for @Jacksonized annotation.

Oh - I didn't read enough into the lombok annotation. I just though it added POJO methods that jackson used via reflection or something. You're right that it seems to be adding jackson annotations though.

I have mixed feelings about this now. I wonder what @lopitz and @justinabrahms think.

toddbaert (changes requested):
See [here](https://github.com/open-feature/java-sdk/pull/487#discussion_r1240315233).

pom.xml
lopitz: @toddbaert , @Kavindu-Dodan 

Jackson is just on json serializer/deserializer. There's also JSON and JAXB implementations and I guess a gazillion others. imho, we should not rely on jackson, just because a single user complains about it being hard to use. seems like the real issue is the missing default constructor in the dto. hence, i'd simply add a `@NoArgsConstructor` annotation and leave it with that. Looking at this class tells me also, that it is fine, as it is a `@Data` annotated class and not a `@Value` annotated one. I guess, we should have a default constructor for all data classes to stay consistent and allow for things like object creation via reflection...

toddbaert: > @toddbaert , @Kavindu-Dodan
> 
> Jackson is just on json serializer/deserializer. There's also JSON and JAXB implementations and I guess a gazillion others. imho, we should not rely on jackson, just because a single user complains about it being hard to use. seems like the real issue is the missing default constructor in the dto. hence, i'd simply add a `@NoArgsConstructor` annotation and leave it with that. Looking at this class tells me also, that it is fine, as it is a `@Data` annotated class and not a `@Value` annotated one.

I agree. I was hopeful that JAXB bindings would solve this issue for us, but I found out in this thread they're non-standard after Java9.

>  I guess, we should have a default constructor for all data classes to stay consistent and allow for things like object creation via reflection...

Ya, like I said in the associated issue, I think supporting an empty constructor is a good idea, and we should probably do so for all such "DTO" classes.

Kavindu-Dodan: Thank you @lopitz  & @toddbaert  for your suggestions. 

I checked your proposals but still, there's a blocker to proceed here. The de/serialization error is there because of the `ImmutableMetadata` class [1]. This class was made immutable as this is a recommendation from the spec [2]. Given that  `ImmutableMetadata` is a special class, I do not see a workaround to this.

I also think adding 3rd party libraries to support such a use case is justifiable as this goes away from being simple. And the best recommendation would be to propose custom wrappers for de/serialization purposes.

[1] - https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/ProviderEvaluation.java#L17
[2] - https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#conditional-requirement-14141

toddbaert: I think we can still add the empty constructor. I t might make things easier, though it doesn't solve the entire problem.

Kavindu-Dodan: @toddbaert Thank you for the #491 

With this conclusion, I will close this PR as we agree on not exposing any specific annotation support

Kavindu-Dodan:
we will not support a specific JSON de/serialization framework, however #491 will help with empty constructor support.

PR #531, opened 2023-07-26
fix: getState mandatory on EventProvider

toddbaert:
In 3 event provider implementations, [flagd-java](https://github.com/open-feature/java-sdk-contrib/pull/361#pullrequestreview-1541468446), [go-feature-flag-web](https://github.com/open-feature/js-sdk-contrib/pull/474#discussion_r1274012627), and even my own implementation of flagd-js, authors have forgotten to properly implement `getState` (which means the SDK assumes the provider is already `READY` and doesn't run `initialize()`.)

This is an easy pitfall. To help with this, I'm making the implementation of `getState` required in the abstract `EventProvider` class, and will do similarly in the JS-SDK. I think this is a reasonable proposal because the `EventProvider` class is very new, and in the majority of cases a provider that can fire async events requires some kind of remote connection, so it will need initialization. This is technically a breaking change, but it would only impact the experimental events API, specifically the `EventProvider` implementations which could only be a few days old at this point, and which (due to my previous point) likely should have implemented this anyway.

It also adds some more doc about the importance of this method.

Relates to: https://github.com/open-feature/js-sdk/pull/506

### DIFF
--- a/src/main/java/dev/openfeature/sdk/FeatureProvider.java
+++ b/src/main/java/dev/openfeature/sdk/FeatureProvider.java
@@ -31,7 +31,7 @@ public interface FeatureProvider {
      * if they have special initialization needed prior being called for flag
      * evaluation.
      * <p>
-     * It is ok, if the method is expensive as it is executed in the background. All
+     * It is ok if the method is expensive as it is executed in the background. All
      * runtime exceptions will be
      * caught and logged.
      * </p>
@@ -46,7 +46,7 @@ public interface FeatureProvider {
      * Providers can overwrite this method, if they have special shutdown actions
      * needed.
      * <p>
-     * It is ok, if the method is expensive as it is executed in the background. All
+     * It is ok if the method is expensive as it is executed in the background. All
      * runtime exceptions will be
      * caught and logged.
      * </p>
@@ -57,7 +57,11 @@ public interface FeatureProvider {
 
     /**
      * Returns a representation of the current readiness of the provider.
-     * Providers which do not implement this method are assumed to be ready immediately.
+     * If the provider needs to be initialized, it should return {@link ProviderState#NOT_READY}.
+     * If the provider is in an error state, it should return {@link ProviderState#ERROR}.
+     * If the provider is functioning normally, it should return {@link ProviderState#READY}.
+     * 
+     * <p><i>Providers which do not implement this method are assumed to be ready immediately.</i></p>
      * 
      * @return ProviderState
      */
--- a/src/test/java/dev/openfeature/sdk/EventProviderTest.java
+++ b/src/test/java/dev/openfeature/sdk/EventProviderTest.java
@@ -121,6 +121,11 @@ class EventProviderTest {
             // TODO Auto-generated method stub
             throw new UnsupportedOperationException("Unimplemented method 'getObjectEvaluation'");
         }
+
+        @Override
+        public ProviderState getState() {
+            return ProviderState.READY;            
+        }
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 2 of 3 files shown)


src/test/java/dev/openfeature/sdk/EventProviderTest.java
```diff
@@ -121,6 +121,11 @@ public ProviderEvaluation<Value> getObjectEvaluation(String key, Value defaultVa
             // TODO Auto-generated method stub
             throw new UnsupportedOperationException("Unimplemented method 'getObjectEvaluation'");
         }
+
+        @Override
+        public ProviderState getState() {
+            return ProviderState.READY;
```
thisthat, line 127: nit
```suggestion
            return ProviderState.READY;
```

thisthat (approved):
I support this change 👍 
Should we provide an annotation to mark specific APIs experimental to feel less guilty when we do breaking changes? 🤔 
Something similar exists in other projects:
- [Guava](https://github.com/google/guava/blob/master/guava/src/com/google/common/annotations/Beta.java)
- [gRPC](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/ExperimentalApi.java)

toddbaert:
> I support this change +1 Should we provide an annotation to mark specific APIs experimental to feel less guilty when we do breaking changes? 🤔 Something similar exists in other projects:
> 
> * [Guava](https://github.com/google/guava/blob/master/guava/src/com/google/common/annotations/Beta.java)
> * [gRPC](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/ExperimentalApi.java)

This is an interesting idea IMO. Can you elaborate more on the user-experience of this? Does the annotation add something that developers can easily see in their IDE? Does it suppose a dev would open the source file?

thisthat:
> This is an interesting idea IMO. Can you elaborate more on the user-experience of this? Does the annotation add something that developers can easily see in their IDE? Does it suppose a dev would open the source file? @thisthat

AFAIK, no IDE prints any special info for annotations. However, when I work on implementing an interface/abstract class I do look in the source file. Annotations are preserved in case of decompilation. It would be just a "FYI" for implementers.

toddbaert:
> > This is an interesting idea IMO. Can you elaborate more on the user-experience of this? Does the annotation add something that developers can easily see in their IDE? Does it suppose a dev would open the source file? @thisthat
> 
> AFAIK, no IDE prints any special info for annotations. However, when I work on implementing an interface/abstract class I do look in the source file. Annotations are preserved in case of decompilation. It would be just a "FYI" for implementers.

OK that's kinda what I was assuming. I still like the idea.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
```diff
@@ -57,7 +57,11 @@ default void shutdown() {
 
     /**
      * Returns a representation of the current readiness of the provider.
-     * Providers which do not implement this method are assumed to be ready immediately.
+     * If the provider needs to be initialized, it should return {@link ProviderState#NOT_READY}.
+     * If the provider is in an error state, it should return {@link ProviderState#ERROR}.
+     * If the provider is functioning normally, it should return {@link ProviderState#READY}.
+     * 
+     * <p><i>Providers which do not implement this method are assumed to be ready immediately.</i></p>
```
toddbaert, line 64: At a minimum we should add these docs, but I think the small breaking change below is worth adding.

beeme1mr (approved):

Kavindu-Dodan (approved):

Kavindu-Dodan:
Approved. We can make this a breaking change and bump the minor version. Given that the eventing support is new, I assume there are only a handful of adoptions of the previous sdk version.

PR #545, opened 2023-08-04
chore: fix race with events tests

toddbaert:
I believe this was causing intermittent failures with tests. This static field on a package private member meant that the instance we created for testing (which we shutdown) was shared with the "real" API instance (see [here](https://github.com/open-feature/java-sdk/actions/runs/5765050607/job/15630249339#step:7:753) for an example failure).

I don't think this constitutes a production bug, but it should make our tests more reliable.

Edit: I ran this suite locally 10+ times and in GH 5 times and it seems to run reliably now.


Kavindu-Dodan (approved):
casued by the test `ShutdownBehaviorSpecTest` where we call executor shutdown

Issue #523, opened 2023-07-21
In-memory provider for e2e testing and minimal SDK usage


Kavindu-Dodan:
As proposed by **OFEP: SDK end-to-end test strategy** [^1], we should create an in-memory provider for Java-sdk

We already have contributions for Python[^2] & Go [^3], and we can use same flag format, constructors & error handling strategy for this implementation.

- [x] Implement the in-memory provider
- [x] Migrate existing e2e tests[^4] to newly create in-memory provider 
  - search repo for `https://github.com/open-feature/java-sdk/issues/523` for related TODOs. 

[^1]: https://github.com/open-feature/ofep/pull/71
[^2]: https://github.com/open-feature/python-sdk/pull/157
[^3]: https://github.com/open-feature/go-sdk/pull/192
[^4]: https://github.com/open-feature/java-sdk/tree/main/src/test/java/dev/openfeature/sdk/e2e

toddbaert:
Small related PR here: https://github.com/open-feature/java-sdk/pull/543

liran2000:
@Kavindu-Dodan @toddbaert as PR merged, can be resolved ?

Kavindu-Dodan:
@liran2000 yes, this is complete. Thank you for working on this 🙌

PR #546, opened 2023-08-07
feat: In-memory provider for e2e testing and minimal usage

liran2000:
## This PR
[Issue 523](https://github.com/open-feature/java-sdk/issues/523) - In-memory provider for e2e testing and minimal SDK usage


src/main/java/dev/openfeature/sdk/FlagEvaluationDetails.java
```diff
@@ -40,6 +40,7 @@ public static <T> FlagEvaluationDetails<T> from(ProviderEvaluation<T> providerEv
                 .value(providerEval.getValue())
                 .variant(providerEval.getVariant())
                 .reason(providerEval.getReason())
+                .errorMessage(providerEval.getErrorMessage())
```
Kavindu-Dodan, line 43: +1 for this fix

src/test/java/dev/openfeature/sdk/testutils/Flags.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private Map<String, Flag> flags;
+
+    public enum State {
+        ENABLED, DISABLED
+    }
+
+    public enum Variant {
+        on, off
+    }
```
Kavindu-Dodan, line 49: I think this is not used ?

src/test/resources/features/testing-flags.json
```diff
@@ -0,0 +1,114 @@
...
+        "empty": {},
+        "template": {
+          "showImages": true,
+          "title": "Check out these pics!",
+          "imagesPerPage": 100
+        }
+      },
+      "defaultVariant": "template"
+    },
+    "context-aware": {
```
Kavindu-Dodan, line 47: we can get rid of  this flag as this rely on JSON logic :)

Kavindu-Dodan:
Overall great work. I would love to get this to mergeable status. Please check comments from my initial review and let me know if you have any concerns

src/test/java/dev/openfeature/sdk/testutils/Flag.java
```diff
@@ -0,0 +1,18 @@
...
+
+@ToString
+@JsonIgnoreProperties(ignoreUnknown = true)
+@NoArgsConstructor
+@Getter
+public class Flag {
+    private Flags.State state;
+    private Map<String, Object> variants;
+    private String defaultVariant;
+}
```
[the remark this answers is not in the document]
liran2000: Trying to understand what is the purpose and value here with enhancing with additional testing capabilities. The in-memory provider is for testing purposes, how is adding a context evaluator helping with testing actual flow ? it will only test the testing provider ?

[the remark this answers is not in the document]
Kavindu-Dodan: The current implementation resides as a `test` source. This mismatch with the requirement of the specification. 

When we agreed on the in-memory provider through OFEP and added it to the specification [^1], the agreement was to make the in-memory an extra provider built into the SDK itself. And we agreed to support evaluation contexts through lambda/callback.

So the packaging of the implementation should be corrected first. I am proposing to move it to a package named `memprovider` or any better-named package.

Regarding the purpose, the main benefit of having evaluation context support is testing SDK. It allows us to write end-to-end tests (or to migrate existing ones based on flagd) for context evaluations and verify SDK correctness. Besides, end users can use the provider to prototype OpenFeature features. 




[^1]: https://github.com/open-feature/spec/blob/main/specification/appendix-A.md#in-memory-provider

CONTRIBUTING.md
```diff
@@ -16,20 +16,6 @@ If you think we might be out of date with the spec, you can check that by invoki
 
 If you're adding tests to cover something in the spec, use the `@Specification` annotation like you see throughout the test suites.
 
-## End-to-End Tests
```
Kavindu-Dodan, line 19: Rather than removing, IMO we should mention that the e2e tests use `InMemoryProvider`. Consider the GO SDK readme for reference - https://github.com/open-feature/go-sdk/blob/main/e2e/README.md

src/test/java/dev/openfeature/sdk/e2e/StepDefinitions.java
```diff
@@ -47,13 +44,21 @@ public class StepDefinitions {
...
     @Given("an openfeature client is registered with cache disabled")
     public static void setup() {
-        // TODO: when the FlagdProvider is updated to support caching, we might need to disable it here for this test to work as expected.
-        FlagdProvider provider = new FlagdProvider();
-        provider.setDeadline(3000); // set a generous deadline, to prevent timeouts in actions
+        ClassLoader classLoader = StepDefinitions.class.getClassLoader();
+        File file = new File(classLoader.getResource("features/testing-flags.json").getFile());
+        Path resPath = file.toPath();
+        String conf = new String(java.nio.file.Files.readAllBytes(resPath), "UTF8");
+        Flags flags = Flags.builder().setConfigurationJson(conf).build();
```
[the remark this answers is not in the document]
liran2000: Leftover, will remove it

Kavindu-Dodan:
Some observations and suggestions for improvement 

In the current implementation, flag configurations are loaded from a string. I think this is not ideal for general usage and diverge from the specification [^1]

> Provider must be initiated with a pre-defined set of flags provided to a constructor

I am suggesting changing the constructor to accept a known map of flags. The end user may use any mechanism to parse their flags and initialize the provider. 

Further, this should allow us to simplify `Flags` class and make it a simple pojo (no builders and parsing)

[^1]: https://github.com/open-feature/spec/blob/main/specification/appendix-A.md#sdk-end-to-end-testing

pom.xml
```diff
@@ -494,62 +486,6 @@
         </plugins>
       </build>
     </profile>
-
```
Kavindu-Dodan, line 497: We do not need to remove this profile. It is still needed and works normally 

Changes you made at StepDefinisions.java [1] is sufficient to fulfill the comment 

[1] - https://github.com/open-feature/java-sdk/pull/546/files#diff-ca6f171ac07058b6e01935480aefe8e8a07379832fc06c334a9ad8887ae3881f

Kavindu-Dodan: Without this section, we do not initialize the test-harness submodule. See the build result  - https://github.com/open-feature/java-sdk/actions/runs/5824793560/job/15795407914?pr=546

liran2000: By this comment it can be removed, if you still think it should remain update here and I can return it.
```
<!-- this profile handles running the flagd e2e tests -->
      <!-- TODO: this profile can likely be removed with TODO: this section should be updated with https://github.com/open-feature/java-sdk/issues/523 -->
      <!-- TODO: we should pull the submodule and run these tests unconditionall once flagd isn't required -->
```

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -0,0 +1,184 @@
...
+            value = (T) flag.getVariants().get(flag.getDefaultVariant());
+        }
+        return ProviderEvaluation.<T>builder()
+            .value(value)
+            .variant(flag.getDefaultVariant())
+            .reason(Reason.STATIC.toString())
+            .build();
+    }
+
+    private static Value objectToValue(Object object) {
```
Kavindu-Dodan, line 143: I think this should be exposed in `Value` class itself. This will help other provider implementations

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    /**
+     * transform a map to a Structure type.
+     *
+     * @param map map of objects
+     * @return a Structure object in the SDK format
+     */
+    public static Structure mapToStructure(Map<String, Object> map) {
```
Kavindu-Dodan, line 178: This can be exposed from `MutableStructure` class so that it can be commonly used where needed

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -0,0 +1,184 @@
+package dev.openfeature.sdk.providers.memory;
```
Kavindu-Dodan, line 1: I think this is a good package name and differentiate from contrib provider implemenations (`dev.openfeature.contrib.providers.*`)

pom.xml
Kavindu-Dodan: yes, we still need them. Tests are now independent from flagd but we need this profile to intialize git submodule and run gherkin tetsts

See the GitHub action - https://github.com/open-feature/java-sdk/blob/main/.github/workflows/pullrequest.yml#L42-L43

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        client = OpenFeatureAPI.getInstance().getClient();
+
+        client.onProviderReady(eventDetails -> isProviderReady.set(true));
+
+        client.onProviderConfigurationChanged(eventDetails -> isConfigurationChanged.set(true));
+        provider.updateFlags(flags);
+    }
+
+    public static Map<String, Flag<?>> buildFlags() {
```
Kavindu-Dodan, line 45: Given the public accessor status & multiple usages, I think this fits well into testutils - https://github.com/open-feature/java-sdk/tree/main/src/test/java/dev/openfeature/sdk/testutils

Probably with a class named `TestFlags`

src/test/java/dev/openfeature/sdk/e2e/StepDefinitions.java
```diff
@@ -1,22 +1,18 @@
...
-import dev.openfeature.contrib.providers.flagd.FlagdProvider;
-import dev.openfeature.sdk.Client;
-import dev.openfeature.sdk.EvaluationContext;
-import dev.openfeature.sdk.FlagEvaluationDetails;
-import dev.openfeature.sdk.ImmutableContext;
-import dev.openfeature.sdk.OpenFeatureAPI;
-import dev.openfeature.sdk.Reason;
-import dev.openfeature.sdk.Structure;
-import dev.openfeature.sdk.Value;
+import dev.openfeature.sdk.*;
```
Kavindu-Dodan, line 3: Please consider avoiding wildcard imports (check other places too)

toddbaert:
I'll take a look at this early next week! Thanks @liran2000

src/main/java/dev/openfeature/sdk/MutableStructure.java
```diff
@@ -108,4 +110,17 @@ public Map<String, Object> asObjectMap() {
                 e -> convertValue(getValue(e.getKey()))
             ));
     }
+
+    /**
+     * transform a map to a Structure type.
```
Kavindu-Dodan, line 115:
```suggestion
     * Transform an object map to a {@link Structure} type.
```

src/test/java/dev/openfeature/sdk/testutils/TestFlagsUtils.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Test flags utils.
+ */
+@UtilityClass
+public class TestFlagsUtils {
+
+    /**
+     * Building flags for testing purposes.
+     * @return
```
Kavindu-Dodan, line 21:
```suggestion
```

Kavindu-Dodan: suggesting the removal as return is not defined

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -0,0 +1,119 @@
...
+            "showImages", new Value(true),
+            "title", new Value("Check out these pics!"),
+            "imagesPerPage", new Value(100)
+        )));
+        assertEquals(expectedObject, client.getObjectValue("object-flag", new Value(true)));
+    }
+
+    @SneakyThrows
+    @Test
+    void mapToStructureTest() {
```
Kavindu-Dodan, line 95: Let's move this to `StructureTest`

Kavindu-Dodan (approved):
Except for the latest reviews, I do not see any other concerns. Approval from my end :)

CONTRIBUTING.md
```diff
@@ -18,17 +18,7 @@ If you're adding tests to cover something in the spec, use the `@Specification`
...
-<!-- TODO: this section should be updated with https://github.com/open-feature/java-sdk/issues/523 -->
-
-The continuous integration runs a set of [gherkin e2e tests](https://github.com/open-feature/test-harness/blob/main/features/evaluation.feature) using [`flagd`](https://github.com/open-feature/flagd). These tests do not run with the default maven profile. If you'd like to run them locally, you can start the flagd testbed with
-
-```
-docker run -p 8013:8013 ghcr.io/open-feature/flagd-testbed:latest
-```
-and then run 
-```
-mvn test -P e2e-test
```
toddbaert, line 30: We might want to keep this line if we are keeping the profile.

pom.xml
toddbaert: I was thinking we could just always pull the submodule and copy the gherkin tests, since now they don't require the external flagd process. I dont really mind if we keep these integration tests on a separate profile or not.

I think if we keep the profile though, we should add back the little bit of documentation in the contributing guide about it.

src/main/java/dev/openfeature/sdk/MutableStructure.java
```diff
@@ -108,4 +110,17 @@ public Map<String, Object> asObjectMap() {
...
+     *
+     * @param map map of objects
+     * @return a Structure object in the SDK format
+     */
+    public static Structure mapToStructure(Map<String, Object> map) {
+        return new MutableStructure(
+            map.entrySet().stream()
+                .filter(e -> e.getValue() != null)
+                .collect(Collectors.toMap(Map.Entry::getKey, e -> objectToValue(e.getValue()))));
+    }
```
toddbaert, line 125: Should this be made generic and be defined on the interface? We want to make sure we maintain as much functionality as possible between `MutableStructure` and `ImmutableStructure`

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -0,0 +1,166 @@
...
+    @Getter
+    private ProviderState state = ProviderState.NOT_READY;
+
+    @Override
+    public Metadata getMetadata() {
+        return () -> NAME;
+    }
+
+    public InMemoryProvider(Map<String, Flag<?>> flags) {
+        this.flags = new HashMap<>(flags);
```
toddbaert, line 44: Good idea to copy this ref.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Initialize the provider.
+     * @param evaluationContext evaluation context
+     * @throws Exception on error
+     */
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        super.initialize(evaluationContext);
+        state = ProviderState.READY;
+        log.info("finished initializing provider, state: {}", state);
```
toddbaert, line 55: I would make this a debug.

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        OpenFeatureAPI.getInstance().setProvider(provider);
+
+        // TODO: setProvider with wait for init, pending https://github.com/open-feature/ofep/pull/80
+        Thread.sleep(500);
+
+        client = OpenFeatureAPI.getInstance().getClient();
+
+        client.onProviderReady(eventDetails -> isProviderReady.set(true));
+
+        client.onProviderConfigurationChanged(eventDetails -> configurationChangedEventCount.incrementAndGet());
```
toddbaert, line 42: Just curious why you `incrementAndGet` here, and then assert `2` later, instead of just incrementing here and asserting 1 later? Is that testing something that I don't understand?

src/test/java/dev/openfeature/sdk/testutils/TestFlagsUtils.java
```diff
@@ -0,0 +1,73 @@
...
+ * Test flags utils.
+ */
+@UtilityClass
+public class TestFlagsUtils {
+
+    /**
+     * Building flags for testing purposes.
+     * @return map of flags
+     */
+    public static Map<String, Flag<?>> buildFlags() {
```
toddbaert, line 23: I think this API looks pretty solid. It's what I expected. Interested in @justinabrahms @thisthat 's takes.

toddbaert (changes requested):
Overall this looks really good, thanks @liran2000 !!

I have a few small comments, in particular [this one](https://github.com/open-feature/java-sdk/pull/546#discussion_r1293577949) (unless I'm missing something). Also watch out for the Javadoc stuff I mentioned to make sure people don't accidentally think some of these classes are for general SDK usage.

I think with a few small changes this can be merged!

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
liran2000: line 43: _updateFlags(flags)_ - update config by replacing all flags
line 44: _updateFlag("addedFlag", Flag.builder()...)_ - update a single flag

each of these method expected to emit ProviderConfigurationChanged event - this provides test coverage including events for each of the 2 methods.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
toddbaert: Sorry, I wasn't clear. I was saying the way you did it was a good idea instead of copying it. It's a good idea exactly because it prevents the bug :). You already did it right :smile:

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -0,0 +1,88 @@
...
+            .variant("off", false)
+            .defaultVariant("on")
+            .build());
+    }
+
+    @SneakyThrows
+    @Test
+    void eventsTest() {
+        assertTrue(isProviderReady.get());
+        assertEquals(2, configurationChangedEventCount.get());
```
toddbaert, line 55: There's an inherent race condition in this test. Here is an example failure: https://github.com/open-feature/java-sdk/actions/runs/5857702730/job/15882568501#step:6:150

This is because events run in a new thread and don't block, so it's possible for the `beforeAll` to fire the event, but the callback hasn't finished before this test runs.

I would recommend you use some sort of retry with a timeout, like Mochito's verify, like we do [here](https://github.com/open-feature/java-sdk/blob/b17a20e43fb2446da3f2b07d7a9d577a9b1dffea/src/test/java/dev/openfeature/sdk/EventsTest.java#L97). It should be true within a few ms.

liran2000: From inspecting, I think the issue was different here, you can update if I missed anything.
1. Handlers were set after setting the provider - see the [change](https://github.com/open-feature/java-sdk/pull/546/commits/41987b83b0657e3376980bc3ab44dab610dd18a0#diff-08b7a651a304317563611f9d4260b2f5e2ec554ba9fa6ae0e9a19a52efbd0d04).
This caused the possible race condition, as initialization was done in background, then calling the handler - which is set later.
2. "events run in a new thread and don't block" - I understand this is for provider ready event. But following the flow of _ProviderConfigurationChanged - provider.updateFlags(flags) -> emitProviderConfigurationChanged(details); -> this.onEmit.accept(this, event, details);_
I see this as direct non-blocking event call in same thread. Different provider implementation can to call the emit in a new thread.
3. I revisited the events in my last commit, you can re-review.

 As an early adopter of provider events with these issues, have some thoughts:
a. as (1) above was sort of implementation confusion, is there some way to enforce it  by the sdk ? should it be enforced ?
for example, when adding a new handler - check first if provider is set, otherwise throw exception ?
Should consider multiple / named providers though.
b. From EventProvider javadoc: "Please note that the SDK will automatically emit PROVIDER_READY" - I confused with it previously and manually emitted it without reading it possibly - can/should it be enforced by sdk ?
c. events are same for all providers in case of multiple named providers ?
d. I could call _OpenFeatureAPI.getInstance().onProviderReady_ and also _OpenFeatureAPI.getInstance().getClient().onProviderReady_. Why both are exposed ? Can you elaborate a bit on what is the difference between these ?
e. per (2) above, should the sdk emit these on the background ?

what do you think ?

toddbaert: > Handlers were set after setting the provider - see the [change](https://github.com/open-feature/java-sdk/pull/546/commits/41987b83b0657e3376980bc3ab44dab610dd18a0#diff-08b7a651a304317563611f9d4260b2f5e2ec554ba9fa6ae0e9a19a52efbd0d04).
This caused the possible race condition, as initialization was done in background, then calling the handler - which is set later.

OK then this is working as expected. We want to support setting new providers, and we also want to allow users to add handlers at any time (before or after setting the provider) and for handlers to "move" to the new provider.

> 2.
> "events run in a new thread and don't block" - I understand this is for provider ready event. But following the flow of ProviderConfigurationChanged - provider.updateFlags(flags) -> emitProviderConfigurationChanged(details); -> this.onEmit.accept(this, event, details);
I see this as direct non-blocking event call in same thread. Different provider implementation can to call the emit in a new thread.

No matter where, "emit" happens, the SDK runs the _handler function_ in a new thread, see [here](https://github.com/open-feature/java-sdk/blob/b17a20e43fb2446da3f2b07d7a9d577a9b1dffea/src/main/java/dev/openfeature/sdk/EventSupport.java#L134).

> b. From EventProvider javadoc: "Please note that the SDK will automatically emit PROVIDER_READY" - I confused with it previously and manually emitted it without reading it possibly - can/should it be enforced by sdk ?

Even though the SDK automatically emits this event after `initialize()`, we still need to allow provider authors to sometimes emit it manually... for example, if the provider disconnects later and goes into ERROR state, it will then emit READY again when it reconnects. I'm not sure what we can do to prevent people from manually emitting it after init, which is not needed, as you say.

> c. events are same for all providers in case of multiple named providers ?
d. I could call OpenFeatureAPI.getInstance().onProviderReady and also OpenFeatureAPI.getInstance().getClient().onProviderReady. Why both are exposed ? Can you elaborate a bit on what is the difference between these ?

`Clients` are scoped to providers. The API is not. The API gets _all_ events. That's the difference between these. Client's are always only associated with a single provider, and evaluate flags from just that provider, and get events from just that provider.

> e. per (2) above, should the sdk emit these on the background ?

It's the provider that emits the events, which are "fire and forget" from it's perspective. They handlers associated with events run on a separate thread.

toddbaert: > thank you for clarifying !
> I will try revisiting the tests for the issues.

Absolutely no problem! I think GH might be struggling right now though :grimacing: things seem to be queuing up and not completing :sweat_smile:

toddbaert (approved):
Implementation looks good to me, and tests look stable now. Thanks again @liran2000 

Approving. I will wait until EOD before merging to give others a chance to review.

toddbaert:
@liran2000 I opened a small PR to satisfy codecov: https://github.com/open-feature/java-sdk/pull/561, and made some other minor changes I didn't notice here.

PR #561, opened 2023-08-15
chore: fix jacoco coverage minimum, throw in memory provider

toddbaert:
Small changes to satisfy jacoco coverage failing in main (bring above 80%).

Also throwing in the new in-memory provider instead of returning error resolutions.

cc @liran2000 I missed these small things on your PR. I hope they make sense.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -1,25 +1,28 @@
 package dev.openfeature.sdk.providers.memory;
 
-import dev.openfeature.sdk.Value;
-import dev.openfeature.sdk.Metadata;
-import dev.openfeature.sdk.EventProvider;
-import dev.openfeature.sdk.ProviderState;
-import dev.openfeature.sdk.ProviderEventDetails;
-import dev.openfeature.sdk.ErrorCode;
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -87,68 +90,52 @@ public void updateFlag(String flagKey, Flag<?> flag) {
...
     private <T> ProviderEvaluation<T> getEvaluation(
-            String key, T defaultValue, EvaluationContext evaluationContext, Class<?> expectedType
+            String key, EvaluationContext evaluationContext, Class<?> expectedType
     ) throws OpenFeatureError {
         if (!ProviderState.READY.equals(state)) {
-            ErrorCode errorCode = ErrorCode.PROVIDER_NOT_READY;
-            if (ProviderState.ERROR.equals(state)) {
-                errorCode = ErrorCode.GENERAL;
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError("provider not yet initialized");
```
toddbaert, line 126: @liran2000 throwing like this ends in the same result, but it's a bit easier to trace, and also simpler to write. This is how we generally recommend it's done. There is a small performance cost in serialization the exception, but I think it's worth it.

beeme1mr (approved):

Kavindu-Dodan (approved):

PR #563, opened 2023-08-16
feat: add method to set provider and block during init

liran2000:
The API SHOULD provide functions to set a provider and wait for the initialize function to return or throw

## This PR
implements spec [Requirement 1.1.8](https://github.com/open-feature/spec/blob/ee80ce2c0dd4d2e2172638f5f3ff225989a85362/specification/sections/01-flag-evaluation.md#requirement-118).

following [blocking setProvider PR](https://github.com/open-feature/spec/pull/201).


liran2000:
hi @toddbaert,

on the approach to implement it, I thought it can be good in this case to call initialize on the same caller thread as it needs to wait.

@thomaspoignant for your recent spec addition.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -55,6 +55,18 @@ public boolean isDefaultProvider(FeatureProvider provider) {
...
+            Consumer<FeatureProvider> afterSet,
+            Consumer<FeatureProvider> afterInit,
+            Consumer<FeatureProvider> afterShutdown,
+            BiConsumer<FeatureProvider, String> afterError,
+            boolean waitForInit) {
+        if (provider == null) {
+            throw new IllegalArgumentException("Provider cannot be null");
+        }
+        initializeProvider(null, provider, afterSet, afterInit, afterShutdown, afterError, waitForInit);
+    }
```
toddbaert, line 68: This is a package private class, so we _could_ also just have one constructor, and update calls to use that, passing nulls/defaults where needed. I don't mind either way though.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -81,39 +106,72 @@ public void setProvider(String clientName,
...
-                }
-                shutDownOld(oldProvider, afterShutdown);
-            } catch (Exception e) {
-                log.error("Exception when initializing feature provider {}", newProvider.getClass().getName(), e);
-                afterError.accept(newProvider, e.getMessage());
+        if (waitForInit) {
+            initializeProvider(newProvider, afterInit, afterShutdown, afterError, oldProvider);
+        } else {
+            taskExecutor.submit(() -> {
+                // initialization happens in a different thread
```
toddbaert, line 154:
```suggestion
                // initialization happens in a different thread if we're not waiting it
```

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -81,39 +106,72 @@ public void setProvider(String clientName,
...
+            if (ProviderState.NOT_READY.equals(newProvider.getState())) {
+                newProvider.initialize(OpenFeatureAPI.getInstance().getEvaluationContext());
+                afterInit.accept(newProvider);
             }
-        });
+            shutDownOld(oldProvider, afterShutdown);
+        } catch (Exception e) {
+            log.error("Exception when initializing feature provider {}", newProvider.getClass().getName(), e);
+            afterError.accept(newProvider, e.getMessage());
+        }
```
toddbaert, line 174: Could you either combine both `initializeProvider` methods, or name them differently? I find overloading these a bit confusing in this case since they are doing different things (one is actually initializing, while the other is just deciding to block or not).

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -31,11 +31,7 @@ static void beforeAll() {
         Map<String, Flag<?>> flags = buildFlags();
         provider = spy(new InMemoryProvider(flags));
         OpenFeatureAPI.getInstance().onProviderConfigurationChanged(eventDetails -> {});
-        OpenFeatureAPI.getInstance().setProvider(provider);
-
-        // TODO: setProvider with wait for init, pending https://github.com/open-feature/ofep/pull/80
```
toddbaert, line 36: :+1:

src/main/java/dev/openfeature/sdk/ProviderRepository.java
liran2000: cool, then no need to worry about breaking changes on public methods, I'll update it.

So in this case of package private class, the methods can be be changed to non-public as well, right ?

toddbaert (changes requested):
Looks good! Thanks! Interested in @thomaspoignant 's opinion.

My only requested change is [this](https://github.com/open-feature/java-sdk/pull/563/files#r1296193754), which is a naming issue mostly (let me know if you disagree). Otherwise just optional suggestions.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
toddbaert: > So in this case of package private class, the methods can be be changed to non-public as well, right ?

In Java, I'm of the opinion that with package private classes, certain methods should still be public. This makes it clear which ones are intended to be used from the outside. This is a style choice though.

toddbaert:
> hi @toddbaert,
> 
> on the approach to implement it, I thought it can be good in this case to call initialize on the same caller thread as it needs to wait.
> 
> @thomaspoignant for your recent spec addition.

I'm fine with this approach.

toddbaert:
@liran2000 I've updated the PR title to be more "friendly" since it goes into our release notes.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -100,11 +100,12 @@ public EvaluationContext getEvaluationContext() {
...
-                    provider,
-                    (p) -> attachEventProvider(p),
-                    (p) -> emitReady(p),
-                    (p) -> detachEventProvider(p),
-                    (p, message) -> emitError(p, message));
+                provider,
+                (p) -> attachEventProvider(p),
+                (p) -> emitReady(p),
+                (p) -> detachEventProvider(p),
+                (p, message) -> emitError(p, message),
```
toddbaert, line 107: nit: use method refs

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -117,11 +118,45 @@ public void setProvider(FeatureProvider provider) {
...
+     * Set the default provider and wait for initialization to finish.
+     */
+    public void setProviderAndWait(FeatureProvider provider) {
+        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
+            providerRepository.setProvider(
+                provider,
+                (p) -> attachEventProvider(p),
+                (p) -> emitReady(p),
+                (p) -> detachEventProvider(p),
+                (p, message) -> emitError(p, message),
```
toddbaert, line 140: nit: use method refs

toddbaert (approved):
LGTM, left 2 more small nits!

Kavindu-Dodan (approved):

toddbaert:
@thomaspoignant since you came up with this spec proposal, I'm interested in your opinion of this implementation.

justinabrahms (approved):

Issue #600, opened 2023-09-08
[Bug] SDK prevents JVM shutdown when shutdown is not called


Kavindu-Dodan:
**Description**

Consider below example where Java SDK is being used, 

```Java
import dev.openfeature.sdk.OpenFeatureAPI;
import dev.openfeature.sdk.providers.memory.InMemoryProvider;

import java.util.HashMap;

public class Main {
    public static void main(String[] args) throws InterruptedException {
        OpenFeatureAPI api = OpenFeatureAPI.getInstance();

        api.setProvider(new InMemoryProvider(new HashMap<>()));

        // api.shutdown();

        Runtime.getRuntime()
                .addShutdownHook(new Thread(() ->
                System.out.println("exit")));
    }
}
```

If you execute this program, your JVM will never exit and the shutdown hook will not get executed.

 However, if we uncomment `api.shutdown()`, the JVM will exit and shutdown hook get exectued.
 
 The reason for this behavior is the task executor that helps event handling of the SDK [^1] and the provider repository's task executor [^2]
 
 **Solution**
 
 The task executor should utilize deamon threads so that SDK does not block abrupt program exit requirements. 
 
 [^1]: https://github.com/open-feature/java-sdk/blob/v1.6.0/src/main/java/dev/openfeature/sdk/EventSupport.java#L26
 [^2]: https://github.com/open-feature/java-sdk/blob/v1.6.0/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L25

PR #601, opened 2023-09-08
fix: make executor threads deamon

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/java-sdk/issues/600

Makes task executor threads daemons.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/EventSupport.java
+++ b/src/main/java/dev/openfeature/sdk/EventSupport.java
@@ -1,8 +1,8 @@
 package dev.openfeature.sdk;
 
-import edu.umd.cs.findbugs.annotations.Nullable;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


toddbaert (approved):
Great idea!

src/main/java/dev/openfeature/sdk/EventSupport.java
```diff
@@ -1,5 +1,8 @@
 package dev.openfeature.sdk;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
```
toddbaert, line 3: I think this is the wrong import... not your change but can you fix it?

Kavindu-Dodan: yup, will fix that to have consistency across other usages

Kavindu-Dodan: done :)

toddbaert (approved):

PR #603, opened 2023-09-11
feat: provider state check

liran2000:
According to the [spec](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-245):
>Requirement 2.4.5
The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.
It's recommended to set an informative error code, such as PROVIDER_NOT_READY if evaluation in attempted before the provider is initialized.

Problem:
Provider state check is implemented per provider.
Solution:
Provider state check is implemented at core SDK.

### Notes

- Hooks:
May be considered as some "behavior change", as in case of provider not ready, error is thrown, and hooks are not called, which is similar behavior to other errors handling.
- This is raised by several developers when implementing a provider, and suggested [here](https://github.com/open-feature/java-sdk-contrib/pull/424#discussion_r1321295710) by @ivarconr.
- Some providers implementation may have to adjust implementation and test for this. If implementation is not adopted, this is not expected to break something.

@toddbaert @Kavindu-Dodan @thomaspoignant

### DIFF
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -9,6 +9,7 @@ import java.util.function.Consumer;
 
 import dev.openfeature.sdk.exceptions.GeneralError;
 import dev.openfeature.sdk.exceptions.OpenFeatureError;
+import dev.openfeature.sdk.exceptions.ProviderNotReadyError;
 import dev.openfeature.sdk.internal.AutoCloseableLock;
 import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
 import dev.openfeature.sdk.internal.ObjectUtils;
@@ -163,6 +164,14 @@ public class OpenFeatureClient implements Client {
             T defaultValue,
             FeatureProvider provider,
             EvaluationContext invocationContext) {
+
+        if (!ProviderState.READY.equals(provider.getState())) {
+            if (ProviderState.NOT_READY.equals(provider.getState())) {
+                throw new ProviderNotReadyError("provider not yet initialized");
+            }
+            throw new GeneralError("unknown error");
+        }
+
         switch (type) {
             case BOOLEAN:
                 return provider.getBooleanEvaluation(key, (Boolean) defaultValue, invocationContext);
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -16,9 +16,7 @@ import dev.openfeature.sdk.ProviderState;
 import dev.openfeature.sdk.Reason;
 import dev.openfeature.sdk.Value;
 import dev.openfeature.sdk.exceptions.FlagNotFoundError;
-import dev.openfeature.sdk.exceptions.GeneralError;
 import dev.openfeature.sdk.exceptions.OpenFeatureError;
-import dev.openfeature.sdk.exceptions.ProviderNotReadyError;
 import dev.openfeature.sdk.exceptions.TypeMismatchError;
 import lombok.Getter;
 import lombok.SneakyThrows;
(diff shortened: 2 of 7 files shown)


Kavindu-Dodan:
> @liran2000 there are other [provider states](https://openfeature.dev/specification/types#provider-status) as well (ERROR) for instance, and STALE is added in v0.7.0.
> 
> I don't think STALE is a problem for this idea - but if we're going to automatically check this state, should we check for the ERROR state as well? That seems difficult because we don't know the nature of the error. It seems inconsistent for the SDK to handle the one and not the other. What do you think?
> 
> @Kavindu-Dodan @lukas-reining @beeme1mr what do you guys think about this as a concept?

The way I see the spec section `Requirement 2.4.5` is that this must be enforced by the provider but not by the SDK itself. The reason is that I see SDK being a facilitator for events but not necessarily an enforcer 🤔  IMO this helps OF SDK implementation and maintenance effforts

liran2000:
Error state is handled by going to GeneralError, but yes, ERROR and STALE can be added to the code in a possibly adjusted way.

I will try to emphasize again the need for this, you can share if something is not accurate.
Currently each provider has to implement something like the following [code block](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java#L261):

```
if (!ProviderState.READY.equals(state)) {
	if (ProviderState.NOT_READY.equals(state)) {
		/*
		 should be handled by the SDK framework, ErrorCode.PROVIDER_NOT_READY and default value
		 should be returned when evaluated via the client.
		 */
		throw new ProviderNotReadyError("provider not initialized yet");
	}
	throw new GeneralError("unknown error, provider state: " + state);
}
```

This create 2 problems:
1. The code is copied and duplicated per provider implementation.
2. It is not trivial and clear whether this code should be implemented by throwing errors, or by returning evaluation with relevant state/status. This question raised by several developers,  recent example from [here](https://github.com/open-feature/java-sdk-contrib/pull/424#discussion_r1321104957), and I also encountered it where it was clear to me when implementing a provider.

This solution comes to solve these problems.

Possible down side of the solution is to have less control of flow on these states, so I understand the conflict here.

Do you possibly have some other ideas for these problems ?

toddbaert:
> I will try to emphasize again the need for this, you can share if something is not accurate. Currently each provider has to implement something like the following [code block](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/GoFeatureFlagProvider.java#L261):
> 
> ```
> if (!ProviderState.READY.equals(state)) {
> 	if (ProviderState.NOT_READY.equals(state)) {
> 		/*
> 		 should be handled by the SDK framework, ErrorCode.PROVIDER_NOT_READY and default value
> 		 should be returned when evaluated via the client.
> 		 */
> 		throw new ProviderNotReadyError("provider not initialized yet");
> 	}
> 	throw new GeneralError("unknown error, provider state: " + state);
> }
> ```
> 
> This create 2 problems:
> 
> 1. The code is copied and duplicated per provider implementation.
> 2. It is not trivial and clear whether this code should be implemented by throwing errors, or by returning evaluation with relevant state/status. This question raised by several developers,  recent example from [here](https://github.com/open-feature/java-sdk-contrib/pull/424#discussion_r1321104957), and I also encountered it where it was clear to me when implementing a provider.

I totally agree this is not idea, and there is a lot I like about this proposal. I would like a solution, but the solution can't introduce more problems, and I'm _worried_ this one might (but I'm not sure).

> Possible down side of the solution is to have less control of flow on these states, so I understand the conflict here.

> Error state is handled by going to GeneralError, but yes, ERROR and STALE can be added to the code in a possibly adjusted way.

Yes. The problem I see is that we also need to handle ERROR state to be consistent - but if we do that, how can the provider give more details about the error? If the SDK just throws `GeneralError` we have less information from the provider than we do now. Maybe that's OK? The provider can log things, of course.

Basically, the more logic we move into the SDK, the less providers have to implement (good), but the less control we give providers (bad). Striking this balance is hard.

PR #663, opened 2023-10-23
fix: null handling with Structure, Value

toddbaert:
* fix NPEs with Structure.asObjectMap and Structure.convertValue
* more consistent null handling in structures (empty `Value` -> `null` and vice versa)
* update spotbugs and fix a few new detected issues


src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -79,9 +79,10 @@ public Map<String, Object> asObjectMap() {
...
                 .stream()
-                .collect(Collectors.toMap(
-                        Map.Entry::getKey,
-                        e -> convertValue(getValue(e.getKey()))
-                ));
+                // custom collector, workaround for Collectors.toMap in JDK8
+                // https://bugs.openjdk.org/browse/JDK-8148463
+                .collect(HashMap::new,
+                        (accumulated, entry) -> accumulated.put(entry.getKey(), convertValue(entry.getValue())),
+                        HashMap::putAll);
```
toddbaert, line 86: `Collectors.toMap` doesn't handle null in Java 8! (see bug)

src/main/java/dev/openfeature/sdk/Value.java
```diff
@@ -19,11 +19,15 @@
...
 @EqualsAndHashCode
-@SuppressWarnings({"PMD.BeanMembersShouldSerialize", "checkstyle:MissingJavadocType"})
+@SuppressWarnings({"PMD.BeanMembersShouldSerialize", "checkstyle:MissingJavadocType", "checkstyle:NoFinalizer"})
 public class Value implements Cloneable {
 
     private final Object innerObject;
 
+    protected final void finalize() {
+        // DO NOT REMOVE, spotbugs: CT_CONSTRUCTOR_THROW
+    }
```
toddbaert, line 29: The other thing we could do here is make the class final. See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ct-be-wary-of-letting-constructors-throw-exceptions-ct-constructor-throw

spotbugs-exclusions.xml
```diff
@@ -26,6 +26,26 @@
         <Class name="dev.openfeature.sdk.OpenFeatureAPI"/>
         <Bug pattern="EI_EXPOSE_REP2"/>
     </And>
+    <And>
+        Added in spotbugs 4.8.0 - EventProvider shares a name with something from the standard lib (confuising), but change would be breaking
```
beeme1mr, line 30:
```suggestion
        Added in spotbugs 4.8.0 - EventProvider shares a name with something from the standard lib (confusing), but change would be breaking
```

spotbugs-exclusions.xml
```diff
@@ -26,6 +26,26 @@
...
+        <Class name="dev.openfeature.sdk.EventProvider"/>
+        <Bug pattern="PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES"/>
+    </And>
+    <And>
+        Added in spotbugs 4.8.0 - Metadata shares a name with something from the standard lib (confuising), but change would be breaking
```
beeme1mr, line 35:
```suggestion
        Added in spotbugs 4.8.0 - Metadata shares a name with something from the standard lib (confusing), but change would be breaking
```

spotbugs-exclusions.xml
```diff
@@ -26,6 +26,26 @@
...
+        <Class name="dev.openfeature.sdk.Metadata"/>
+        <Bug pattern="PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES"/>
+    </And>
+    <And>
+        Added in spotbugs 4.8.0 - Reason shares a name with something from the standard lib (confuising), but change would be breaking
```
beeme1mr, line 40:
```suggestion
        Added in spotbugs 4.8.0 - Reason shares a name with something from the standard lib (confusing), but change would be breaking
```

spotbugs-exclusions.xml
```diff
@@ -26,6 +26,26 @@
...
+        <Class name="dev.openfeature.sdk.Reason"/>
+        <Bug pattern="PI_DO_NOT_REUSE_PUBLIC_IDENTIFIERS_CLASS_NAMES"/>
+    </And>
+    <And>
+        Added in spotbugs 4.8.0 - FlagValueType.STRING shares a name with something from the standard lib (confuising), but change would be breaking
```
beeme1mr, line 45:
```suggestion
        Added in spotbugs 4.8.0 - FlagValueType.STRING shares a name with something from the standard lib (confusing), but change would be breaking
```

src/main/java/dev/openfeature/sdk/Value.java
Kavindu-Dodan: I think making `Value` final makes sense. Let's do that if you also agree

toddbaert: I think it's better to do that as well, and will probably satisfy sonar also.

My only concern was I think you could consider this a breaking change, though I highly doubt anyone would be extending Value, and it's not an expected use of the SDK.

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
```diff
@@ -79,9 +79,10 @@ public Map<String, Object> asObjectMap() {
...
     }
-}
+}
```
thisthat, line 88: nit
```suggestion
}

```

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
thisthat: wouldn't the `asMap()` and the constructor also suffer from the same problem? 🤔

src/main/java/dev/openfeature/sdk/Value.java
thisthat: I would extend the class to help me handle my special `Object` type instead of being limited by `List`, `Instant` and so on.

toddbaert: Ok so you think it's better to just add this final finalizer?

src/main/java/dev/openfeature/sdk/ImmutableStructure.java
toddbaert: @thisthat @Kavindu-Dodan I found _lots_ of instances of this. I've fixed them all. Fortunately we have good test coverage so even when I made some mistakes with the immutability (forgot to `.clone()`), the tests caught it.

src/main/java/dev/openfeature/sdk/Structure.java
```diff
@@ -134,7 +138,9 @@ default <T extends Structure> Map<String, Value> merge(Function<Map<String, Valu
      */
     static Structure mapToStructure(Map<String, Object> map) {
         return new MutableStructure(map.entrySet().stream()
-            .filter(e -> e.getValue() != null)
```
toddbaert, line 137: We were FILTERING null values before. This resulted in maps with null entries not having converted (empty) `Values`. I think that's wrong. With this change, all nulls in maps are converted to null-containing `Values` and vice-versa.

src/main/java/dev/openfeature/sdk/AbstractStructure.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public Map<String, Object> asObjectMap() {
+        return attributes
+                .entrySet()
+                .stream()
+                // custom collector, workaround for Collectors.toMap in JDK8
+                // https://bugs.openjdk.org/browse/JDK-8148463
+                .collect(HashMap::new,
+                        (accumulated, entry) -> accumulated.put(entry.getKey(), convertValue(entry.getValue())),
+                        HashMap::putAll);
```
toddbaert, line 33: This was common to both the `MutableContext` and `ImmutableContext`, so I've extracted it to an abstract.

src/test/java/dev/openfeature/sdk/StructureTest.java
```diff
@@ -98,6 +98,20 @@ void mapToStructureTest() {
         assertEquals(new Value(Instant.ofEpochSecond(0)), res.getValue("Instant"));
         assertEquals(new HashMap<>(), res.getValue("Map").asStructure().asMap());
         assertEquals(new Value(immutableContext), res.getValue("ImmutableContext"));
-        assertNull(res.getValue("nullKey"));
+        assertEquals(new Value((String)null), res.getValue("nullKey"));
```
toddbaert, line 101: Modified assertion associated with [this](https://github.com/open-feature/java-sdk/pull/663/files#r1370486094) change.

toddbaert, line 101:
```suggestion
        assertEquals(new Value(), res.getValue("nullKey"));
```

src/test/java/dev/openfeature/sdk/StructureTest.java
```diff
@@ -98,6 +98,20 @@ void mapToStructureTest() {
         assertEquals(new Value(Instant.ofEpochSecond(0)), res.getValue("Instant"));
         assertEquals(new HashMap<>(), res.getValue("Map").asStructure().asMap());
         assertEquals(new Value(immutableContext), res.getValue("ImmutableContext"));
-        assertNull(res.getValue("nullKey"));
+        assertEquals(new Value(), res.getValue("nullKey"));
```
toddbaert, line 101: Modified assertion associated with [this](https://github.com/open-feature/java-sdk/pull/663/files#r1370486094) change.

src/main/java/dev/openfeature/sdk/Value.java
Kavindu-Dodan: Fine with me

Kavindu-Dodan (approved):

thisthat (approved):

PR #725, opened 2023-12-11
fix: tolerate duplicate provider registrations

Kavindu-Dodan:
## This PR

Fixes a bug that prevented the same provider from being registered with multiple names (or the same provider as default and with another name).

Note that I have not changed the behavior of `EventProvider.java` where it throws if multiple eventing attachments were attempted be to registered.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -106,15 +106,15 @@ class ProviderRepository {
               BiConsumer<FeatureProvider, String> afterError,
               boolean waitForInit) {
 
-        // provider is set immediately, on this thread
-        FeatureProvider oldProvider = clientName != null
-                ? this.providers.put(clientName, newProvider)
-                : this.defaultProvider.getAndSet(newProvider);
-
         if (!isProviderRegistered(newProvider)) {
             // only run afterSet if new provider is not already attached
             afterSet.accept(newProvider);
         }
+
+        // provider is set immediately, on this thread
+        FeatureProvider oldProvider = clientName != null
+                ? this.providers.put(clientName, newProvider)
+                : this.defaultProvider.getAndSet(newProvider);
 
         if (waitForInit) {
             initializeProvider(newProvider, afterInit, afterShutdown, afterError, oldProvider);
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -108,9 +108,14 @@ class ProviderRepository {
 
         // provider is set immediately, on this thread
         FeatureProvider oldProvider = clientName != null
-            ? this.providers.put(clientName, newProvider)
-            : this.defaultProvider.getAndSet(newProvider);
-        afterSet.accept(newProvider);
+                ? this.providers.put(clientName, newProvider)
+                : this.defaultProvider.getAndSet(newProvider);
+
+        if (!isProviderRegistered(newProvider)) {
+            // only run afterSet if new provider is not already attached
+            afterSet.accept(newProvider);
+        }
+
         if (waitForInit) {
             initializeProvider(newProvider, afterInit, afterShutdown, afterError, oldProvider);
         } else {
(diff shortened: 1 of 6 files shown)


src/test/java/dev/openfeature/sdk/OpenFeatureAPITest.java
```diff
@@ -40,6 +44,20 @@ void namedProviderOverwrittenTest() {
                 .isEqualTo(DoSomethingProvider.name);
     }
 
+    @Test
+    void providerToMultipleNames() {
+        FeatureProvider provider = new InMemoryProvider(Collections.EMPTY_MAP);
```
Kavindu-Dodan, line 49: `InMemoryProvider` to have the behaviour of an eventing provider

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -106,11 +106,16 @@ private void prepareAndInitializeProvider(@Nullable String clientName,
...
+        }
+
         // provider is set immediately, on this thread
         FeatureProvider oldProvider = clientName != null
-            ? this.providers.put(clientName, newProvider)
-            : this.defaultProvider.getAndSet(newProvider);
-        afterSet.accept(newProvider);
+                ? this.providers.put(clientName, newProvider)
+                : this.defaultProvider.getAndSet(newProvider);
+
```
toddbaert, line 118: Is there a reason you changed the order of these?

toddbaert (approved):
[this](https://github.com/open-feature/java-sdk/pull/725/files#r1423117011) is my only concern. If it works, I think you should bring the order back to what it was before.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
Kavindu-Dodan: This is a tricky logic and I couldn't find any good alternative. If we execute `isProviderRegistered` after the collection update, we always get `true` value, making afterSet.accept() not to execute. 

And there's no other good mechanism to detect if the provider is already registered/operational. So I had to make this change.

toddbaert: OK - no tests fail so I assume this has no unintended side-effects.

Kavindu-Dodan: It's tests that helped me to improve the logic. Earlier, I had the order which you proposed. But then with tests, I realised we can't check for registration after updating the collections as registration checks rely on the collection content.

src/test/java/dev/openfeature/sdk/ProviderRepositoryTest.java
```diff
@@ -9,6 +9,7 @@
 import static org.awaitility.Awaitility.await;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.atMostOnce;
```
thisthat, line 12: nit: I don't think we need this
maybe we can add spotless or checkstyle in our CI

src/test/java/dev/openfeature/sdk/OpenFeatureAPITest.java
thisthat: maybe worth adding also a `FeatureProvider nop = new NoOpProvider();` to `clientC` and verify that it is registered correctly

Kavindu-Dodan: @thisthat thanks, updated the test to have non eventing noop provider as well - https://github.com/open-feature/java-sdk/pull/725/files#diff-6d6f404f03467a33bd72fef14be09ac8e70503ab9d4288b01ab179aeda1bb88eR48-R64

src/test/java/dev/openfeature/sdk/ProviderRepositoryTest.java
Kavindu-Dodan: nice catch 👍  

We already have checkstyle enabled but the rule to validate the unused imports were missing. I added with this in the PR - https://github.com/open-feature/java-sdk/pull/725/files#diff-b36112ff71ee5b411c9149781cee470c6413dbf498272fa70eb00b8b7555207fR67

beeme1mr (approved):

PR #739, opened 2023-12-20
perf: improve error handling

thisthat:
## This PR

While working with OF, I noticed that in case a flag is not found, the whole stack trace is attached to the exception.
I think this info is not necessary for this type of exception since it is hidden from the end user, and the SDK only requires the `ErrorCode` field. 

I quickly configured some JMH tests that use the InMemory FlagD provider to test the performance gains in evaluating a flag that doesn't exist. The Score value represents the amount of FF evaluations done in a single millisecond. 
In a single thread case, the performance gain is >60%. 

### Before

```
"Benchmark","Mode","Threads","Samples","Score","Score Error (99,9%)","Unit"
"booleanNotFound_01Thread","thrpt",1,10,"459,324808","8,254488","ops/ms"
"booleanNotFound_02Thread","thrpt",2,10,"791,244898","8,644711","ops/ms"
"booleanNotFound_05Thread","thrpt",5,10,"1465,736432","17,059916","ops/ms"
```

### After

```
"Benchmark","Mode","Threads","Samples","Score","Score Error (99,9%)","Unit"
"booleanNotFound_01Thread","thrpt",1,10,"765,474969","17,161956","ops/ms"
"booleanNotFound_02Thread","thrpt",2,10,"1159,854112","23,483354","ops/ms"
"booleanNotFound_05Thread","thrpt",5,10,"1828,460599","45,932954","ops/ms"
```

### DIFF
--- a/src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
+++ b/src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
@@ -11,7 +11,7 @@ public class FlagNotFoundError extends OpenFeatureError {
     @Getter private final ErrorCode errorCode = ErrorCode.FLAG_NOT_FOUND;
 
     @Override
-    public Throwable fillInStackTrace() {
+    public synchronized Throwable fillInStackTrace() {
         return this;
     }
 }
--- a/src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
+++ b/src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
@@ -9,4 +9,9 @@ import lombok.experimental.StandardException;
 public class FlagNotFoundError extends OpenFeatureError {
     private static final long serialVersionUID = 1L;
     @Getter private final ErrorCode errorCode = ErrorCode.FLAG_NOT_FOUND;
+
+    @Override
+    public Throwable fillInStackTrace() {
+        return this;
+    }
 }


beeme1mr (approved):
Do we need the stack trace on any OpenFeatureError?

src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
```diff
@@ -9,4 +9,9 @@
 public class FlagNotFoundError extends OpenFeatureError {
     private static final long serialVersionUID = 1L;
     @Getter private final ErrorCode errorCode = ErrorCode.FLAG_NOT_FOUND;
+
+    @Override
+    public Throwable fillInStackTrace() {
```
toddbaert, line 14:
```suggestion
    public synchronized Throwable fillInStackTrace() {
```

Sonarcloud is recommending this: https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=BUG&pullRequest=739&id=open-feature_java-sdk&open=AYyHw5peA4eBhJGjCSRs

toddbaert (approved):
Agree with this. Just see [here](https://github.com/open-feature/java-sdk/pull/739/files#r1432917044).

src/main/java/dev/openfeature/sdk/exceptions/FlagNotFoundError.java
thisthat: oh true, the method in Throwable is sync 🤦

toddbaert:
> Do we need the stack trace on any OpenFeatureError?

We have cases where consumers do things like:

```java
catch (IOException e) {
  throw new GeneralError("Something bad happend", e);
}
```

This is especially important for things like `GeneralError` where something unexpected has happened due to a logic error or system failure. I don't know how `fillInStackTrace` impacts this. I presume the "Caused by" normally seen in the stack-trace would be lost, and therefor I think in logs, etc, the underlying "cause" would be lost as well.

@thisthat

thisthat:
@toddbaert the suggested `fillInStackTrace()` implementation prints only `dev.openfeature.sdk.exceptions.FlagNotFoundError` instead of the full list of methods calls from the main to the throw of the exception.
For end users, the stack trace is not relevant because they (hopefully) would never see any exception.
So I think we should look into the code of existing providers and how they use the exceptions.

toddbaert:
> @toddbaert the suggested `fillInStackTrace()` implementation prints only `dev.openfeature.sdk.exceptions.FlagNotFoundError` instead of the full list of methods calls from the main to the throw of the exception. For end users, the stack trace is not relevant because they (hopefully) would never see any exception. So I think we should look into the code of existing providers and how they use the exceptions.

@thisthat  Yes, I understand that. I was making a point relative to @beeme1mr 's question:

> Do we need the stack trace on any OpenFeatureError?

My concern is that if we do this for other exceptions which are more complicated than `FlagNotFound`, we lose the causal information.

PR #797, opened 2024-02-09
chore: various improvements as suggested by sonar

Kavindu-Dodan:
## This PR

Contains improvements to code smells and issues suggested by sonarqube.


toddbaert (approved):

src/main/java/dev/openfeature/sdk/Value.java
```diff
@@ -20,7 +20,7 @@
 @ToString
 @EqualsAndHashCode
 @SuppressWarnings({"PMD.BeanMembersShouldSerialize", "checkstyle:MissingJavadocType", "checkstyle:NoFinalizer"})
-public class Value implements Cloneable {
+public class Value {
```
liran2000, line 23: this change is ok, isn't it breaking change ?

liran2000 (approved):
approve, notice review comment

src/main/java/dev/openfeature/sdk/Value.java
toddbaert: @Kavindu-Dodan I think @liran2000 is right, I didn't notice it but this does seem to be breaking.

toddbaert: I think we need to revert the changes in this file.

Kavindu-Dodan: Reverted this so we are not breaking :)

toddbaert (approved):

PR #805, opened 2024-02-14
fix: improve targetingKey handling in the context

### LINKED ISSUE #801: [Feature] Consider exposing TargetingKey with all value exporters
Consider `ImmutableStructure.asMap()` [^1] or `ImmutableStructure. asObjectMap()` [^2] methods. These expose the attributes but they do not include the targeting key [^3]. 

This could lead to unexpected behaviors where evaluations miss targeting key if provider authors overlook the implementation details (see [^4]). In contrast, go-sdk perform a context flattening to include targeting key [^5] and js-sdk context is a record hence this issue does not exist in them.

**AC**

If agreed, we should include the targeting key in bulk exporters (ex:- ctx.asObject) as well as improve overall context handling by documentation improvements 

[^1]: https://github.com/open-feature/java-sdk/blob/v1.7.3/src/main/java/dev/openfeature/sdk/MutableStructure.java#L90
[^2]: https://github.com/open-feature/java-sdk/blob/v1.7.3/src/main/java/dev/openfeature/sdk/AbstractStructure.java#L25 
[^3]: https://github.com/open-feature/java-sdk/blob/v1.7.3/src/main/java/dev/openfeature/sdk/ImmutableContext.java#L21 
[^4]: https://github.com/open-feature/java-sdk-contrib/pull/676
[^5]: https://github.com/open-feature/go-sdk/blob/main/openfeature/client.go#L687-L695

Kavindu-Dodan:
## This PR

Fixes #801. 

**What's changed?** 

I have changed how we store the targetingKey internally (in both mutable and immutable contexts). Instead of storing in a field, we now store targetingKey as a property of the delegated Structure implemementation. 

**Why** 

Both mutable and immutable contexts use a Structure implementation instance and delegate operations onto this instance. For example, `asObjectMap` and `asMap()` are exposed through this delegation. With this change, we keep the same delegation and expose targetingKey from the above-mentioned attribute exporters. 

**Impact**

There are no behaviour change other than for avoiding storing of empty string (`""`) for targetingKey.

### DIFF
--- a/src/test/java/dev/openfeature/sdk/EvalContextTest.java
+++ b/src/test/java/dev/openfeature/sdk/EvalContextTest.java
@@ -1,17 +1,16 @@
 package dev.openfeature.sdk;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import io.cucumber.java.hu.Ha;
-import org.junit.jupiter.api.Test;
+import static dev.openfeature.sdk.EvaluationContext.TARGETING_KEY;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class EvalContextTest {
     @Specification(number="3.1.1",
@@ -184,7 +183,7 @@ public class EvalContextTest {
 
         ctx2.setTargetingKey("  ");
         ctxMerged = ctx1.merge(ctx2);
-        assertEquals(key1, ctxMerged.getTargetingKey());
+        assertEquals(key2, ctxMerged.getTargetingKey());
     }
 
     @Test void asObjectMap() {
@@ -214,6 +213,7 @@ public class EvalContextTest {
 
 
         Map<String, Object> want = new HashMap<>();
+        want.put(TARGETING_KEY, key1);
         want.put("stringItem", "stringValue");
         want.put("boolItem", false);
         want.put("integerItem", 1);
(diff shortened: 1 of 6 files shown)


src/test/java/dev/openfeature/sdk/EvalContextTest.java
```diff
@@ -184,7 +183,7 @@ public class EvalContextTest {
 
         ctx2.setTargetingKey("  ");
         ctxMerged = ctx1.merge(ctx2);
-        assertEquals(key1, ctxMerged.getTargetingKey());
```
Kavindu-Dodan, line 187: This is the only major change - we do not accept empty value or spaces for targeting key. Eaerlier this was handled in the merge starge where we filtered for null or empty values.

toddbaert:
@Kavindu-Dodan do you think we should do the same thing in [dotnet](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Model/EvaluationContext.cs)?

kinyoklion:
> @Kavindu-Dodan This seems like a good idea to me. We recently had a bug in flagd-java due to this. I think the risk is low but I'm interested in what others think.
> 
> Do you think we should do the same thing in [dotnet](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/Model/EvaluationContext.cs)? cc @kinyoklion @austindrenski

I like this idea. Prior to it being formalized into the API of the context itself I had been targeting the "targetingKey" based on a string constant. 

A snippet from the documentation for the LaunchDarkly dotnet provider.
```
var evaluationContext = EvaluationContext.Builder()
  .Set("kind", "organization")
  .Set("targetingKey", "my-org-key") // Could also use "key" instead of "targetingKey".
  .Build();
```

By storing it in an attribute it would theoretically maintain compatibility with those assumptions.

lukas-reining (approved):
Makes sense and the change looks good to me!

kinyoklion (approved):

toddbaert (approved):
@Kavindu-Dodan would you mind creating a similar .NET issue if you agree?

Kavindu-Dodan:
> @Kavindu-Dodan would you mind creating a similar .NET issue if you agree?

Created a .Net issue - https://github.com/open-feature/dotnet-sdk/issues/235 

I will merge this as we have an agreement on the change :)

PR #934, opened 2024-05-13
feat: implement domain scoping

### LINKED ISSUE #843: [FEATURE] Implement domain scoping
## Overview

A `domain` is an identifier that logically binds clients with providers, allowing multiple providers to be used simultaneously within a single application.

Domains were added to the spec in the following PR: https://github.com/open-feature/spec/pull/229

## Tasks

```[tasklist]
### Tasks
- [ ] [Requirement 1.1.3](https://openfeature.dev/specification/sections/flag-evaluation#requirement-113): The API MUST provide a function to bind a given provider to one or more clients using a domain. If the domain already has a bound provider, it is overwritten with the new mapping.
- [ ] [Requirement 1.1.6](https://openfeature.dev/specification/sections/flag-evaluation#requirement-116): The API MUST provide a function for creating a client which accepts `domain`
- [ ] [Requirement 1.2.2](https://openfeature.dev/specification/sections/flag-evaluation#requirement-122): The client interface MUST define a metadata member or accessor, containing an immutable domain field or accessor of type string, which corresponds to the domain value supplied during client creation.
- [ ] [Requirement 3.2.2.3](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3223): The API MUST have a method for setting evaluation context for a domain.
- [ ] [Requirement 3.2.2.4](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3224): The API MUST have a mechanism to manage evaluation context for an associated domain.
- [ ] Update the `named provider` section of the readme to `domain`. See the JS SDK readme for an example: https://github.com/open-feature/js-sdk/tree/main/packages/server#domains
``` 

## Reference implementations

Support was added in the JS SDK [here](https://github.com/open-feature/js-sdk/pull/805) and the Python SDK [here](https://github.com/open-feature/python-sdk/pull/271).

> [!NOTE]
> Some SDKs support `named clients` which predate `domains`. The `domain` term is intended to clarify intended behavior.

jarebudev:
## This PR
- implements most of #843 (see notes as have currently left out the 2 experimental requirements)

### Related Issues

Resolves #843 

### Notes

As this change was getting a bit large I wanted to make sure that the main requirements of the issue were implemented correctly first of all so I've not yet done the two requirements that are listed as experimental.

I can either add them into this PR or split them out into a separate issue, either way is fine with me, but wanted to check that this change is on the right path first of all.

### Follow-up Tasks

I've not implemented these two requirements

- [Requirement 3.2.2.3](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3223): The API MUST have a method for setting evaluation context for a domain.
- [Requirement 3.2.2.4](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3224): The API MUST have a mechanism to manage evaluation context for an associated domain.


toddbaert:
> I've not implemented these two requirements
> 
> [Requirement 3.2.2.3](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3223): The API MUST have a method for setting evaluation context for a domain.
> [Requirement 3.2.2.4](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3224): The API MUST have a mechanism to manage evaluation context for an associated domain.

These requirements only apply to the static context SDKs (read: client-side SDKs) which Java is not, so you can skip them!

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -354,8 +357,18 @@ public FlagEvaluationDetails<Value> getObjectDetails(String key, Value defaultVa
...
+            @Override
+            public String getName() {
+                return name;
+            }
+
+            @Override
+            public String getDomain() {
+                return domain;
+            }
+        };
```
toddbaert, line 371: See: https://github.com/open-feature/java-sdk/pull/934/files#r1605189937

src/main/java/dev/openfeature/sdk/EventDetails.java
```diff
@@ -19,9 +21,9 @@ static EventDetails fromProviderEventDetails(ProviderEventDetails providerEventD
     static EventDetails fromProviderEventDetails(
             ProviderEventDetails providerEventDetails,
             String providerName,
-            String clientName) {
+            String domain) {
         return builder()
-                .clientName(clientName)
+                .domain(domain)
```
toddbaert, line 26: Let's also set the clientName to the domain, so nothing breaks.

src/main/java/dev/openfeature/sdk/EventDetails.java
```diff
@@ -9,7 +9,9 @@
 @Data
 @SuperBuilder(toBuilder = true)
 public class EventDetails extends ProviderEventDetails {
+    @Deprecated
```
toddbaert, line 12: :+1:

src/main/java/dev/openfeature/sdk/ClientMetadata.java
```diff
@@ -0,0 +1,10 @@
+package dev.openfeature.sdk;
+
+/**
+ * Holds identifying information about a Client.
+ */
+public interface ClientMetadata {
```
Kavindu-Dodan, line 6: We should avoid this and use `Metadata` interface with new non-breaking default method :)

src/main/java/dev/openfeature/sdk/EventSupport.java
```diff
@@ -122,11 +121,11 @@ public void removeGlobalHandler(ProviderEvent event, Consumer<EventDetails> hand
...
 
     /**
-     * Get all client names for which we have event handlers registered.
+     * Get all domain names for which we have event handlers registered.
      * 
-     * @return set of client names
+     * @return set of domain names
      */
-    public Set<String> getAllClientNames() {
+    public Set<String> getAllDomainNames() {
```
Kavindu-Dodan, line 128: I think we can avoid this change, and even make `getAllClientNames` package private 🤔

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -354,8 +357,18 @@ public FlagEvaluationDetails<Value> getObjectDetails(String key, Value defaultVa
     }
 
     @Override
-    public Metadata getMetadata() {
-        return () -> name;
+    public ClientMetadata getMetadata() {
```
Kavindu-Dodan, line 360: Let's avoid this breaking change and utilize `Metadata`

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -135,10 +135,13 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
         assertEquals(h2, api.getHooks().get(1));
     }
 
-    @Specification(number="1.1.6", text="The API MUST provide a function for creating a client which accepts the following options:  - name (optional): A logical string identifier for the client.")
-    @Test void namedClient() {
-        assertThatCode(() -> api.getClient("Sir Calls-a-lot")).doesNotThrowAnyException();
-        // TODO: Doesn't say that you can *get* the client name.. which seems useful?
+    @Specification(number="1.1.6", text="The API MUST provide a function for creating a client which accepts the following options:  - domain (optional): A logical string identifier for binding clients to provider.")
+    @Test void domainName() {
+        assertNull(api.getClient().getMetadata().getDomain());
```
Kavindu-Dodan, line 140: +1 for addressing the todo

Kavindu-Dodan:
> Thanks @jarebudev ! This looks great, there's only a few minor changes that need to be made non-breaking, then I can approve.
> 
> Thanks again for your time and effort!

Same opinion, we can make the change non-breaking. I proposed [this](https://github.com/open-feature/java-sdk/pull/934/files#r1605195997)

jarebudev:
> > Thanks @jarebudev ! This looks great, there's only a few minor changes that need to be made non-breaking, then I can approve.
> > Thanks again for your time and effort!
> 
> Same opinion, we can make the change non-breaking. I proposed [this](https://github.com/open-feature/java-sdk/pull/934/files#r1605195997)

Thanks both for your feedback - I'll apply your suggestions :)

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
jarebudev: no problem, was a good opportunity to resolve it!

src/main/java/dev/openfeature/sdk/Metadata.java
```diff
@@ -5,4 +5,8 @@
  */
 public interface Metadata {
     String getName();
```
Kavindu-Dodan, line 7: I think we should deprecate this now (@toddbaert  appreciate your opinion)

```suggestion
     @Deprecated
    default String getName(){
        return "";
    }
```

Kavindu-Dodan:
@jarebudev thank you for the latest change. I have [this doubt ](https://github.com/open-feature/java-sdk/pull/934#discussion_r1613678381) and I will approve once I clarify this with @toddbaert  :)

src/main/java/dev/openfeature/sdk/Metadata.java
toddbaert: Agreed

toddbaert (approved):
Approved. Agree with your deprecation suggestion @Kavindu-Dodan

Kavindu-Dodan:
> Approved. Agree with your deprecation suggestion @Kavindu-Dodan

Unfortunately this was not straightforward. We used same `Metadata` interface for both clients and providers. Given that domains are part of `Client`, I had to introduce a new interface `ClientMetadata`. See this commit https://github.com/open-feature/java-sdk/pull/934/commits/2d2c5cfe708cb0899b4b8b2d39735239a4292325

toddbaert:
> > Approved. Agree with your deprecation suggestion @Kavindu-Dodan
> 
> Unfortunately this was not straightforward. We used same `Metadata` interface for both clients and providers. Given that domains are part of `Client`, I had to introduce a new interface `ClientMetadata`. See this commit [2d2c5cf](https://github.com/open-feature/java-sdk/commit/2d2c5cfe708cb0899b4b8b2d39735239a4292325)

Isn't this change breaking since we changed a return type?

Kavindu-Dodan:
> > > > Approved. Agree with your deprecation suggestion @Kavindu-Dodan
> > > 
> > > 
> > > Unfortunately this was not straightforward. We used same `Metadata` interface for both clients and providers. Given that domains are part of `Client`, I had to introduce a new interface `ClientMetadata`. See this commit [2d2c5cf](https://github.com/open-feature/java-sdk/commit/2d2c5cfe708cb0899b4b8b2d39735239a4292325)
> > 
> > 
> > Isn't this change breaking since we changed a return type?
> 
> Yes, could consider as a breaking change as `Client` now use `ClientMetadata`. But if someone use lambda, then it's a simple functional interface with a string return, so shouldn't notice it.

I provided more update here with https://github.com/open-feature/java-sdk/pull/934/commits/12f7f4dad04c5c95e6cab22d910eedf947b089d9

`ClientMetadata` expose a deprecated `getName()` so that wherever users have `client.getMetadata().getName()` is no longer broken. Internally, the returning string matches the domain.

Kavindu-Dodan (approved):

PR #951, opened 2024-05-23
fix: run error hook when provider returns error code

### LINKED ISSUE #928: [BUG] Error hooks only run if the provider throws
See the issue reported for JS SDK [^1]

Similarly, we need to check details for error and execute error hook in Java SDK [^2]

[^1]: - https://github.com/open-feature/js-sdk/issues/925
[^2]: - https://github.com/open-feature/java-sdk/blob/v1.8.0/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java#L126-L141

jarebudev:
## This PR

- ensures that the error hook is ran when the provider returns an error code

### Related Issues

Fixes #928 

### Notes

saw this bug and seemed like a straightforward fix, wasn't sure if anyone has already picked it up so feel free to reject if it's already been taken on

### DIFF
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -127,9 +127,7 @@ public class OpenFeatureClient implements Client {
 
             details = FlagEvaluationDetails.from(providerEval, key);
             if (details.getErrorCode() != null) {
-                Exception e = ExceptionUtils.instantiateErrorByErrorCode(details.getErrorCode(),
-                        details.getErrorMessage());
-                hookSupport.errorHooks(type, hookCtx, e, mergedHooks, hints);
+                throw ExceptionUtils.instantiateErrorByErrorCode(details.getErrorCode(), details.getErrorMessage());
             } else {
                 hookSupport.afterHooks(type, hookCtx, details, mergedHooks, hints);
             }
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -11,6 +11,7 @@ import dev.openfeature.sdk.exceptions.GeneralError;
 import dev.openfeature.sdk.exceptions.OpenFeatureError;
 import dev.openfeature.sdk.internal.AutoCloseableLock;
 import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
+import dev.openfeature.sdk.internal.ErrorUtils;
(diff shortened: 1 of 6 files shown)


src/main/java/dev/openfeature/sdk/internal/ErrorUtils.java
```diff
@@ -0,0 +1,42 @@
+package dev.openfeature.sdk.internal;
```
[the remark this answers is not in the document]
jarebudev: Yeh sounds good to me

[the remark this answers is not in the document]
Kavindu-Dodan: Thanks :)

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -125,7 +126,13 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
                     defaultValue, provider, mergedCtx);
 
             details = FlagEvaluationDetails.from(providerEval, key);
-            hookSupport.afterHooks(type, hookCtx, details, mergedHooks, hints);
+            if (details.getErrorCode() != null) {
+                Exception e = ExceptionUtils.instantiateErrorByErrorCode(details.getErrorCode(),
+                        details.getErrorMessage());
+                hookSupport.errorHooks(type, hookCtx, e, mergedHooks, hints);
```
toddbaert, line 132: The only sort-of weird edge case I can see here is if the error hooks themselves throw (unlikely) then they will possibly run again because of the catch block. Maybe we could have some boolean or some other way to track this?

It's a bit of an edge case, and maybe it could apply to the JS implementation too.

cc @Kavindu-Dodan @beeme1mr

Kavindu-Dodan: Good point, description of the error hook says,

> Run when evaluation encounters an error. This will always run. Errors thrown will be swallowed.

so we should avoid handling error from hook execution 🤔

Kavindu-Dodan: @toddbaert I added https://github.com/open-feature/java-sdk/pull/951/commits/31290014675e460bc42493ec80b9ae692d0cf30c to throw the derived exception so we have the same flow as if it comes from an exception. 

One could say this is an anti-pattern, but I think this is a valid usage

toddbaert: Good solution, yep, a bit of an any pattern... if anything we could prevent the stack trace generation for performance, but that feels like a nice-to-have.

toddbaert (approved):

PR #1015, opened 2024-07-22
feat: add mechanism for open-close-principle to OpenFeatureApi

aepfli:
With this approach we allow people to modify the OpenFeatureApi Singleton.

The user can either provide an environment variable called `OPEN_FEATURE_API_CLASS` or a property file called `openfeature.properties` with which they can control the class, which will be used for instantiation.

first the environmental variable
second the properties file
third the default

(optional we could add system properties)

The provided class needs to extend OpenFeatureApi, and if we're not able to instantiate this class, we will fall to the default

This is just a proposal, of how we could add the open-close principle to the OpenFeatureApi, this approach could also work in other languages like python or c# - not sure of the others.


aepfli:
after a night of sleep, i still think this is a feasible idea, but it will not help me in my endavours.

Issue #1047, opened 2024-08-14
Hooks are not passed merged context


toddbaert:
We are correctly passing the merged context to the flag evaluation as outlined in [3.2.3](https://openfeature.dev/specification/sections/evaluation-context#requirement-323), however, we are not passing that context to the hooks themselves.

Solution:

The before hooks should be passed a merge of `API (global) > Transaction > Client > Invocation` and the resulting returned context from the hooks should then be passed to the evaluation and all subsequent hooks.

Steps: 

- fix as described
- tests which fail before the fix, and pass after

Here's equivalent code in the JS implementation: https://github.com/open-feature/js-sdk/blob/cb4e56ae803ffe15dd4367b73d82315ce5d053a8/packages/server/src/client/internal/open-feature-client.ts#L248-L268

UtkarshSharma2612:
Can I work on this issue?

toddbaert:
> Can I work on this issue?

Hey @UtkarshSharma2612 ! Sure - but it's a significant issue with a relatively high priority to fix. I was going to start on it today, but if you think it's possible that you can complete it by next week, I will assign it to you. If that is too rushed, I can help you find some other issue or you can help with review.

UtkarshSharma2612:
Hey @toddbaert I am relatively new to open source contribution and I am not sure I will be able to complete by next week it will be a bit rushed and not justified for a high priority issue. Can you help me find another issue, I am willing to help with review and other issues for sure 😄

toddbaert:
@UtkarshSharma2612 [here](https://github.com/open-feature/java-sdk-contrib/issues/901) is another pretty clean-cut issue that you can work on. It's also Java. Take a look!

PR #1049, opened 2024-08-15
fix: updated context not passed to all hooks

toddbaert:
Fixes an issue where the fully merged and updated context was used in the evaluation, but not in all hooks.

Fixes: https://github.com/open-feature/java-sdk/issues/1047

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -73,7 +73,7 @@ public final class ImmutableContext implements EvaluationContext {
      * Merges this EvaluationContext object with the passed EvaluationContext, overriding in case of conflict.
      *
      * @param overridingContext overriding context
-     * @return resulting merged context
+     * @return new, resulting merged context
      */
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -116,7 +116,7 @@ public class OpenFeatureClient implements Client {
                     openfeatureApi.getHooks());
 
             EvaluationContext mergedCtx = hookSupport.beforeHooks(type, HookContext.from(key, type, this.getMetadata(),
-            provider.getMetadata(), mergeEvaluationContext(ctx), defaultValue), mergedHooks, hints);
+                    provider.getMetadata(), mergeEvaluationContext(ctx), defaultValue), mergedHooks, hints);
 
             afterHookContext = HookContext.from(key, type, this.getMetadata(),
                     provider.getMetadata(), mergedCtx, defaultValue);
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -105,7 +105,7 @@ public class OpenFeatureClient implements Client {
 
         FlagEvaluationDetails<T> details = null;
         List<Hook> mergedHooks = null;
-        HookContext<T> hookCtx = null;
+        HookContext<T> afterHookContext = null;
         FeatureProvider provider;
 
         try {
@@ -115,12 +115,11 @@ public class OpenFeatureClient implements Client {
             mergedHooks = ObjectUtils.merge(provider.getProviderHooks(), flagOptions.getHooks(), clientHooks,
                     openfeatureApi.getHooks());
 
-            hookCtx = HookContext.from(key, type, this.getMetadata(),
-                    provider.getMetadata(), ctx, defaultValue);
+            EvaluationContext mergedCtx = hookSupport.beforeHooks(type, HookContext.from(key, type, this.getMetadata(),
+            provider.getMetadata(), mergeEvaluationContext(ctx), defaultValue), mergedHooks, hints);
 
-            EvaluationContext ctxFromHook = hookSupport.beforeHooks(type, hookCtx, mergedHooks, hints);
-
-            EvaluationContext mergedCtx = mergeEvaluationContext(ctxFromHook, ctx);
+            afterHookContext = HookContext.from(key, type, this.getMetadata(),
+                    provider.getMetadata(), mergedCtx, defaultValue);
 
             ProviderEvaluation<T> providerEval = (ProviderEvaluation<T>) createProviderEvaluation(type, key,
                     defaultValue, provider, mergedCtx);
@@ -129,7 +128,7 @@ public class OpenFeatureClient implements Client {
             if (details.getErrorCode() != null) {
                 throw ExceptionUtils.instantiateErrorByErrorCode(details.getErrorCode(), details.getErrorMessage());
             } else {
-                hookSupport.afterHooks(type, hookCtx, details, mergedHooks, hints);
+                hookSupport.afterHooks(type, afterHookContext, details, mergedHooks, hints);
             }
         } catch (Exception e) {
             log.error("Unable to correctly evaluate flag with key '{}'", key, e);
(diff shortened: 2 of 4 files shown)


src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -15,10 +16,19 @@ public class HookContext<T> {
     @NonNull String flagKey;
     @NonNull FlagValueType type;
     @NonNull T defaultValue;
-    @NonNull EvaluationContext ctx;
+    @NonNull @NonFinal EvaluationContext ctx;
```
toddbaert, line 19: Needed to override lombok's automatic `final` for `@Value` classes.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -78,7 +78,7 @@ public String getTargetingKey() {
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
         if (overridingContext == null) {
-            return new ImmutableContext(this.asMap());
+            return this;
```
toddbaert, line 81: No point in creating a new object here.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -116,11 +116,10 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
 
             hookCtx = HookContext.from(key, type, this.getMetadata(),
-                    provider.getMetadata(), ctx, defaultValue);
+                    provider.getMetadata(), mergeEvaluationContext(ctx), defaultValue);
 
-            EvaluationContext ctxFromHook = hookSupport.beforeHooks(type, hookCtx, mergedHooks, hints);
-
-            EvaluationContext mergedCtx = mergeEvaluationContext(ctxFromHook, ctx);
+            EvaluationContext mergedCtx = hookSupport.beforeHooks(type, hookCtx, mergedHooks, hints);
+            hookCtx.setEvaluationCtx(mergedCtx);
```
toddbaert, line 122: Merge the context first, _then_ run the before hooks with that, _then_ update the context passed to the subsequent hooks.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -377,21 +391,54 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
         invocationAttributes.put("invocation", new Value("4"));
         EvaluationContext invocationCtx = new ImmutableContext(invocationAttributes);
 
-        c.getBooleanValue("key", false, invocationCtx);
-
-        // assert the connect overrides
+        c.getBooleanValue("key", false, invocationCtx, FlagEvaluationOptions.builder().hook(hook).build());
```
toddbaert, line 394: The failed before the fix.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -377,21 +391,54 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
...
                     arg.getValue("common2").asString().equals("3") &&
                     arg.getValue("common3").asString().equals("4") &&
                     arg.getValue("common4").asString().equals("3") &&
                     arg.getValue("common5").asString().equals("4") &&
-                    arg.getValue("common6").asString().equals("4");
+                    arg.getValue("common6").asString().equals("4") &&
+                    arg.getValue("common7").asString().equals("5");
         }));
+
+        // assert the connect overrides in after hook
```
toddbaert, line 427: The failed before the fix.

toddbaert:
@UtkarshSharma2612 feel free to add yourself as a reviewer if you want :pray:

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -377,21 +391,55 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
         invocationAttributes.put("invocation", new Value("4"));
         EvaluationContext invocationCtx = new ImmutableContext(invocationAttributes);
 
-        c.getBooleanValue("key", false, invocationCtx);
-
-        // assert the connect overrides
+        c.getBooleanValue("key", false, invocationCtx, FlagEvaluationOptions.builder().hook(hook).build());
+
+        // assert the correct overrides in before hook
```
toddbaert, line 396: This failed before the fix.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -377,21 +391,55 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
...
                     arg.getValue("common2").asString().equals("3") &&
                     arg.getValue("common3").asString().equals("4") &&
                     arg.getValue("common4").asString().equals("3") &&
                     arg.getValue("common5").asString().equals("4") &&
-                    arg.getValue("common6").asString().equals("4");
+                    arg.getValue("common6").asString().equals("4") &&
+                    arg.getValue("common7").asString().equals("5");
         }));
+
+        // assert the correct overrides in after hook
```
toddbaert, line 427: This failed before the fix.

src/test/java/dev/openfeature/sdk/HookSpecTest.java
```diff
@@ -503,7 +504,7 @@ public void finallyAfter(HookContext<Boolean> ctx, Map<String, Object> hints) {
         order.verify(hook2).before(captor.capture(), any());
 
         HookContext hc = captor.getValue();
-        assertEquals(hc.getCtx(), ctx);
+        assertEquals(hc.getCtx().getTargetingKey(), targetingKey);
```
toddbaert, line 507: The instance of the evaluation context now changes here, but that's not something we guarantee.

src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -4,6 +4,7 @@
 import lombok.NonNull;
 import lombok.Value;
 import lombok.With;
+import lombok.experimental.NonFinal;
 
 /**
  * A data class to hold immutable context that {@link Hook} instances use.
```
kinyoklion, line 10: Isn't immutable anymore. Which does have some implications. I've not looked at the implementation yet, but a hook implementation could reasonable expect to hold a reference to this context and now it could mutate.

Given this class my initial expectation would be for this to be cloned for each hook invocation with an updated context.

 `hookContext.withNewEvaluationConetxt(EvaluationContext context)` which would return a new instance with a new immutable field.

Like dotnet: https://github.com/open-feature/dotnet-sdk/blob/bb4f3526c2c2c2ca48ae61e883d6962847ebc5a6/src/OpenFeature/Model/HookContext.cs#L69C33-L69C57

toddbaert: You're right.

The challenge is at least in a couple hooks, we use the HookContext itself as a WeakMap key. If the HookContext changes across the evaluation, this functionality doesn't work. WeakMaps in hooks are especially useful for maintaining state across stages (before/after).

toddbaert: We don't explicitly make any guarantees around either of these in the spec.... but the weakmap use-case in hooks is so useful it would be good to have some consistent reference across stages.

kinyoklion: So, we should update these comments and make it explicit that the content can really only be trusted for the duration of your invocation. 

Though I do think the weak map situation represents a specification weakness that should be addressed with a specific way to have invocation data. As is it depends on implementation details, and conversely places unexpected limitations on the implementation.

toddbaert: Agree that we should update comments at a minimum. I feel a bit stuck, because I don't like this mutability, but on the other hand see a lot of value in the consistent reference.

Do you think it would be silly to have some kind of explicit reference handle? We could instantiate some dedicated object for this in the HookContext and never change it if we wanted, and even make it an optional point in the spec.

kinyoklion: I think an explicit handle of some kind is a lot better than pulling double duty with another field. It makes it something explicit instead of an implementation detail and makes it obvious to maintainers. The idea of a weak map itself is somewhat technology specific though, so I am not sure it is great for the spec. I am not sure how I would do something equivalent in functional/immutable languages. 

I think, ideally, we would add a specific way to store and retrieve data in the hook invocation. 

I tried it in the LaunchDarkly implementation hooks: https://github.com/launchdarkly/open-sdk-specs/tree/main/specs/HOOK-hooks

Example usage in C#: https://github.com/launchdarkly/dotnet-core/blob/10045f0f97c0f923d23fa6a7c3af0bd701d2ba42/pkgs/telemetry/src/TracingHook.cs#L155

https://github.com/launchdarkly/dotnet-core/blob/10045f0f97c0f923d23fa6a7c3af0bd701d2ba42/pkgs/telemetry/src/TracingHook.cs#L168

It does have a bit of overhead though as you have to at least create 1 object per hook per evaluation. Though it isn't much overhead compared to the other transformations that can happen, like the context ones.

toddbaert: I really like the `SeriesData` concept.

I don't know of any Java Hooks currently using WeakMaps (just JS ones) so I'm inclined to go with your initial suggestion and remove the mutability I introduced, and also implement some kind of storage in another PR. If it's sensible we can think about bringing it into the spec.

:pray:

aepfli (approved):

src/main/java/dev/openfeature/sdk/HookContext.java
toddbaert: I've removed this mutability, I create a new hook-context now from the result of the "before".

PR #1057, opened 2024-08-21
fix: Use ConcurrentHashMap for InMemoryProvider

### LINKED ISSUE #1054: Thread safe in-memory provider
Update the [in-memory provider](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java) to be thread safe.

- Use a `ConcurrentHashMap` or other mechanisms to ensure the internal flag map is accessible/mutable from multiple threads safely.
- Ensure other mutable state is synchronized, if any can be identified.
- No specific additional testing is required, but existing tests should pass.

ryanprayogo:
## Use ConcurrentHashMap for InMemoryProvider

- Use `ConcurrentHashMap` for `flags` in `InMemoryProvider` to make it thread-safe

### Related Issues

Fixes #1054

### DIFF
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -67,7 +67,6 @@ public class InMemoryProvider extends EventProvider {
      */
     public void updateFlags(Map<String, Flag<?>> newFlags) {
         Set<String> flagsChanged = new HashSet<>(newFlags.keySet());
-        flagsChanged.removeAll(this.flags.keySet());
         this.flags.putAll(newFlags);
 
         ProviderEventDetails details = ProviderEventDetails.builder()
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -2,10 +2,10 @@ package dev.openfeature.sdk.providers.memory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import dev.openfeature.sdk.EvaluationContext;
 import dev.openfeature.sdk.EventProvider;
@@ -44,7 +44,7 @@ public class InMemoryProvider extends EventProvider {
     }
 
     public InMemoryProvider(Map<String, Flag<?>> flags) {
-        this.flags = new HashMap<>(flags);
+        this.flags = new ConcurrentHashMap<>(flags);
     }
 
     /**
@@ -67,7 +67,7 @@ public class InMemoryProvider extends EventProvider {
         Set<String> flagsChanged = new HashSet<>();
         flagsChanged.addAll(this.flags.keySet());
         flagsChanged.addAll(flags.keySet());
-        this.flags = new HashMap<>(flags);
+        this.flags = new ConcurrentHashMap<>(flags);
         ProviderEventDetails details = ProviderEventDetails.builder()
             .flagsChanged(new ArrayList<>(flagsChanged))
             .message("flags changed")
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -33,7 +33,7 @@ public class InMemoryProvider extends EventProvider {
     @Getter
     private static final String NAME = "InMemoryProvider";
 
-    private Map<String, Flag<?>> flags;
+    private final Map<String, Flag<?>> flags;
 
     @Getter
     private ProviderState state = ProviderState.NOT_READY;
@@ -61,13 +61,13 @@ public class InMemoryProvider extends EventProvider {
 
     /**
      * Updating provider flags configuration, replacing existing flags.
-     * @param flags the flags to use instead of the previous flags.
+     * @param newFlags the flags to use instead of the previous flags.
      */
-    public void updateFlags(Map<String, Flag<?>> flags) {
-        Set<String> flagsChanged = new HashSet<>();
-        flagsChanged.addAll(this.flags.keySet());
-        flagsChanged.addAll(flags.keySet());
-        this.flags = new ConcurrentHashMap<>(flags);
+    public void updateFlags(Map<String, Flag<?>> newFlags) {
+        Set<String> flagsChanged = new HashSet<>(newFlags.keySet());
+        flagsChanged.removeAll(this.flags.keySet());
+        this.flags.putAll(newFlags);
+
         ProviderEventDetails details = ProviderEventDetails.builder()
             .flagsChanged(new ArrayList<>(flagsChanged))
             .message("flags changed")
--- a/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
+++ b/src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
@@ -48,7 +48,7 @@ public class InMemoryProvider extends EventProvider {
     }
 
     /**
-     * Initialize the provider.
+     * Initializes the provider.
      * @param evaluationContext evaluation context
      * @throws Exception on error
      */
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):

AndiHofi (changes requested):
This is still not thread-safe.

Either the `flags` field has to be final and modified in-place.

Or use a normal HashMap and a lock for each access to the map.

Or an AtomicReference and the feature flags are held in an immutable Map (does not need to be immutable, but it prevents from changing the map accidentally after setting it to the AtomicReference.

The simplest option is to make the `flags` field final and update it with a normal `put(key, value)`

Also, in `updateFlags(Map flags)` the flagsChanged set should only contain the keys of the changed flags, not all flags the provider. That bug existed earlier already.

The test is likely unstable, independent if the InMemoryProvider is threadsafe or not. There is no guarantee that any of the scheduled tasks even started execution when the assertion for that flag runs.

toddbaert:
> This is still not thread-safe.
> 
> Either the `flags` field has to be final and modified in-place.
> 
> Or use a normal HashMap and a lock for each access to the map.
> 
> Or an AtomicReference and the feature flags are held in an immutable Map (does not need to be immutable, but it prevents from changing the map accidentally after setting it to the AtomicReference.
> 
> The simplest option is to make the `flags` field final and update it with a normal `put(key, value)`
> 
> Also, in `updateFlags(Map flags)` the flagsChanged set should only contain the keys of the changed flags, not all flags the provider. That bug existed earlier already.

I believe this is all correct. The use of the concurrent map is _safer_ but there are still thread safety issues with this new implementaiton

> The test is likely unstable, independent if the InMemoryProvider is threadsafe or not. There is no guarantee that any of the scheduled tasks even started execution when the assertion for that flag runs.

Yes, I generally find this sort of thing difficult to test, which is why I don't think testing this is a hard requirement. As long as there's a consensus on the thread safety I think we're OK to merge.

toddbaert (changes requested):
An improvement, certainly! but still a bit to go: https://github.com/open-feature/java-sdk/pull/1057#issuecomment-2302275383

 :pray: :

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -82,7 +82,7 @@ public void updateFlags(Map<String, Flag<?>> flags) {
     public void updateFlag(String flagKey, Flag<?> flag) {
         this.flags.put(flagKey, flag);
         ProviderEventDetails details = ProviderEventDetails.builder()
-            .flagsChanged(Arrays.asList(flagKey))
+            .flagsChanged(Collections.singletonList(flagKey))
```
ryanprayogo, line 85: This change is to make it clear to the next developer that it is the intention to only have a single value passed into this parameter.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -61,13 +61,13 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
 
     /**
      * Updating provider flags configuration, replacing existing flags.
-     * @param flags the flags to use instead of the previous flags.
+     * @param newFlags the flags to use instead of the previous flags.
```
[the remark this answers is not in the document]
toddbaert: Ya I think this is clearer. While you're at it I would perhaps recommend improving the awkwardness of the description:

"~Updating~ Updates the providers ~flags~ flag configuration, replacing existing flags." or something like that.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
toddbaert: I like it.

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -105,4 +114,42 @@ void shouldThrowIfNotInitialized() {
...
+                return null;
+            });
+        }
+
+        executorService.invokeAll(updates);
+
+        for (int i = 0; i < 10000; i++) {
+            assertTrue(client.getBooleanValue("multithreadedFlag" + i, false));
+        }
+    }
```
toddbaert, line 154: I'm a bit dubious of the value of this test. Did you actually see it fail at least sometimes without your changes?

toddbaert (approved):
I think this covers it! Thank you!

I'll wait a bit for a follow-up from @AndiHofi since their observations were pretty astute before.

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
ryanprayogo: I did write this test before making any changes and did see it had failed sometimes.

After my change, I have not observed any failures in this test.

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -105,4 +114,42 @@ void shouldThrowIfNotInitialized() {
...
+    void emitChangedFlagsOnlyIfThereAreChangedFlags() {
+        Consumer<EventDetails> handler = mock(Consumer.class);
+        Map<String, Flag<?>> flags = buildFlags();
+
+        OpenFeatureAPI.getInstance().onProviderConfigurationChanged(handler);
+        OpenFeatureAPI.getInstance().setProviderAndWait(provider);
+
+        provider.updateFlags(flags);
+
+        verify(handler, times(1))
```
toddbaert, line 129: Oh, I think this test is a bit flaky. You probably will need to use Awaitility to make this assertion within some time frame, since it happens asynchronously on another thread. If you search for `await` in this repo you'll see a bunch of similar examples.

toddbaert: I did this [here](https://github.com/open-feature/java-sdk/pull/1057/commits/e139de4f5653f1db5dbea56eaa5f333ec9c63d31) since it was a quick change. I hope you don't mind @ryanprayogo

ryanprayogo: @toddbaert I do not mind, thank you.

Would it be worth adding an `.atMost(...)` so the test doesn't hang forever in the case that the assertion doesn't pass?

toddbaert: There's a default time limit (I think 5s?). After that point the test fails with a timeout.

ryanprayogo: Ah yes, you're right 👌

AndiHofi (changes requested):
See comments in code

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -60,14 +60,16 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
...
+     * @param newFlags the new flag configurations
      */
-    public void updateFlags(Map<String, Flag<?>> flags) {
-        Set<String> flagsChanged = new HashSet<>();
-        flagsChanged.addAll(this.flags.keySet());
-        flagsChanged.addAll(flags.keySet());
-        this.flags = new HashMap<>(flags);
+    public void updateFlags(Map<String, Flag<?>> newFlags) {
+        Set<String> flagsChanged = new HashSet<>(newFlags.keySet());
+        flagsChanged.removeAll(this.flags.keySet());
```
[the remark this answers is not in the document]
toddbaert: Oh, nice catch.

Ya I think it's fine just to list all flags here. This is just a testing util. 

@ryanprayogo you can mention that all flags will be put into the change notification. In a way that's technically correct anyway since the set is immutable. This is a testing util primarily, and anyway providers get to decide what constitutes "changed" for their purposes.

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -105,4 +115,42 @@ void shouldThrowIfNotInitialized() {
...
+        OpenFeatureAPI.getInstance().setProviderAndWait(provider);
+
+        provider.updateFlags(flags);
+
+        await().untilAsserted(() -> verify(handler, times(1))
+                .accept(argThat(details -> details.getFlagsChanged().isEmpty())));
+    }
+
+    @Test
+    void multithreadedTest() throws InterruptedException {
```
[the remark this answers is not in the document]
toddbaert: I'm not sure the test is unstable, but I do question it's value.

toddbaert:
@ryanprayogo I've made the small changes recommended by @AndiHofi . I'm sure you could have done them, but I want to get this into the next release I was hoping to do today! Thanks for fixing this (and also for your other Java 21 fix!)

PR #1060, opened 2024-08-21
fix: Pin byte-buddy(-agent) version to 1.14.19 to workaround Mockito issue

### LINKED ISSUE #1059: Some tests are failing under Java 21
Some tests that are using Mockito are failing under Java 21 with this error
```
Cannot call abstract real method on java object!
Calling real methods is only possible when mocking non abstract method.
```
This seem to be caused by an [open issue in Mockito](https://github.com/mockito/mockito/issues/3121).

There is a workaround available (set the `net.bytebuddy.experimental` system property to `true`), which can be set in the unit tests' `@BeforeAll` annotated method.

Steps to reproduce:
- Use Java 21 (reproducible with Temurin JDK 21.0.4)
- Run `mvn clean test`
- Observe failures

ryanprayogo:
## Pin byte-buddy(-agent) version to 1.14.19 to workaround Mockito issue

Workaround to make Mockito tests to pass under Java 21 due to this [open Mockito bug](https://github.com/mockito/mockito/issues/3121)

### Related Issues

Fixes #1059

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,8 @@
   <dependencyManagement>
     <dependencies>
 
+      <!-- Start mockito workaround -->
+      <!-- https://github.com/mockito/mockito/issues/3121 -->
       <dependency>
         <groupId>net.bytebuddy</groupId>
         <artifactId>byte-buddy</artifactId>
@@ -164,6 +166,7 @@
         <version>1.14.19</version>
         <scope>test</scope>
       </dependency>
+      <!-- End mockito workaround-->
 
       <dependency>
         <groupId>io.cucumber</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -239,6 +239,9 @@
         <configuration>
           <argLine>
             ${surefireArgLine}
+            <!-- workaround for Mockito issue under Java 21
+             https://github.com/mockito/mockito/issues/3121 -->
+            -Dnet.bytebuddy.experimental=true
           </argLine>
           <excludes>
             <!-- tests to exclude -->
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,20 @@
     <dependencies>
 
       <dependency>
+        <groupId>net.bytebuddy</groupId>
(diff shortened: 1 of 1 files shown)


justinabrahms:
The fix looks reasonable given the issue output.

The linked issue suggests that a more recent version of mockito may alleviate the need to set these vars. The PR merged in 2023. They had a release this past May. Could we just upgrade mockito and have that work instead?

pom.xml
```diff
@@ -13,7 +13,7 @@
     <junit.jupiter.version>5.11.0</junit.jupiter.version>
      <!-- exclusion expression for e2e tests -->
     <testExclusions>**/e2e/*.java</testExclusions>
-    <module-name>${groupId}.${artifactId}</module-name>
+    <module-name>${project.groupId}.${project.artifactId}</module-name>
```
ryanprayogo, line 16: Fixing maven warnings
```
[WARNING] The expression ${groupId} is deprecated. Please use ${project.groupId} instead.
[WARNING] The expression ${artifactId} is deprecated. Please use ${project.artifactId} instead.
```

toddbaert:
This solves the issue for me locally! very nice to have!

toddbaert:
> The fix looks reasonable given the issue output.
> 
> The linked issue suggests that a more recent version of mockito may alleviate the need to set these vars. The PR merged in 2023. They had a release this past May. Could we just upgrade mockito and have that work instead?

We could try mockito 5. This is the latest version of 4.

toddbaert:
> > The fix looks reasonable given the issue output.
> > The linked issue suggests that a more recent version of mockito may alleviate the need to set these vars. The PR merged in 2023. They had a release this past May. Could we just upgrade mockito and have that work instead?
> 
> We could try mockito 5. This is the latest version of 4.

@justinabrahms  ahh, we can't. Mockito 5 doesn't support Java 8. For as long as we want to have Java 8 support, we need to stick with Mockito 4. This PR seems like the best of both worlds.

toddbaert (approved):

PR #1084, opened 2024-09-03
feat: add logging hook, rm logging from evaluation

toddbaert:
Implements: https://github.com/open-feature/spec/pull/269

This PR removes logging from the main evaluation code path, with a couple exceptions: 

- unhandled errors in before/finally/error hooks (after hooks may throw in order to do validation)
- unhandled errors in event handlers

(IMO these represent bugs in code, since they should be handled, and not transient errors).

It also adds a `LoggingHook` as per the spec link above, which adds SLF4J structured logs for evaluation objects.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/HookSupport.java
+++ b/src/main/java/dev/openfeature/sdk/HookSupport.java
@@ -17,9 +17,18 @@ import lombok.extern.slf4j.Slf4j;
 @SuppressWarnings({ "unchecked", "rawtypes" })
 class HookSupport {
 
-    public void errorHooks(FlagValueType flagValueType, HookContext hookCtx, Exception e, List<Hook> hooks,
+    public EvaluationContext beforeHooks(FlagValueType flagValueType, HookContext hookCtx, List<Hook> hooks,
             Map<String, Object> hints) {
-        executeHooks(flagValueType, hooks, "error", hook -> hook.error(hookCtx, e, hints));
+        Stream<EvaluationContext> result = callBeforeHooks(flagValueType, hookCtx, hooks, hints);
+        return hookCtx.getCtx().merge(
+                result.reduce(hookCtx.getCtx(), (EvaluationContext accumulated, EvaluationContext current) -> {
+                    return accumulated.merge(current);
+                }));
+    }
+
+    public void afterHooks(FlagValueType flagValueType, HookContext hookContext, FlagEvaluationDetails details,
+            List<Hook> hooks, Map<String, Object> hints) {
+        executeHooksUnchecked(flagValueType, hooks, hook -> hook.after(hookContext, details, hints));
     }
 
     public void afterAllHooks(FlagValueType flagValueType, HookContext hookCtx, List<Hook> hooks,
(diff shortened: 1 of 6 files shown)


src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -44,6 +53,17 @@ private <T> void executeHooks(
         }
     }
 
+    // before, error, and finally hooks shouldn't throw
+    private <T> void executeChecked(Hook<T> hook, Consumer<Hook<T>> hookCode, String hookMethod) {
+        try {
+            hookCode.accept(hook);
+        } catch (Exception exception) {
+            log.error("Unhandled exception when running {} hook {} (only 'after' hooks should throw)", hookMethod,
```
toddbaert, line 61: This ONLY runs for hooks which we expect not to throw (before/finally/error).

src/main/java/dev/openfeature/sdk/hooks/logging/LoggingHook.java
```diff
@@ -0,0 +1,92 @@
...
+    static final String DOMAIN_KEY = "domain";
+    static final String PROVIDER_NAME_KEY = "provider_name";
+    static final String FLAG_KEY_KEY = "flag_key";
+    static final String DEFAULT_VALUE_KEY = "default_value";
+    static final String EVALUATION_CONTEXT_KEY = "evaluation_context";
+    static final String ERROR_CODE_KEY = "error_code";
+    static final String ERROR_MESSAGE_KEY = "error_message";
+    static final String REASON_KEY = "reason";
+    static final String VARIANT_KEY = "variant";
+    static final String VALUE_KEY = "value";
```
beeme1mr, line 33: Please update the spec PR to include the updated names.

beeme1mr (commented):
Please add a section that describes the logging hook to the readme.

src/main/java/dev/openfeature/sdk/hooks/logging/LoggingHook.java
toddbaert: done

toddbaert:
> Please add a section that describes the logging hook to the readme.

Please check out my addition.

src/main/java/dev/openfeature/sdk/hooks/logging/LoggingHook.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * A hook for logging flag evaluations.
+ * Useful for debugging.
+ * Flag evaluation data is logged at debug and error in before/after stages and error stages, respectively.
+ */
+@Slf4j
+@edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED",
+    justification = "we can ignore return values of chainables (builders) here")
+public class LoggingHook implements Hook<Object> {
```
aepfli, line 22: yes it is a logging hook, but should we be a little bit more specific, like this is an slf4j Logger hook - there can be other logging hooks for different kind of logging wrappers etc.

aepfli (approved):

src/main/java/dev/openfeature/sdk/hooks/logging/LoggingHook.java
toddbaert: I don't think I want to put `SLF4j` or anyhting like that in the class name, since slf4j is the framework the SDK uses, it's taken for granted. I will add something about the structured logging to the javadoc though.

beeme1mr (approved):

src/main/java/dev/openfeature/sdk/hooks/logging/LoggingHook.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Construct a new LoggingHook.
+     * @param includeEvaluationContext include a serialized evaluation context in the log message (defaults to false)
+     */
+    public LoggingHook(boolean includeEvaluationContext) {
+        this.includeEvaluationContext = includeEvaluationContext;
+    }
+
+    @Override
+    public Optional<EvaluationContext> before(HookContext<Object> hookContext, Map<String, Object> hints) {
```
Kavindu-Dodan, line 53: nit - implementation ignores the `hints` param. IMO we can iterate and add all hints to the Log output [^1]

[^1]: https://www.slf4j.org/api/org/slf4j/spi/LoggingEventBuilder.html#addKeyValue(java.lang.String,java.lang.Object)

Kavindu-Dodan (approved):
Looks good. Please check this remark though [1]

[1] - https://github.com/open-feature/java-sdk/pull/1084/files#r1744351346

PR #1095, opened 2024-09-07
feat: error resolution flow control without exceptions

aepfli:
So far, we used exceptions to handle our flow control, but Exceptions are costly. In the end, we enriched our
evaluation Details with errorCode and errorMessage. If desired, the providers can also handle this to reduce the execution footprint in erroneous cases, which don't have to be exceptions. Till now, we also converted those into exceptions and threw them, with this little code change, we would return already the details, and call the hooks, without exceptions being thrown.

For example, FlagNotFound is such a case; it can be expected and not exceptional. Hence, the exception flow might add too much overhead to performance-critical environments.

This implementation now allows the prover implementors to decide whether to support either or both.

### Follow-up Tasks

Most likely this is something we also want to do in all other sdks.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -135,7 +135,7 @@ public class OpenFeatureClient implements Client {
                 OpenFeatureError error = ExceptionUtils.instantiateErrorByErrorCode(
                         details.getErrorCode(),
                         details.getErrorMessage());
-                details.setValue(defaultValue);
+                enrichDetailsWithErrorDefaults(defaultValue, details);
                 hookSupport.errorHooks(type, afterHookContext, error, mergedHooks, hints);
             } else {
                 hookSupport.afterHooks(type, afterHookContext, details, mergedHooks, hints);
@@ -150,14 +150,18 @@ public class OpenFeatureClient implements Client {
                 details.setErrorCode(ErrorCode.GENERAL);
             }
             details.setErrorMessage(e.getMessage());
-            details.setValue(defaultValue);
-            details.setReason(Reason.ERROR.toString());
+            enrichDetailsWithErrorDefaults(defaultValue, details);
             hookSupport.errorHooks(type, afterHookContext, e, mergedHooks, hints);
         } finally {
             hookSupport.afterAllHooks(type, afterHookContext, mergedHooks, hints);
         }
 
         return details;
+    }
+
+    private static <T> void enrichDetailsWithErrorDefaults(T defaultValue, FlagEvaluationDetails<T> details) {
+        details.setValue(defaultValue);
+        details.setReason(Reason.ERROR.toString());
     }
 
     /**
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -21,7 +21,7 @@ import lombok.extern.slf4j.Slf4j;
  * You should not instantiate this or reference this class.
  * Use the dev.openfeature.sdk.Client interface instead.
  * @see Client
- * 
+ *
  * @deprecated // TODO: eventually we will make this non-public. See issue #872
  */
 @Slf4j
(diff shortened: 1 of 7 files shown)


aepfli:
I am not good with performance tests, but I tried the following locally:

```java

        IntStream.range( 0, 1000000 ).parallel().forEach(i -> {
            String key = "key" + i;
            c.getBooleanDetails(key, false);
        });
```
With the exceptions, it took more than 3s; with details, it took only 1.5s.

Remarks:
- I changed the implementation of the `AlwaysBrokenWithDetailsProvider` to set a different message all the time so that we never create the same object if the JIT somehow starts to optimize.
- I removed the asserts because they took most of the time, and i was only aiming for run speed.
- I ran those tests for the fun about 5 times, with close results.
- I know this is not the best performance test, but it gives a little bit of insights

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -132,7 +132,11 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
 
             details = FlagEvaluationDetails.from(providerEval, key);
             if (details.getErrorCode() != null) {
-                throw ExceptionUtils.instantiateErrorByErrorCode(details.getErrorCode(), details.getErrorMessage());
+                OpenFeatureError error = ExceptionUtils.instantiateErrorByErrorCode(
+                        details.getErrorCode(),
+                        details.getErrorMessage());
+                details.setValue(defaultValue);
```
guidobrei, line 138: The previous implementation also sets the reason on the FlagEvaluationDetails

toddbaert: Yep this is important. The spec dictates it should be set in error conditions. Regardless of what the provider sets, we should set the reason to ERROR.

toddbaert (changes requested):
I agree with @guidobrei 's point [here](https://github.com/open-feature/java-sdk/pull/1095/files#r1751449602).

In addition, I'm wondering if we should remove stack traces from some or all Errors, in addition to the FlagNotFound.

I think for a few, like `ProviderNotReady` and `TypeMismatch` we can also remove the stack trace, especially since these will be frequently instantiated by the SDK here and it's very obvious what the problem is.

aepfli:
> In addition, I'm wondering if we should remove stack traces from some or all Errors, in addition to the FlagNotFound.
> 
> I think for a few, like `ProviderNotReady` and `TypeMismatch` we can also remove the stack trace, especially since these will be frequently instantiated by the SDK here and it's very obvious what the problem is.

I added this via a parent class to the errors mentioned. This way, we can easily identify them later. Wdyt? (we can also try to achieve this with an interface and default methods)

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
aepfli: added and also added to the tests for verification

guidobrei:
> I think for a few, like `ProviderNotReady` and `TypeMismatch` we can also remove the stack trace, especially since these will be frequently instantiated by the SDK here and it's very obvious what the problem is.

Not sure for the `TypeMismatch`. If `client.getBooleanValue("flag-key", false);` is called from user code for a string flag, I'd like to see the root cause.

toddbaert:
> > I think for a few, like `ProviderNotReady` and `TypeMismatch` we can also remove the stack trace, especially since these will be frequently instantiated by the SDK here and it's very obvious what the problem is.
> 
> Not sure for the `TypeMismatch`. If `client.getBooleanValue("flag-key", false);` is called from user code for a string flag, I'd like to see the root cause.

OK I'm fine with that.

toddbaert (approved):

toddbaert:
Approved but you can make the `TypeMismatch` change @guidobrei suggestes.

jarebudev:
> ~I am not good with performance tests, but I tried the following locally:~
> 
> ```java
>         IntStream.range( 0, 1000000 ).parallel().forEach(i -> {
>             String key = "key" + i;
>             c.getBooleanDetails(key, false);
>         });
> ```
> 
> ~With the exceptions, it took more than 3s; with details, it took only 1.5s.~
> 
> Remarks:
> 
> * I changed the implementation of the `AlwaysBrokenWithDetailsProvider` to set a different message all the time so that we never create the same object if the JIT somehow starts to optimize.
> * I removed the asserts because they took most of the time, and i was only aiming for run speed.
> * I ran those tests for the fun about 5 times, with close results.
> * I know this is not the best performance test, but it gives a little bit of insights
> 
> // EDIT: forget about my tests, i might have made a mistake - i cant reproduce this - hence that, those are bad tests.
> 
> // EDIT: https://www.baeldung.com/java-exceptions-performance is a nice comprehensive post about this

This has got me wondering if there is value in creating some JMH benchmarks for the sdk...?

Kavindu-Dodan (approved):

beeme1mr (approved):

Issue #844, opened 2024-03-15
[FEATURE] Make provider interface "stateless", SDK maintains provider state


toddbaert:
#### Stateless providers

Providers no longer maintain their own state: the state for each provider is maintained in the SDK automatically, and updated according to the success/failures of lifecycle methods (init/shutdown) or events emitted from providers spontaneously.

Similar work in C#: https://github.com/open-feature/dotnet-sdk/pull/276

```[tasklist]
### Tasks
- [x] [Requirement 1.7.1](https://openfeature.dev/specification/sections/flag-evaluation#requirement-171): The client MUST define a provider status accessor which indicates the readiness of the associated provider, with possible values NOT_READY, READY, STALE, ERROR, or FATAL.
- [x] [Requirement 1.7.3](https://openfeature.dev/specification/sections/flag-evaluation#requirement-173): The client's provider status accessor MUST indicate READY if the initialize function of the associated provider terminates normally.
- [x] [Requirement 1.7.4](https://openfeature.dev/specification/sections/flag-evaluation#requirement-174): The client's provider status accessor MUST indicate ERROR if the initialize function of the associated provider terminates abnormally.
- [x] [Requirement 1.7.5](https://openfeature.dev/specification/sections/flag-evaluation#requirement-175): The client's provider status accessor MUST indicate FATAL if the initialize function of the associated provider terminates abnormally and indicates error code PROVIDER_FATAL.
- [x] [Requirement 1.7.6](https://openfeature.dev/specification/sections/flag-evaluation#requirement-176): The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in NOT_READY.
- [x] [Requirement 1.7.7](https://openfeature.dev/specification/sections/flag-evaluation#requirement-177): The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in FATAL.
- [x] [Requirement 1.7.8](https://openfeature.dev/specification/sections/flag-evaluation#requirement-178): Implementations SHOULD propagate the error code returned from any provider lifecycle methods.
- [x] [Requirement 5.1.5](https://openfeature.dev/specification/sections/events#requirement-515): PROVIDER_ERROR events SHOULD populate the provider event details's error code field.
- [x] [Requirement 5.3.5](https://openfeature.dev/specification/sections/events#requirement-535): If the provider emits an event, the value of the client's provider status MUST be updated accordingly.
```

gunishmatta:
[...]

Issue #867, opened 2024-03-25
Maven dependency for Jakarta and Java 17


Exnadella:
Dear Team,

We're requesting the release of the dependency for Jakarta and Java 17, as we're transitioning most of our stack from Java 8 to Java 17, which necessitates migrating from Javax to Jakarta for compatibility. Additionally, could we consider using Spring Boot 3.2.0 for demonstration purposes?

Kavindu-Dodan:
@Exnadella thank you for raising concern over the `Javax` annotation. I can confirm that OpenFeature Java SDK uses `javax.annotation.Nullable`. 

Could you please share any issues that you faced during the migration to Java 17 ? For example, do you have checks for `Javax` ? Otherwise,  if you intend to avoid dependency on `jsr305`, then you may do this by adding an exclusion rule to OF Java SDK. 

For example, for maven 

```xml
        <dependency>
            <groupId>dev.openfeature</groupId>
            <artifactId>sdk</artifactId>
            <version>1.7.6</version>
            <exclusions>
                <exclusion>
                    <groupId>com.google.code.findbugs</groupId>
                    <artifactId>jsr305</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

A missing annotation should not fail your build or application.

Kavindu-Dodan:
@toddbaert yeah, that's an option and a quick solution for this

toddbaert:
@DBlanchard88 is working on this.

toddbaert:
@Exnadella We have removed all these annotations. The [next release](https://github.com/open-feature/java-sdk/pull/865) will not contain them.

PR #1096, opened 2024-09-09
feat: make provider interface "stateless"; SDK maintains provider state

chrfwow:
## This PR

Providers no longer maintain their own state: the state for each provider is maintained in the SDK automatically, and updated according to the success/failures of lifecycle methods (init/shutdown) or events emitted from providers spontaneously.

### Related Issues

https://github.com/open-feature/java-sdk/issues/844

Please remove `getState` in implementing providers (the interface method has been marked as deprecated).


src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -94,7 +138,7 @@ public void emitProviderStale(ProviderEventDetails details) {
     /**
      * Emit a {@link ProviderEvent#PROVIDER_ERROR} event.
      * Shorthand for {@link #emit(ProviderEvent, ProviderEventDetails)}
-     * 
+     *
      * @param details The details of the event
      */
     public void emitProviderError(ProviderEventDetails details) {
```
chrfwow, line 144: should there also be a method like `emitFatalProviderError`?

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
```diff
@@ -1,51 +1,47 @@
...
-import dev.openfeature.sdk.fixtures.HookFixtures;
-
 class DeveloperExperienceTest implements HookFixtures {
     transient String flagKey = "mykey";
 
-    @Test void simpleBooleanFlag() {
+    @Test void simpleBooleanFlag() throws Exception {
         OpenFeatureAPI api = OpenFeatureAPI.getInstance();
-        api.setProvider(new NoOpProvider());
+        api.setProvider(TestEventsProvider.initialized());
```
chrfwow, line 21: Some of the tests had to be changed, as now an uninitialized provider will not allow the lookup of a value.

toddbaert:
Thanks a lot for this! I will try to review tomorrow! :pray:

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -47,88 +29,81 @@ public InMemoryProvider(Map<String, Flag<?>> flags) {
...
-        if (!ProviderState.READY.equals(state)) {
-            if (ProviderState.NOT_READY.equals(state)) {
+        if (!ProviderState.READY.equals(getState())) {
+            if (ProviderState.NOT_READY.equals(getState())) {
                 throw new ProviderNotReadyError("provider not yet initialized");
             }
+            if (ProviderState.FATAL.equals(getState())) {
+                throw new FatalError("provider in fatal error state");
+            }
             throw new GeneralError("unknown error");
```
chrfwow, line 107: The behaviour here is different than that of the `OpenFeatureClient`, which would not throw this `GeneralError`. Which bevaviour is the expected one?

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -16,21 +18,56 @@
...
+            throw openFeatureError;
+        } catch (Exception e) {
+            providerState = ProviderState.ERROR;
+            throw new GeneralError(e);
+        }
+    }
+
+    protected void doInitialization(EvaluationContext evaluationContext) throws Exception {
+        // Intentionally left blank, to be implemented by inheritors
+    }
```
toddbaert, line 51: This implementation is logically correct, but this won't work retroactively with existing implementations (many of which can be found [here](https://github.com/open-feature/dotnet-sdk/pull/276), though there's others maintained by vendors, etc).

We need to do this in a non-breaking way, without breaking any existing contracts or behvaiors. In [C#,](https://github.com/open-feature/dotnet-sdk/pull/276) I accomplished this by creating a new package-private ProviderState field on the provider which the SDK updates after it has successfully (or unsuccessfully) run the `initialize` method, and basically ignoring the previous provider state field.

This allows existing providers to work without breaking changes.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -100,15 +96,18 @@ public Client getClient(String domain) {
...
-     * @param domain a identifier which logically binds clients with providers
+     *
+     * @param domain  a identifier which logically binds clients with providers
      * @param version a version identifier
      * @return a new client instance
      */
     public Client getClient(String domain, String version) {
-        return new OpenFeatureClient(this,
+        return new OpenFeatureClient(
+                () -> providerRepository.getProvider(domain),
```
toddbaert, line 106: Inside the `ProviderRespository`, you will have to modify [this private method](https://github.com/open-feature/java-sdk/blob/64082617fade3fa7d647302cc0c5b698f7038d81/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L129) to call the initialize function on the provider if it's not yet initialized or in the process of being initialized, then update a new package-private status member indicating readiness.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
```diff
@@ -47,88 +29,81 @@ public InMemoryProvider(Map<String, Flag<?>> flags) {
...
-     * @param evaluationContext evaluation context
-     * @throws Exception on error
-     */
-    @Override
-    public void initialize(EvaluationContext evaluationContext) throws Exception {
-        super.initialize(evaluationContext);
-        state = ProviderState.READY;
-        log.debug("finished initializing provider, state: {}", state);
-    }
-
```
toddbaert, line 61: As a proof that nothing is breaking, you should be able to get away with making no changes to this provider.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -50,11 +87,18 @@ void detach() {
...
      * @param details The details of the event
      */
     public void emit(ProviderEvent event, ProviderEventDetails details) {
+        if (ProviderEvent.PROVIDER_ERROR.equals(event)) {
+            providerState = ProviderState.ERROR;
+        } else if (ProviderEvent.PROVIDER_STALE.equals(event)) {
+            providerState = ProviderState.STALE;
+        } else if (ProviderEvent.PROVIDER_READY.equals(event)) {
+            providerState = ProviderState.READY;
+        }
```
toddbaert, line 101: Can you add a specific test for this? It relates to specification requirement [5.3.5](https://openfeature.dev/specification/sections/events/#requirement-535), so you can use our `@Specification` annotation for tracking compliance.

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: I would take the approach of just checking if the error is a Fatal error.

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
toddbaert: It might be better to use `setProviderAndWait` which will initialize the provider before continuing.

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
toddbaert: See: https://github.com/open-feature/java-sdk/pull/1096/files/98c620f14a81a0fda47207eb9d912826a33ec446#r1752386247

We should be able to delete this code entirely since initialization is tracked in the SDK, but things should also work fine if we don't touch this class at all.

toddbaert (changes requested):
Really good first pass! I've left some comments. The most important is [here](https://github.com/open-feature/java-sdk/pull/1096#discussion_r1752376425).

If you need more help or clarification, please reach out on slack.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -16,21 +18,56 @@
...
+    }
+
+    protected void doInitialization(EvaluationContext evaluationContext) throws Exception {
+        // Intentionally left blank, to be implemented by inheritors
+    }
+
+    @Override
+    public final void shutdown() {
+        providerState = ProviderState.NOT_READY;
+        doShutdown();
```
liran2000, line 56: why this split is needed ?
without it, provider can implement shutdown() same as today and calling _super.shutdown()_ in it.
providers which were implemented this way, can continue to work as is.
similar with initialize().

src/test/java/dev/openfeature/sdk/testutils/TestEventsProvider.java
```diff
@@ -35,8 +26,10 @@ public TestEventsProvider(int initTimeoutMs, boolean initError, String initError
         this.initErrorMessage = initErrorMessage;
     }
 
-    public TestEventsProvider(ProviderState initialState) {
-        this.state = initialState;
+    public static TestEventsProvider initialized() throws Exception {
```
liran2000, line 29: rename to something like _newTestEventsProvider()_ ?

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
```diff
@@ -1,51 +1,47 @@
...
-import dev.openfeature.sdk.testutils.FeatureProviderTestUtils;
-import org.junit.jupiter.api.Test;
-
-import dev.openfeature.sdk.fixtures.HookFixtures;
-
 class DeveloperExperienceTest implements HookFixtures {
     transient String flagKey = "mykey";
 
-    @Test void simpleBooleanFlag() {
+    @Test void simpleBooleanFlag() throws Exception {
```
liran2000, line 19: can use _@SneakyThrows_.
also at other places.

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
```diff
@@ -94,8 +90,12 @@ class MutatingHook implements Hook {
 
             @Override
             // change the provider during a before hook - this should not impact the evaluation in progress
-            public Optional before(HookContext ctx, Map hints) {
-                FeatureProviderTestUtils.setFeatureProvider(new NoOpProvider());
+            public Optional before(HookContext ctx, Map hints)  {
```
liran2000, line 93: can use _@SneakyThrows_ and remove try/catch ?

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -105,8 +100,8 @@ void getApiInstance() {
     @Test void shouldReturnNotReadyIfNotInitialized() {
         FeatureProvider provider = new InMemoryProvider(new HashMap<>()) {
             @Override
-            public void initialize(EvaluationContext evaluationContext) throws Exception {
-                Awaitility.await().wait(3000);
+            protected void doInitialization(EvaluationContext evaluationContext) throws Exception {
+                Thread.sleep(10000);
```
liran2000, line 104: why change the implementation and wait that much longer ?

src/test/java/dev/openfeature/sdk/providers/memory/InMemoryProviderTest.java
```diff
@@ -125,4 +118,15 @@ void emitChangedFlagsOnlyIfThereAreChangedFlags() {
...
     }
+
+    @SneakyThrows
+    @Test
+    void shouldThrowGeneralErrorAfterEmittingErrorEvent() {
+        InMemoryProvider inMemoryProvider = new InMemoryProvider(new HashMap<>());
+        inMemoryProvider.initialize(null);
+        inMemoryProvider.emitProviderError(ProviderEventDetails.builder().build());
+
+        // ErrorCode.GENERAL should be returned when evaluated via the client
```
liran2000, line 129: why would it return error after provider error event ?
any reference to this behavior ?

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
beeme1mr: ^ Apparently, that's someones username 😄

src/main/java/dev/openfeature/sdk/EventProvider.java
liran2000: @toddbaert I wonder if there can be some additional automated step to see if the _sdk-contrib_ repository providers are not breaking, possibly as warning / non-mandatory. There is the _InMemoryProvider_ on this repository which is a good sanity, maybe checking the other providers as well can help. Or is too complex ? What do you think ?

src/main/java/dev/openfeature/sdk/providers/memory/InMemoryProvider.java
aepfli: so do you suggest to generate a Second deprecated InMemoryProvider to always ensure backwards compatibility and run the tests for both? or do you think that is overkill?

src/main/java/dev/openfeature/sdk/EventProvider.java
chrfwow: How can I do that? The `ProviderEventDetails` parameter does not have a field containing a potential error code. If I can get it from `getEventMetadata()`, what `key` is the error code associated with?

src/main/java/dev/openfeature/sdk/FeatureProviderWrapper.java
```diff
@@ -0,0 +1,118 @@
+package dev.openfeature.sdk;
+
+import dev.openfeature.sdk.exceptions.OpenFeatureError;
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class FeatureProviderWrapper implements FeatureProvider, EventProviderListener {
```
toddbaert, line 8:
```suggestion
class FeatureProviderWrapper implements FeatureProvider, EventProviderListener {
```
I guess this doesn't need to be public?

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -2,6 +2,8 @@
 
 import dev.openfeature.sdk.internal.TriConsumer;
 
+import javax.annotation.Nullable;
```
toddbaert, line 5: We can't use javax, unfortunately. It's non-standard and creates a problematic dependecny for consumers (see https://github.com/open-feature/java-sdk/issues/867)

src/main/java/dev/openfeature/sdk/FeatureProviderWrapper.java
```diff
@@ -0,0 +1,118 @@
...
+    private final FeatureProvider delegate;
+    private final AtomicBoolean isInitialized = new AtomicBoolean();
+    private ProviderState state = ProviderState.NOT_READY;
+
+    public FeatureProviderWrapper(FeatureProvider delegate) {
+        this.delegate = delegate;
+        if (delegate instanceof EventProvider) {
+            ((EventProvider) delegate).setEventProviderListener(this);
+        }
+    }
```
toddbaert, line 18: Wrapping and delegating is a great solution here! Nice idea. It may be easier to do this with lombok's delegate though (see where we do it [here](https://github.com/open-feature/java-sdk/blob/c68f78e17b6125d74644a5735f9d58348edcc383/src/main/java/dev/openfeature/sdk/MutableContext.java#L24))

src/test/java/dev/openfeature/sdk/OpenFeatureClientTest.java
```diff
@@ -78,23 +83,99 @@ void mergeContextTest() {
...
+        MockProvider mockProvider = new MockProvider(ProviderState.NOT_READY);
+        OpenFeatureAPI api = mock(OpenFeatureAPI.class);
+        OpenFeatureClient client = new OpenFeatureClient(() -> mockProvider, api, "name", "version");
+        FlagEvaluationDetails<Boolean> details = client.getBooleanDetails("key", true);
+
+        assertThat(mockProvider.isEvaluationCalled()).isFalse();
+        assertThat(details.getErrorCode()).isEqualTo(ErrorCode.PROVIDER_NOT_READY);
+    }
+
+    private static class MockProvider implements FeatureProvider {
```
toddbaert, line 129: You might be able to use the `TestEventsProvider` for this but that's up to you.

toddbaert:
You have some checkstyle violations: https://github.com/open-feature/java-sdk/actions/runs/10825623982/job/30071997539?pr=1096

You can see these (and more) by running `mvn verify` locally.

src/test/java/dev/openfeature/sdk/AlwaysBrokenProvider.java
```diff
@@ -4,11 +4,11 @@
 
 public class AlwaysBrokenProvider implements FeatureProvider {
 
+    private final String name = "always broken";
+
     @Override
     public Metadata getMetadata() {
-        return () -> {
-            throw new FlagNotFoundError(TestConstants.BROKEN_MESSAGE);
-        };
```
toddbaert, line 11: This causes an error during init, which we don't really want.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -99,7 +99,7 @@ void getApiInstance() {
 
     @Specification(number="2.4.5", text="The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.")
     @Test void shouldReturnNotReadyIfNotInitialized() {
-        FeatureProvider provider = new InMemoryProvider(new HashMap<>()) ;
+        FeatureProvider provider = new TestEventsProvider(100);
```
toddbaert, line 102: This test was flaky, since after your changes, it was possible that the provider was initialized before line 105 ran. This adds an artificial delay to init to make sure.

Notice that I'm NOT using `setProviderAndWait` here because we explicitly DONT want the provider to be ready for this test.

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: Good point. I think that's another thing we need to add. I might have missed it in the issue, but it's in the spec: https://openfeature.dev/specification/types#event-details

toddbaert (changes requested):
Really, really good work.

You have some checkstyle violations: https://github.com/open-feature/java-sdk/actions/runs/10825623982/job/30071997539?pr=1096

You can see these (and more) by running `mvn verify` locally. I also found some flaky tests [which I pushed some simple fixes for](https://github.com/open-feature/java-sdk/pull/1096/commits/724596319a3b3b81adb29a224dfb6db0c0e63518); basically ignoring the providers state breaks some of the assumptions in these tests; the good news is most of them are easily fixed by simply using `setProviderAndWait` to make them deterministic.

I think these are the only things blocking my approval:

- https://github.com/open-feature/java-sdk/pull/1096/files#r1757398053
- https://github.com/open-feature/java-sdk/pull/1096/files#r1757389512
- https://github.com/open-feature/java-sdk/pull/1096/files#r1757507982
- https://github.com/open-feature/java-sdk/pull/1096/files#r1757408195
- (optional, but recommended) https://github.com/open-feature/java-sdk/pull/1096/files#r1757401571

src/main/java/dev/openfeature/sdk/EventProvider.java
chrfwow: Is it queried like this?
```
String errorCode = details.getEventMetadata().getString("error code");
if("FATAL".equals(errorCode)){
    ...
}
```
I can't see how this should work from the documentation

guidobrei (commented):
Not sure if we want to have SDK side state machine mandatory.

If yes, the wrapper concept makes sense. However, this approach is a little complex to implement and there are some downsides:
* `Provider.getState()` does not really reflect the provider state anymore
* If someone gets the hands on a `FeatureProvider` which is not a `FeatureProviderWrapper` or a `EventProvider` and calls a method that should cause a state transition, this transition will not be seen by the SDK

I would prefer an opt-in approach, where providers can outsource state management to the SDK but don't have to.

What do you think?

src/main/java/dev/openfeature/sdk/NoOpProvider.java
```diff
@@ -59,11 +59,22 @@ public ProviderEvaluation<Double> getDoubleEvaluation(String key, Double default
...
                 .reason(Reason.DEFAULT.toString())
                 .build();
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (obj instanceof FeatureProviderWrapper) {
```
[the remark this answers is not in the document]
chrfwow: This is needed to keep the current tests succeeding.  
Furthermore, a `FeatureProviderWrapper` should be equal to its `FeatureProvider`, as they are one functional unit, and should behave to the outside as if there was no wrapper. Consequently, it the `FeatureProviderWrapper` equals its `FeatureProvider`, the reverse should also be true.

toddbaert:
> Not sure if we want to have SDK side state machine mandatory.
> 
> If yes, the wrapper concept makes sense. However, this approach is a little complex to implement and there are some downsides:
> 
> * `Provider.getState()` does not really reflect the provider state anymore
> * If someone gets the hands on a `FeatureProvider` which is not a `FeatureProviderWrapper` or a `EventProvider` and calls a method that should cause a state transition, this transition will not be seen by the SDK
> 
> I would prefer an opt-in approach, where providers can outsource state management to the SDK but don't have to.
> 
> What do you think?

@guidobrei 

The init State machine is mandatory now as it's in the OpenFeature spec. This sdk is one of the last implementations not to have it. We've had lots of issues in the past with providers being confused about updating this state; this makes the old state useless, yes, but it's easily removed.

There's a blog post about why the community decided on this change here: https://openfeature.dev/blog/reconciling-with-state

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: I'm saying the event details should have a first class member representing the error code (not in the event data).

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -50,11 +49,14 @@ void detach() {
 
     /**
      * Emit the specified {@link ProviderEvent}.
-     * 
+     *
      * @param event   The event type
      * @param details The details of the event
      */
     public void emit(ProviderEvent event, ProviderEventDetails details) {
```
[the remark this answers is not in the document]
toddbaert: I think this is a good point, and we can probably reduce this visibility without considering this a breaking change since events are still labeled "experimental" so that would break our stability guarantee.

src/main/java/dev/openfeature/sdk/NoOpProvider.java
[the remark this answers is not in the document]
guidobrei: > Consequently, it the `FeatureProviderWrapper` equals its `FeatureProvider`, the reverse should also be true.

But we can't ensure this with this implementation strategy, because we do not know who is implementing the `FeatureProvider` interface. All implementations of `FeatureProvider` out there (that do not extend `NoopProvider` or `EventProvider` are not covered by this approach and will behave differently.

guidobrei:
> The init State machine is mandatory now as it's in the OpenFeature spec.

@toddbaert 

If we have now a breaking change in the spec, why do we want to do a non-breaking change in the SDK?
The way the Java SDK is built I see little hope to build it clean and non-breaking. 

@chrfwow approach with wrapping all incoming providers in a `StatefulFeatureProvider` seems promising. However, it still has some breaking behavior: Whatever FeatureProvider you add to the SDK, you get a StatefulFeatureProvider when accessing it again. 

Just adding the wrapper as backwards compatible implementation feels a like a workaround. We should go for a proper solution with an additional workaround for existing providers.
We could make the `StatefulFeatureProvider` public abstract and put it in the inheritance chain of our known Providers by making it the base class of all Provider implementations, and additionally ask other implementers to inherit from this class from now on. Maybe we need some additional `StatefulEventProvider`, as we have to ensure that `initialize()`, `shutdown()`... are called on the base class. 
In the meanwhile, we can stay backwards compatible, by wrapping all other implementations that directly implement `FeatureProvider` into a `StatefulFeatureProviderWrapper` (the current implementation by @chrfwow ). This way we minimize the current limitation of not having access to the actual provider that was added to SDK.

chrfwow:
> However, it still has some breaking behavior: Whatever FeatureProvider you add to the SDK, you get a StatefulFeatureProvider when accessing it again.
I could correct this behaviour by returning the delegate in the `ProviderRepository.getProvider()` method as in the [latest commit](https://github.com/open-feature/java-sdk/pull/1096/commits/4bc6f664d2efc3db90aa2184f33a66f31b596b74)

src/main/java/dev/openfeature/sdk/StatefulFeatureProvider.java
```diff
@@ -0,0 +1,99 @@
+package dev.openfeature.sdk;
+
+import dev.openfeature.sdk.exceptions.OpenFeatureError;
+import lombok.experimental.Delegate;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+class StatefulFeatureProvider implements FeatureProvider, EventProviderListener {
```
toddbaert, line 8: As we discussed, let's remove this implementation of `FeatureProvider`, and the delegation of the associated methods.

Instead, as @guidobrei suggests, let's make this a true wrapper class that adds statefulness and encapsulates a `FeatureProvider` and providers an accessor for it. Then we can avoid having to implement the equality overrides such as [this](https://github.com/open-feature/java-sdk/pull/1096/files#diff-7bbe5cc5c58570709c6e4af7221fde633a9171736c3992c16fe9ff5d78fa75e2R69-R79) and [this](https://github.com/open-feature/java-sdk/pull/1096/files#diff-0bc88e6e0f33e7a2876841d5a59be2ceb032f4600719d8c7668728e473aec94eR70-R79).

toddbaert: I `FeatureProviderStateManager` or something like that might be a good name since this will no longer actually be a provider.

src/main/java/dev/openfeature/sdk/ProviderAccessor.java
```diff
@@ -0,0 +1,9 @@
+package dev.openfeature.sdk;
+
+/**
+ * Provides access to the future provider for the domain of the client.
+ */
+@FunctionalInterface
+interface ProviderAccessor {
+    FeatureProviderStateManager getProviderStateManager();
```
chrfwow, line 8: I know we talked that the FeatureProviderStateManager should at best be invisible outside the PrviderRepository, but I don't see another way to make [this](https://github.com/open-feature/java-sdk/pull/1096/files#diff-e0fc1e23041e08fa8dd72864acce5b949f84987f405accd4935f233de281ef14R125) thread safe

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -99,68 +147,103 @@ public void setProvider(String domain,
...
+                                              FeatureProvider newProvider,
+                                              Consumer<FeatureProvider> afterSet,
+                                              Consumer<FeatureProvider> afterInit,
+                                              Consumer<FeatureProvider> afterShutdown,
+                                              BiConsumer<FeatureProvider, OpenFeatureError> afterError,
+                                              boolean waitForInit) {
+        final FeatureProviderStateManager newStateManager;
+        final FeatureProviderStateManager oldStateManager;
+
+        synchronized (registerStateManagerLock) {
```
chrfwow, line 159: neede to prevend a race condition, afterSet could be invoked more than once

src/test/java/dev/openfeature/sdk/EventsTest.java
```diff
@@ -594,16 +596,18 @@ void mustPersistAcrossChanges() {
...
+        @SneakyThrows
         void removedEventsShouldNotRun() {
             final String name = "removedEventsShouldNotRun";
             final Consumer<EventDetails> handler1 = mockHandler();
             final Consumer<EventDetails> handler2 = mockHandler();
 
             TestEventsProvider provider = new TestEventsProvider(INIT_DELAY);
-            OpenFeatureAPI.getInstance().setProvider(name, provider);
+            provider.initialize(null);
+            OpenFeatureAPI.getInstance().setProviderAndWait(name, provider);
```
toddbaert, line 610:
```suggestion
            OpenFeatureAPI.getInstance().setProviderAndWait(name, provider);
```

At least with your latest changes this works fine. I can't make sense of the manual init call.

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
```diff
@@ -111,9 +117,59 @@ public Optional before(HookContext ctx, Map hints) {
...
+        OpenFeatureAPI api = OpenFeatureAPI.getInstance();
+        TestEventsProvider provider = new TestEventsProvider();
+        api.setProviderAndWait(domain, provider);
+        Client client = api.getClient(domain);
+        assertThat(client.getProviderState()).isEqualTo(ProviderState.READY);
+        provider.emitProviderStale(ProviderEventDetails.builder().build());
+        assertThat(client.getProviderState()).isEqualTo(ProviderState.STALE);
+        provider.emitProviderReady(ProviderEventDetails.builder().build());
+        assertThat(client.getProviderState()).isEqualTo(ProviderState.READY);
+    }
```
toddbaert, line 174: Thanks for adding these "integration style" tests! :raised_hands:

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -99,68 +147,103 @@ public void setProvider(String domain,
...
     }
 
     /**
      * Helper to check if provider is already known (registered).
+     *
      * @param provider provider to check for registration
      * @return boolean true if already registered, false otherwise
      */
-    private boolean isProviderRegistered(FeatureProvider provider) {
+    private boolean isStateManagerRegistered(FeatureProviderStateManager provider) {
```
toddbaert, line 237: Nit: update the param name here to be `manager` or something like that.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -99,68 +147,103 @@ public void setProvider(String domain,
...
         return provider != null
-                && (this.providers.containsValue(provider) || this.defaultProvider.get().equals(provider));
+                && (this.stateManagers.containsValue(provider) || this.defaultStateManger.get().equals(provider));
+    }
+
+    private void shutdownProvider(FeatureProviderStateManager provider) {
```
toddbaert, line 242: [Same](https://github.com/open-feature/java-sdk/pull/1096/files#r1765364051)

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -63,41 +65,52 @@ public void emit(ProviderEvent event, ProviderEventDetails details) {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (obj instanceof FeatureProviderStateManager) {
+            return this == ((FeatureProviderStateManager) obj).getProvider();
+        }
+        return this == obj;
+    }
```
toddbaert: I removed this; I don't think we need it anymore since your other changes.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -286,6 +285,24 @@ public FeatureProvider getProvider(String domain) {
...
+    }
+
+    /**
+     * Get the state of the provider for a domain. If no provider with the domain is found, returns the state of the
+     * default provider.
+     *
+     * @param domain The domain to look for.
+     * @return A named {@link FeatureProvider}
+     */
+    public ProviderState getProviderState(String domain) {
```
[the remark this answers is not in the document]
toddbaert: @chrfwow did this but I had to revert it. We had multiple providers in the contribs who used an encapsulation strategy and used this publicly instead of in a subclass.

toddbaert (approved):
I made a couple [small changes](https://github.com/open-feature/java-sdk/pull/1096/commits/124be3644c768d43c229dbae9eca9971518e355d) to remove one more `equals` we didn't need and fix some PMD issues. I also locally build and linked this to the contrib repo and used that version, and everything works, including the flagd e2e tests which are quite robust.

I'm approving, thanks a lot!

cc @guidobrei @aepfli @beeme1mr @liran2000 

Codecov has failed but I'm happy with the coverage. The diff is 85% covered (it wants 95%) and the tests we need are there, IMO.

beeme1mr (approved):
It looks good to me! Please update the provider section of the readme to reflect state changes in this PR or as a follow-up. Thanks

https://github.com/open-feature/java-sdk?tab=readme-ov-file#develop-a-provider

aepfli (approved):

toddbaert (approved):

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -48,18 +43,29 @@ public class OpenFeatureClient implements Client {
...
      * @deprecated Do not use this constructor. It's for internal use only.
-     *             Clients created using it will not run event handlers.
-     *             Use the OpenFeatureAPI's getClient factory method instead.
+     *     Clients created using it will not run event handlers.
+     *     Use the OpenFeatureAPI's getClient factory method instead.
      */
     @Deprecated() // TODO: eventually we will make this non-public. See issue #872
-    public OpenFeatureClient(OpenFeatureAPI openFeatureAPI, String domain, String version) {
+    public OpenFeatureClient(
+            ProviderAccessor providerAccessor,
```
guidobrei, line 51: `ProviderAccessor` is exposed outside its defined visibility scope. `OpenFeatureClient` and the constructor are public, but the `ProviderAccessor` is package privet.

Today I learned... I would have expected the compiler to not allow such constructs.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -116,7 +122,15 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(FlagValueType type, String key
...
-            provider = openfeatureApi.getProvider(this.domain);
+            FeatureProviderStateManager stateManager = providerAccessor.getProviderStateManager();
+            provider = stateManager.getProvider();
+            ProviderState state = stateManager.getState();
+            if (ProviderState.NOT_READY.equals(state)) {
+                throw new ProviderNotReadyError("provider not yet initialized");
+            }
+            if (ProviderState.FATAL.equals(state)) {
+                throw new FatalError("provider is in an irrecoverable error state");
+            }
```
guidobrei, line 133: We had some effort in parallel to handle evaluation errors without exceptions in https://github.com/open-feature/java-sdk/pull/1095.

@toddbaert does this also apply to unexpected Provider states?

In general, how do we handle provider state here? If it's now managed by the SDK, does it only forward the evaluation calls to the provider, if it is in `READY` state, or in any state except `NOT_READY` and `FATAL`?

guidobrei (changes requested):
I have one request for change: I think the `@Delegate` annotation is a leftover from the Decorator approach and can be removed.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: How should I handle this? Should I make the constructor package private, as it is deprecated anyway?

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: Currently we only forward calls if the provider is not in error (or fatal), which differs from the implementation of the `InMemoryProvider`, where they are only forwarded when the provider is `READY`. Is this what we want?

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
guidobrei: I would add a package internal accessor to `getProviderStateManager(String domain)`. Then you can access the state manager in the client and you can remove the ´ProviderAccessor` interface completely.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -286,6 +285,28 @@ public FeatureProvider getProvider(String domain) {
...
+        return providerRepository.getProviderState(domain);
+    }
+
+    public ProviderState getProviderState(FeatureProvider provider) {
```
guidobrei, line 306: I know I've requested this, but maybe it's better to not expose this on the public API surface unless someone requests it. We could also get rid of the additional code in the `ProviderRepository` that handles this request again.

Sorry for being too greedy 🙈

guidobrei: This overload would return a `null` state if the provider is not registered, which is what I would expect (or an exception?), but it's still different to the behavior of other `getProviderState()` overloads.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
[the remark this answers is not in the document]
toddbaert: @guidobrei sorry, I put this comment in the wrong place. I meant for it to be about making the emit protected, which we had to revert (https://github.com/open-feature/java-sdk/pull/1096/files/cd56f25cbfb918117ec53a69ec1d55a87ca6422b#r1758249819)

I think you're right about this one. I don't think these need to be public.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
toddbaert: > Currently we only forward calls if the provider is not in error (or fatal), which differs from the implementation of the InMemoryProvider, where they are only forwarded when the provider is READY. Is this what we want?

Yes, this is what we want. Some providers may be in an error state, and still work in some sense (they may send error telemetry, or read from an out-of-date cache, or even try to reconnect before returning). The only states that should short-circuit are FATAL and NOT_READY. This is defined in the spec (and implemented in other SDKs):

https://openfeature.dev/specification/sections/flag-evaluation#requirement-177
https://openfeature.dev/specification/sections/flag-evaluation#requirement-178

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -82,12 +78,12 @@ void getApiInstance() {
     @Test void providerAndWait() {
         FeatureProvider provider = new TestEventsProvider(500);
         OpenFeatureAPI.getInstance().setProviderAndWait(provider);
-        assertThat(api.getProvider().getState()).isEqualTo(ProviderState.READY);
+        assertThat(api.getProviderState()).isEqualTo(ProviderState.READY);
```
toddbaert, line 81: We can instead get a client (even before we set the provider) and then check the Provider state there, so we can remove this `getProviderState`.

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: Yes I guess the only reason this was done was for [this assertion](https://github.com/open-feature/java-sdk/pull/1096/files#r1768715805) which we can avoid. I agree we should remove these methods.

toddbaert:
@chrfwow I think I agree with @guidobrei 's points. Thanks for being so patient with all these reviews! I think we're almost there but I think the last few changes seem small.

src/main/java/dev/openfeature/sdk/EventProvider.java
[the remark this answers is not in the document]
toddbaert: @chrfwow did this but I had to revert it. We had multiple providers in the contribs who used an encapsulation strategy and used this publicly instead of in a subclass.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
toddbaert: Done: https://github.com/open-feature/java-sdk/pull/1096/commits/45e9ba2e0b33ee84b0449003daff3f03cb820a45

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: Done: https://github.com/open-feature/java-sdk/pull/1096/commits/45e9ba2e0b33ee84b0449003daff3f03cb820a45

src/test/java/dev/openfeature/sdk/OpenFeatureClientTest.java
```diff
@@ -56,36 +56,11 @@ void shouldNotThrowExceptionIfHookHasDifferentTypeArgumentThanHookContext() {
...
-
-        MockProviderRepository mockProviderRepository = new MockProviderRepository(mockProvider, true);
-        OpenFeatureClient client = new OpenFeatureClient(mockProviderRepository, api, "name", "version");
-        client.setEvaluationContext(ctx);
-
-        FlagEvaluationDetails<Boolean> result = client.getBooleanDetails(flag, defaultValue);
-
-        assertThat(result.getValue()).isTrue();
-    }
-
```
toddbaert: This test was brittle and `multi_layer_context_merges_correctly` tests the same functionality much more rigorously.

toddbaert:
Thanks @chrfwow for all your help with this! It works great in the contrib repo so far.

chrfwow:
It was a pleasure!

PR #1098, opened 2024-09-10
chore(style): add editorconfig for easier configuration

aepfli:
Many people struggle with their code styles when contributing for the first time. With this pull request, we're adding a .editorconfig to manage those settings in a centralized place, making everyone's lives easier.

With this iteration, we're adding IntelliJ Idea configuration.

The first commit contains the configuration, and I applied the second one to the whole code base.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,72 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+tab_width = 4
+trim_trailing_whitespace = true
+
+ij_continuation_indent_size = 8
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
+
+# Following the rules of the Google Java Style Guide.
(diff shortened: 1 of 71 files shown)


toddbaert (approved):

.editorconfig
```diff
@@ -0,0 +1,72 @@
...
+
+ij_continuation_indent_size = 8
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
+
+# Following the rules of the Google Java Style Guide.
+# See https://google.github.io/styleguide/javaguide.html
+[*.java]
```
justinabrahms, line 20: Is it possible to get a linter in CI which applies these same requirements? I'd prefer to not rely on someone's editor obeying this config for style correctness.

aepfli: There is still checkstyle in place, which enforces proper styling, but it reduces the pain for new contributors if they have a different code style setup in IntelliJ. For example I see regular people changing to star imports with auto-formatting - which does not need to be the case.

PR #1169, opened 2024-10-18
chore: flaky test

toddbaert:
A test in this suite use the non-blocking provider set, which can result in flakiness in these cases since we aren't listening for a ready event.


src/test/java/dev/openfeature/sdk/OpenFeatureClientTest.java
```diff
@@ -42,7 +42,7 @@ void reset_logs() {
     @DisplayName("should not throw exception if hook has different type argument than hookContext")
     void shouldNotThrowExceptionIfHookHasDifferentTypeArgumentThanHookContext() {
         OpenFeatureAPI api = OpenFeatureAPI.getInstance();
-        api.setProvider("shouldNotThrowExceptionIfHookHasDifferentTypeArgumentThanHookContext", new DoSomethingProvider());
+        api.setProviderAndWait("shouldNotThrowExceptionIfHookHasDifferentTypeArgumentThanHookContext", new DoSomethingProvider());
```
toddbaert, line 45: This test isn't helped by using the non-blocking provider setter here. There are a few tests where we want to not block (to assert behaviors about the SDK before the provider is ready, for instance) but this is not one.

beeme1mr (approved):

PR #1156, opened 2024-10-10
perf: add heap benchmark and reduce allocations

toddbaert:
This PR reduces allocations of heap objects by ~70% for most use-cases. It does this by removing some usage of the streams API in some hotspots, and also being more efficient with `HashMaps` and context merging.

Benchmark results before/after these changes:

```
Benchmark                                         Mode  Cnt          Score   Error      Units
AllocationBenchmark.run                             ss               0.171               s/op
AllocationBenchmark.run:+totalAllocatedBytes        ss       440750800.000              bytes
AllocationBenchmark.run:+totalAllocatedInstances    ss        12086130.000          instances
AllocationBenchmark.run:+totalHeap                  ss       521412608.000  
```

```
Benchmark                                         Mode  Cnt          Score   Error      Units
AllocationBenchmark.run                             ss               0.116               s/op
AllocationBenchmark.run:+totalAllocatedBytes        ss       138760912.000              bytes
AllocationBenchmark.run:+totalAllocatedInstances    ss         5005903.000          instances
AllocationBenchmark.run:+totalHeap                  ss       521412608.000              bytes
```

The performance optimizations are not very cool or exciting; mostly try to avoid creating unnecessary `HashMaps` and don't use streams in hot paths (like context merging in hooks).

PR #1178, opened 2024-10-22
perf: reduce hashmap allocations

toddbaert:
Continuation of https://github.com/open-feature/java-sdk/pull/1156; further reduces allocations, particularly of `ImmutableContext` and `HashMap`.

`HashMap` instances are reduced by more than 66%, memory usage decreased by another 40% in the benchmark.

Open `benchmark.txt` for results.

### DIFF
--- a/benchmark.txt
+++ b/benchmark.txt
@@ -36,9 +36,9 @@ Audit done.
   processing is enabled explicitly (-proc:only, -proc:full).
   Use -Xlint:-options to suppress this message.
   Use -proc:none to disable annotation processing.
-[WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/MutableStructure.java:[19,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
-[WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/ImmutableStructure.java:[22,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
 [WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/EventDetails.java:[9,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
+[WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/ImmutableStructure.java:[22,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
+[WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/MutableStructure.java:[19,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
 [WARNING] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/Value.java:[27,26] finalize() in java.lang.Object has been deprecated and marked for removal
 [INFO] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/NoOpProvider.java: Some input files use or override a deprecated API.
 [INFO] /home/todd/git/java-sdk/src/main/java/dev/openfeature/sdk/NoOpProvider.java: Recompile with -Xlint:deprecation for details.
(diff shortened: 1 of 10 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -183,17 +182,29 @@ private static <T> void enrichDetailsWithErrorDefaults(T defaultValue, FlagEvalu
...
-                : new ImmutableContext();
-        final EvaluationContext clientContext = this.getEvaluationContext() != null
-                ? this.getEvaluationContext()
-                : new ImmutableContext();
-        final EvaluationContext transactionContext = openfeatureApi.getTransactionContext() != null
-                ? openfeatureApi.getTransactionContext()
-                : new ImmutableContext();
-
-        return apiContext.merge(transactionContext.merge(clientContext.merge(invocationContext)));
+        // avoid any unnecessary context instantiations and stream usage here; this is call with every evaluation.
```
toddbaert, line 185: This method is way uglier than before with a nice nested call to `.merge()` but this results in way less memory usage.

src/main/java/dev/openfeature/sdk/Structure.java
```diff
@@ -46,6 +46,14 @@ public interface Structure {
      */
     Map<String, Value> asMap();
 
+    /**
+     * Get all values, as a map of Values.
+     *
+     * @return all attributes on the structure into a Map
+     */
+    Map<String, Value> asUnmodifiableMap();
```
toddbaert, line 54: We can make good use of this internally when we want to just get the attributes as a map but don't care if we actually copy them or not - it just returns an immutable view, not a copy.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
ssharaev: As I see it, we're creating a new `ImmutableContext` for each `merge()` invocation. It seems like we could optimize this a bit more by skipping the merging of empty contexts. How do you feel about that? Here's an example of what I have in mind:
```
        final EvaluationContext apiContext = openfeatureApi.getEvaluationContext();
        final EvaluationContext clientContext = this.getEvaluationContext();
        final EvaluationContext transactionContext = openfeatureApi.getTransactionContext();
        final List<EvaluationContext> contextsToMerge = new ArrayList<>();
        contextsToMerge.add(apiContext);
        contextsToMerge.add(transactionContext);
        contextsToMerge.add(clientContext);
        contextsToMerge.add(invocationContext);

        EvaluationContext merged = new ImmutableContext();
        for (EvaluationContext evaluationContext : contextsToMerge) {
            if (evaluationContext != null && !evaluationContext.isEmpty()) {
                merged = merged.merge(evaluationContext);
            }
        }
        return merged;
```

benchmark.txt
```diff
@@ -70,143 +125,143 @@ Audit done.
...
+ 118:           152           4864  java.nio.file.attribute.FileTime (java.base@21.0.4)
 truncated...
-Total       4474389      138762960
+Total       3237354      105727792
 
-0.113 s/op
-                 +totalAllocatedBytes:     138762960.000 bytes
-                 +totalAllocatedInstances: 4474389.000 instances
+0.132 s/op
+                 +totalAllocatedBytes:     105727792.000 bytes
```
ssharaev, line 254: Wow, great result! 👍

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -79,14 +78,14 @@ public String getTargetingKey() {
     @Override
     public EvaluationContext merge(EvaluationContext overridingContext) {
         if (overridingContext == null || overridingContext.isEmpty()) {
-            return new ImmutableContext(this.asMap());
+            return new ImmutableContext(this.asUnmodifiableMap());
```
guidobrei, line 81: If we're already in an `ImmutableContext` and there's nothing to add, why not just return `this`?

This goes in the direction @ssharaev suggested in https://github.com/open-feature/java-sdk/pull/1178/files#r1811813866

src/main/java/dev/openfeature/sdk/MutableContext.java
```diff
@@ -33,7 +33,7 @@ public MutableContext(String targetingKey) {
     }
 
     public MutableContext(Map<String, Value> attributes) {
-        this("", attributes);
+        this("", new HashMap<>(attributes));
```
guidobrei, line 36: New HashMap is already created in the overloaded constructor.

```suggestion
        this("", attributes);
```

guidobrei, line 36: In the `ImmutableContext` you set the `targetingKey` to null, here to `""`.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
guidobrei: But even if we do so, in the worst case (all contexts are not empty), we create 5 `ImmutableContext` objects per invocation. 
What about pulling the merge logic out into a factory method on the 'ImmutableContext' that accepts 4 named contexts to merge?

src/main/java/dev/openfeature/sdk/ImmutableContext.java
toddbaert: Because then it's not a new instance, and I think always returning a new object from this method is less surprising than _sometimes_ returning the same one.

src/main/java/dev/openfeature/sdk/MutableContext.java
toddbaert: made it null, I don't like the empty strings.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
toddbaert: I've done both of these things, as well as:

- use a new variadic method that looks nicer
- refactor to move the static merge function from `Structure` to `EvaluationContext`

see: https://github.com/open-feature/java-sdk/pull/1178/commits/f3e06863c60d65601ac41a6421b1c0797bde6783

benchmark.txt
toddbaert: Even better now with https://github.com/open-feature/java-sdk/pull/1178#discussion_r1812935142

toddbaert:
@guidobrei @ssharaev Please re-review; according to your advice (including @guidobrei 's comment [here](https://github.com/open-feature/java-sdk/commit/f3e06863c60d65601ac41a6421b1c0797bde6783#r148273195)) I've made additional improvements. Please note, that during this process [I merged an improvement to the benchmark](https://github.com/open-feature/java-sdk/pull/1182) to include more non-empty context data at various levels, so we could better gauge improvements in merging allocations.

With your additional feedback, we now even have a more dramatic improvement: 

before: 

```
+totalAllocatedBytes:     244216640.000 bytes
                 +totalAllocatedInstances: 7264791.000 instances
                 +totalHeap:               521412608.000 bytes
```

after: 

```
                 +totalAllocatedBytes:     139359040.000 bytes
                 +totalAllocatedInstances: 4452140.000 instances
                 +totalHeap:               521412608.000 bytes
```

ssharaev (approved):

src/main/java/dev/openfeature/sdk/AbstractStructure.java
```diff
@@ -18,7 +19,15 @@ public boolean isEmpty() {
...
-        this.attributes = new HashMap<>(attributes);
+        this.attributes = attributes;
+    }
+
+    /**
+     * Returns an unmodifiable representation of the internal attribute map.
+     * @return immutable map
+     */
+    public Map<String, Value> asUnmodifiableMap() {
+        return Collections.unmodifiableMap(attributes);
```
aepfli, line 30: Is it worth to check if this is an unmodifiable nap, before wrapping it?

aepfli (commented):
I am sorry, I tried reviewing this on my mobile, but I failed half way through ;) I will give it a shot later. There is one comment, which might also reduce the amount of collections again. (Not sure, if this is maybe handled by the function call anyways)

src/main/java/dev/openfeature/sdk/AbstractStructure.java
toddbaert: According to SO, `Collections.unmodifiableMap` already does this.

src/main/java/dev/openfeature/sdk/Structure.java
```diff
@@ -107,41 +113,6 @@ default Object convertValue(Value value) {
         throw new ValueNotConvertableError();
     }
 
-    /**
```
toddbaert, line 110: This removal is very technically a breaking change, but it was never intended to be public and it's very awkward to use. It has more to do with `EvaluationContexts` not `Structures` anyway and I doubt it's widely used.

aepfli (approved):

PR #1228, opened 2024-11-22
feat: implemented tracking as per spec

### LINKED ISSUE #1164: [FEATURE] Implement Tracking in Java SDK
Implement tracking as per spec: https://github.com/open-feature/spec/commit/cd99c3560a68c1810971be19b87ea803fb9b0b3a

Functional requirements: 

- implement all the features described [here](https://github.com/open-feature/spec/commit/cd99c3560a68c1810971be19b87ea803fb9b0b3a)

Non functional requirements:

- use as many existing abstractions and implementations as possible
  - ie: `tracking event details` should likely use similar underlying functionality to evaluation context / structure
- provider interface(s) are modified to _optionally_ support tracking; not all providers will support it
- associated testing and README entry (see markdown snippet below)
- no breaking changes

README addition (be careful to correctly link to #hooks and #providers and update the code snippet for the language in question

```md
### Tracking

The tracking API allows you to use OpenFeature abstractions and objects to associate user actions with feature flag evaluations.
This is essential for robust experimentation powered by feature flags.
For example, a flag enhancing the appearance of a UI component might drive user engagement to a new feature; to test this hypothesis, telemetry collected by a hook(#hooks) or provider(#providers) can be associated with telemetry reported in the client's `track` function.

{example-language-in-question}

Note that some providers may not support tracking; check the documentation for your provider for more information.
```

warber:
## This PR

- adds support for Tracking as per Spec

### Related Issues

Fixes #1164

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,17 +120,18 @@ See [here](https://javadoc.io/doc/dev.openfeature/sdk/latest/) for the Javadocs.
 
 ## 🌟 Features
 
-| Status | Features                                                              | Description                                                                                                                                                   |
-| ------ |-----------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| ✅     | [Providers](#providers)                                               | Integrate with a commercial, open source, or in-house feature management tool.                                                                                |
-| ✅     | [Targeting](#targeting)                                               | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context).                            |
-| ✅     | [Hooks](#hooks)                                                       | Add functionality to various stages of the flag evaluation life-cycle.                                                                                        |
-| ✅     | [Logging](#logging)                                                   | Integrate with popular logging packages.                                                                                                                      |
-| ✅     | [Domains](#domains)                                                   | Logically bind clients with providers.                                                                                                                        |
-| ✅     | [Eventing](#eventing)                                                 | React to state changes in the provider or flag management system.                                                                                             |
-| ✅     | [Shutdown](#shutdown)                                                 | Gracefully clean up a provider during application shutdown.                                                                                                   |
-| ✅     | [Transaction Context Propagation](#transaction-context-propagation)   | Set a specific [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context) for a transaction (e.g. an HTTP request or a thread). |   
-| ✅     | [Extending](#extending)                                               | Extend OpenFeature with custom providers and hooks.                                                                                                           |
+| Status | Features                                                            | Description                                                                                                                                                   |
+| ------ |---------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| ✅     | [Providers](#providers)                                             | Integrate with a commercial, open source, or in-house feature management tool.                                                                                |
+| ✅     | [Targeting](#targeting)                                             | Contextually-aware flag evaluation using [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context).                            |
+| ✅     | [Hooks](#hooks)                                                     | Add functionality to various stages of the flag evaluation life-cycle.                                                                                        |
+| ✅     | [Tracking](#tracking)                                               | Associate user actions with feature flag evaluations.                                                                                                         |
+| ✅     | [Logging](#logging)                                                 | Integrate with popular logging packages.                                                                                                                      |
+| ✅     | [Domains](#domains)                                                 | Logically bind clients with providers.                                                                                                                        |
+| ✅     | [Eventing](#eventing)                                               | React to state changes in the provider or flag management system.                                                                                             |
+| ✅     | [Shutdown](#shutdown)                                               | Gracefully clean up a provider during application shutdown.                                                                                                   |
+| ✅     | [Transaction Context Propagation](#transaction-context-propagation) | Set a specific [evaluation context](https://openfeature.dev/docs/reference/concepts/evaluation-context) for a transaction (e.g. an HTTP request or a thread). |   
+| ✅     | [Extending](#extending)                                             | Extend OpenFeature with custom providers and hooks.                                                                                                           |
 
 <sub>Implemented: ✅ | In-progress: ⚠️ | Not implemented yet: ❌</sub>
 
@@ -213,6 +214,16 @@ Once you've added a hook as a dependency, it can be registered at the global, cl
   // add a hook for this evaluation only
   Boolean retval = client.getBooleanValue(flagKey, false, null,
           FlagEvaluationOptions.builder().hook(new ExampleHook()).build());
+```
+
+### Tracking
+
+The [tracking API](https://openfeature.dev/specification/sections/tracking/) allows you to use OpenFeature abstractions to associate user actions with feature flag evaluations.
+This is essential for robust experimentation powered by feature flags. Note that, unlike methods that handle feature flag evaluations, calling `track(...)` may throw an `IllegalArgumentException` if an empty string is passed as the `trackingEventName`.
+
+```java
+OpenFeatureAPI api = OpenFeatureAPI.getInstance();
+api.getClient().track("visited-promo-page", new MutableTrackingEventDetails(99.77).add("currency", "USD"));
 ```
 
 ### Logging
--- a/src/main/java/dev/openfeature/sdk/FeatureProvider.java
+++ b/src/main/java/dev/openfeature/sdk/FeatureProvider.java
@@ -71,4 +71,14 @@ public interface FeatureProvider {
         return ProviderState.READY;
     }
 
+    /**
+     * Feature provider implementations can opt in for to support Tracking by implementing this method.
+     *
+     * @param eventName The name of the tracking event
+     * @param context   Evaluation context used in flag evaluation (Optional)
+     * @param details   Data pertinent to a particular tracking event (Optional)
+     */
+    default void track(String eventName, EvaluationContext context, TrackingEventDetails details) {
+
+    }
 }
--- a/src/main/java/dev/openfeature/sdk/ImmutableTrackingEventDetails.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableTrackingEventDetails.java
@@ -1,0 +1,53 @@
+package dev.openfeature.sdk;
+
+import dev.openfeature.sdk.internal.ExcludeFromGeneratedCoverageReport;
+import lombok.experimental.Delegate;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+
+/**
+ * ImmutableTrackingEventDetails represents data pertinent to a particular tracking event.
+ */
+public class ImmutableTrackingEventDetails implements TrackingEventDetails {
+
+    @Delegate(excludes = DelegateExclusions.class)
+    private final ImmutableStructure structure;
+
+    private final Number value;
+
+    public ImmutableTrackingEventDetails() {
+        this.value = null;
+        this.structure = new ImmutableStructure();
+    }
+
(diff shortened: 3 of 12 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -1,24 +1,16 @@
...
-import java.util.Map;
-import java.util.List;
-import java.util.function.Consumer;
-
-import dev.openfeature.sdk.exceptions.ExceptionUtils;
-import dev.openfeature.sdk.exceptions.FatalError;
-import dev.openfeature.sdk.exceptions.GeneralError;
-import dev.openfeature.sdk.exceptions.OpenFeatureError;
-import dev.openfeature.sdk.exceptions.ProviderNotReadyError;
+import dev.openfeature.sdk.exceptions.*;
```
aepfli, line 3: please no star imports

aepfli (changes requested):
Thank you the changes look good, but please remove the start imports, and (personal preference) use assertJ for assertions. Plus the comment i wrote regarding the assertCode

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
warber: Thanks! I think the IDE "optimized" this. I need to figure out how to configure IntelliJ to avoid wildcard imports for regular files while still allowing them for test files. This is probably still useful for things like static imports of Mockito assertions in tests...

src/test/java/dev/openfeature/sdk/TrackingSpecTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    @SneakyThrows
+    void trackMethodFulfillsSpec() {
+
+        ImmutableContext ctx = new ImmutableContext();
+        MutableTrackingEventDetails details = new MutableTrackingEventDetails(0.0f);
+        client.track("event");
+        client.track("event", ctx);
+        client.track("event", details);
+        client.track("event", ctx, details);
```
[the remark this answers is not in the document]
warber: I think the intention was to explicitly "test" that the API aligns with the spec design. I wasn’t aware of AssertJ—thanks!

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
aepfli: i started with https://github.com/open-feature/java-sdk/pull/1098 i should finish it, to prevent this in the future

aepfli (approved):

toddbaert:
Will review tomorrow!

lukas-reining:
I will also have a look tomorrow if I find time :)

src/main/java/dev/openfeature/sdk/FeatureProvider.java
```diff
@@ -1,5 +1,6 @@
 package dev.openfeature.sdk;
 
+import javax.annotation.Nullable;
```
toddbaert, line 3: We don't allow usage of javax. We actually removed it from the SDK some time ago because it's very explicitly not allowed at some orgs, but it comes in transitively to this repo. Believe it or not, the `Nullable` annotation is present at runtime too.

src/main/java/dev/openfeature/sdk/MutableTrackingEventDetails.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * MutableTrackingEventDetails represents data pertinent to a particular tracking event.
+ */
+@EqualsAndHashCode
+@ToString
+public class MutableTrackingEventDetails implements TrackingEventDetails {
```
toddbaert, line 17: I think to be consistent with the rest if the SDK we should also implement a `ImutableTrackingEventDetails`.

Not exactly a blocker for me but I think it would be a good idea.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -19,6 +11,16 @@
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
```
toddbaert, line 14: Same as https://github.com/open-feature/java-sdk/pull/1228/files#r1860724481

src/main/java/dev/openfeature/sdk/FeatureProvider.java
aepfli: Out of curiosity are there other classes which should not be used, might be worth adding archUnit tests preventing those imports. But we should define a list for that.

warber: Dropped the @Nullable hint (even though it would've been a nice touch). 
Maybe we could consider creating our own annotation for that?

README.md
```diff
@@ -215,6 +215,16 @@ Once you've added a hook as a dependency, it can be registered at the global, cl
...
 ```
 
+### Tracking
+
+The [tracking API](https://openfeature.dev/specification/sections/tracking/) allows you to use OpenFeature abstractions to associate user actions with feature flag evaluations.
+This is essential for robust experimentation powered by feature flags.
+
+```java
+OpenFeatureAPI api = OpenFeatureAPI.getInstance();
+api.getClient().track("visited-promo-page", new MutableTrackingEventDetails(99.77).add("conurrencyCode", "USD"))
```
justinabrahms, line 225:
```suggestion
api.getClient().track("visited-promo-page", new MutableTrackingEventDetails(99.77).add("currencyCode", "USD"))
```

src/main/java/dev/openfeature/sdk/FeatureProvider.java
toddbaert: > Dropped the @nullable hint (even though it would've been a nice touch). Maybe we could consider creating our own annotation for that?

Not a bad idea, I think. If we could keep it to compile time. But I don't think it would have as much utility as the javax ones since code quality tools wouldn't notice it - still might be helpful in IDE's though.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -183,6 +229,19 @@ private static <T> void enrichDetailsWithErrorDefaults(T defaultValue, FlagEvalu
         details.setReason(Reason.ERROR.toString());
     }
 
+    private static void validateTrackingEventName(String str) {
+        Objects.requireNonNull(str);
+        if (str.isEmpty()) {
+            throw new IllegalArgumentException("trackingEventName cannot be empty");
```
toddbaert, line 235: https://openfeature.dev/specification/sections/flag-evaluation#requirement-1410 says:

> Methods, functions, or operations on the client MUST NOT throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the default value in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.

However, this might be an exception, because:

- it's not a transient error - it's essentially a validation; if this throws it will probably never stop throwing until you fix your code
- it's not an evaluation, and instead a side-effect, and presumably wouldn't break the app totally if it threw, even if you didn't wrap it in a try/catch

toddbaert (approved):
Approving - I'd love to hear from @justinabrahms before merging though.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
warber: I agree for exactly the reasons you mentioned. And I've updated the README.md to include this.

src/test/java/dev/openfeature/sdk/TrackingSpecTest.java
```diff
@@ -0,0 +1,187 @@
...
+    private Client client;
+
+    @BeforeEach
+    void getApiInstance() {
+        api = OpenFeatureAPI.getInstance();
+        client = api.getClient();
+    }
+
+
+    @Specification(number = "6.1.1.1", text = "The `client` MUST define a function for tracking the occurrence of " +
```
chrfwow, line 36:
```suggestion
    @Specification(number = "6.1.1", text = "The `client` MUST define a function for tracking the occurrence of " +
```
I think we do not need the last digit

src/test/java/dev/openfeature/sdk/MutableTrackingEventDetailsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class MutableTrackingEventDetailsTest {
+
+
+    @Test
+    void hasDefaultValue() {
+        MutableTrackingEventDetails track = new MutableTrackingEventDetails();
+        assertEquals(0.0f, track.getValue());
```
chrfwow, line 17: Shouldn't the default value be `null` per the spec?

src/main/java/dev/openfeature/sdk/ImmutableTrackingEventDetails.java
```diff
@@ -0,0 +1,47 @@
...
+public class ImmutableTrackingEventDetails implements TrackingEventDetails {
+
+    @Delegate(excludes = DelegateExclusions.class)
+    private final ImmutableStructure structure;
+
+    @Getter
+    private final Number value;
+
+    public ImmutableTrackingEventDetails() {
+        this.value = 0f;
```
lukas-reining, line 23: I think there `0` is not the right default. 
https://openfeature.dev/specification/sections/tracking#requirement-621 defines this as optional, so I think a numeric 0 is semantically wrong if no value is given. (same counts for `MutableTrackingEventDetails`)

lukas-reining (commented):
Looks good to me!
I only have one concerns about the numeric value.

src/test/java/dev/openfeature/sdk/MutableTrackingEventDetailsTest.java
toddbaert: Ya, as @lukas-reining pointed out, we should consider this point carefully: https://github.com/open-feature/java-sdk/pull/1228#discussion_r1870292488

src/main/java/dev/openfeature/sdk/ImmutableTrackingEventDetails.java
warber: yeah good point, I've changed it to `null` and let the getter return an `Optional<Number>`

toddbaert: That's a good solution. We already use optional and I think this conveys what we want.

toddbaert:
Since this is an SDK change, I'ld like to get a few more approvals before merge.

src/main/java/dev/openfeature/sdk/ImmutableTrackingEventDetails.java
lukas-reining: Nice that is a good solution!

lukas-reining (approved):
Looks great! Thanks!

chrfwow (approved):

src/test/java/dev/openfeature/sdk/TrackingSpecTest.java
toddbaert: Technically this is correct: https://openfeature.dev/specification/sections/tracking/#conditional-requirement-6111

toddbaert:
@warber  Thanks so much! Merging - I added one small change because I was unclear before when I said:

> We might want to add to the docs for .track that it can throw this exception

I meant the Javadocs (though it's fine to put it in the README too). I added this just now: https://github.com/open-feature/java-sdk/pull/1228/commits/48187b57c24a57e947388d1cc37feecc0f1b415f

beeme1mr (approved):

PR #1236, opened 2024-12-05
test: Update gherkin test sub module

warber:
## This PR
the git submodule pointed to a wrong/old repository.
This PR fixes it by changing the git module to point to the OF spec repository where the gherkin test definitions reside.

Additionally, I've moved the java test definitions into their own package with the motivation to have it nicely separated, e.g. when introducing the next test suite.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "test-harness"]
-	path = test-harness
-	url = https://github.com/open-feature/test-harness
+[submodule "spec"]
+	path = spec
+	url = https://github.com/open-feature/spec/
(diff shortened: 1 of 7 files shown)


src/test/java/dev/openfeature/sdk/e2e/evaluation/StepDefinitions.java
```diff
@@ -53,6 +53,7 @@ public class StepDefinitions {
     @SneakyThrows
     @BeforeAll()
     @Given("an openfeature client is registered with cache disabled")
```
aepfli, line 55:
```suggestion
    @Given("an openfeature client is registered with cache disabled")
```

.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "test-harness"]
-	path = test-harness
-	url = https://github.com/open-feature/test-harness
+[submodule "spec"]
```
warber, line 1: i think it's not required to call this spec. Alternatively we could keep calling i "test-harness" (same for path).
no hard feeling there tho. wdyt?

aepfli: i would go with spec, to be more precise, as we are using spec and test-harness in the contrib repos

aepfli (approved):

beeme1mr (approved):

Issue #1299, opened 2025-01-23
Deadlock on Provider init


chrfwow:
In my newest [PR](https://github.com/open-feature/java-sdk-contrib/pull/1169), which replaces a busy wait with a wait/notify construct, I discovered that one can run into a deadlock when emitting events during the initialization.
In more detail, the thread that initializes a `FlagdProvider` starts the initialization of the `flagResolver`, and then waits for an event announcing that the provider is ready.
When this ready event is received in the `FlagdProvider`, it calls `EventProvider#emitProviderReady(…)`, which then deadlocks.
### Steps to Reproduce 
I wrote a very basic provider, which always deadlocks on initialization. It can be found [here](https://github.com/chrfwow/DeadlockProvider).
I made my original discovery in this [PR](https://github.com/open-feature/java-sdk-contrib/pull/1169). It can be triggered by running the e2e tests. Notice that the test will keep printing 
```
[Thread-228] WARN dev.openfeature.contrib.providers.flagd.resolver.process.storage.FlagStore - Failed to convey OK satus, queue is full
[Thread-228] WARN dev.openfeature.contrib.providers.flagd.resolver.process.model.FlagParser - Invalid flag configuration: $.flags.equal-greater-lesser-version-flag: must be valid to one and only one schema, but 0 are valid
[Thread-228] WARN dev.openfeature.contrib.providers.flagd.resolver.process.storage.FlagStore - Failed to convey OK satus, queue is full
[Thread-228] WARN dev.openfeature.contrib.providers.flagd.resolver.process.model.FlagParser - Invalid flag configuration: $.flags.equal-greater-lesser-version-flag: must be valid to one and only one schema, but 0 are valid
```
to the console. The warning of the invalid flag config can be ignored, this is expected in the test. The warning `Failed to convey OK satus, queue is full`, however, is an indicator of the problem, as the thread that is supposed to drain this queue is deadlocked.

### Expected Behavior
The `EventProvider` should never deadlock, no matter under which conditions the `emit` (or `emitProviderReady`) method is called. When I wrap the call to the `emitProviderReady()`method inside the `onReady()` method in the `FlagdProvider` in a new thread, the deadlock is prevented.
### Possible Fix
`EventProvider#emit(…)` should not directly call `this.onEmit.accept(this, event, details);`, instead it should store the call in some thread safe queue, which is drained by another thread.

toddbaert:
[...]

PR #1314, opened 2025-02-03
WIP: Move event emitting off the main thread to avoid deadlocks

sideshowcoder:
All code based on work by @chrfwow, based on https://github.com/open-feature/java-sdk/issues/1299.

My current understanding is that when stacking event emitting inside an EventProvider, when using sychronization the EventProvider can deadlock, to avoid this move the event emitting of the main thread. Similar to how this is handled in EventSupport.

Still requires cleanup and moving the tests to a better location, I'm still not sure I 100% understand the reason for the deadlock, but wanted to get the work started by porting the testcase provided by @chrfwow.


src/test/java/dev/openfeature/sdk/EventProviderTest.java
```diff
@@ -75,6 +80,110 @@ void doesNotThrowWhenOnEmitSame() {
         eventProvider.attach(onEmit2); // should not throw, same instance. noop
     }
 
+    @Test
+    @SneakyThrows
+    @Timeout(value = 2, threadMode = Timeout.ThreadMode.SEPARATE_THREAD)
+    @DisplayName("should not deadlock on emit called during emit")
+    void doesNotDeadlockOnEmitStackedCalls() {
```
chrfwow, line 87: Just to be sure, did you verify that this test fails on the current main branch?

chrfwow (changes requested):
Looks good so far, but I have a few suggestions

src/test/java/dev/openfeature/sdk/EventProviderTest.java
sideshowcoder: Yes I did verify iit failing, also tried to understand what is the underlying issue with it before fixing.

aepfli (approved):

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -14,8 +18,15 @@
  *
  * @see FeatureProvider
  */
+@Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
+    private static final int SHUTDOWN_TIMEOUT_SECONDS = 3;
+    private final ExecutorService emitterExecutor = Executors.newCachedThreadPool(runnable -> {
```
toddbaert, line 25: @chrfwow @aepfli @sideshowcoder we could consider making this threadpool static. I'm not 100% sure it's necessary to have a threadpool for each instance of an `EventProvider`. In most cases, there will probably only be one instance, but in the case there's multiple, there could be quite a few and I think it's not worth the overhead.

What do you folks think?

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -14,8 +18,15 @@
  *
  * @see FeatureProvider
  */
+@Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
+    private static final int SHUTDOWN_TIMEOUT_SECONDS = 3;
```
toddbaert, line 24: There's a `SHUTDOWN_TIMEOUT_SECONDS` var already in the `EventSupport` class. Maybe you can use that?

src/main/java/dev/openfeature/sdk/EventProvider.java
sideshowcoder: Makes sense to me, I was actually thinking about making it static, but then followed the pattern from the `EventSupport` where the pool is per instance. But to me a static pool makes more sense.

src/main/java/dev/openfeature/sdk/EventProvider.java
sideshowcoder: That is where I got it from, but didn't want to couple the 2 classes together in this case, can do but since they are different Thread Pools, executing different kind of tasks, it feels like coincidental duplication but not actually the "same thing"

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: TBH the pool in the `EventSupport` could probably be static too... but it's a singleton so I'm not sure it matters much. Maybe you can do that too if there's no downsides anyone can think of.

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: You're right that it's not really the "same thing", but I'm not sure there's much value in there values ever deviating from each-other. :thinking:

src/test/java/dev/openfeature/sdk/EventProviderTest.java
```diff
@@ -75,6 +80,110 @@ void doesNotThrowWhenOnEmitSame() {
...
+        StackedEmitCallsProvider provider = new StackedEmitCallsProvider();
+        OpenFeatureAPI.getInstance().setProviderAndWait(provider);
+    }
+
+    static class StackedEmitCallsProvider extends EventProvider {
```
sideshowcoder, line 92: @toddbaert @aepfli I'm not super happy with the location of the test and related code as it seems to bload this test class quite a bit, my intuition would be to move it out, but I don't see a good location, I was considering to move the support code to `fixtures` but not sure, any suggestion?

toddbaert: There's a "testutils" folder which has some utilities as well as some implementations (TestEventsProvider.java); you could put it there.

src/main/java/dev/openfeature/sdk/EventProvider.java
sideshowcoder: There is actually a problem with making them static, as in this case they can't be shut down when the `shutdown` is called as this is per instance and instead would need to be managed in a global shutdown, likely `OpenFeatureAPI.shutdown()` but since `shutdown` there is more like a reset, creating a new `ProviderRepository` and `EventSupport` I'm not sure this works either. 

So the only way to keep the threadpool from just crashing at the end of the Java Process would be to register a global handler as far as I can see not sure this makes sense. What do you think?

toddbaert: Yes, I see.

I think maybe it's best to keep them instance members then. I think otherwise the per-instance shutdown impacting a global pool will be hard to manage properly and more trouble than it's worth. In any case I'm glad we considered it.

cc @aepfli @chrfwow feel free to disagree.

liran-p: I would prefer to not make it static, having them per provider make it more fair, in case of some provider having more resources etc' and less dependency.

toddbaert:
This all makes sense to me. I see 2 test failures which I'm not really surprised about. It's possible that there's actually bugs here, but more likely it's inter-test interference or test-only race conditions revealed by this change. You can try running the failing tests individually and seeing if they work as expected.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -14,8 +18,14 @@
  *
  * @see FeatureProvider
  */
+@Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
+    private final ExecutorService emitterExecutor = Executors.newCachedThreadPool(runnable -> {
+        final Thread thread = new Thread(runnable);
+        thread.setDaemon(true);
```
liran-p, line 26: daemon can have it stopped before emitting the remaining events and lose data.
The shutdown() method which is later called can gracefully shut it down with a timeout.

src/main/java/dev/openfeature/sdk/EventProvider.java
liran2000: I would prefer to not make it static, having them per provider make it more fair, in case of some provider having more resources etc' and less dependency.

src/main/java/dev/openfeature/sdk/EventProvider.java
liran2000: daemon can have it stopped before emitting the remaining events and lose data.
The shutdown() method which is later called can gracefully shut it down with a timeout.

src/main/java/dev/openfeature/sdk/EventProvider.java
liran2000: Maybe also expose this pool to config, so user can configure a different thread pool?
E.g. If wanting to limit max threads.

toddbaert: I think we can add that later if needed.

beeme1mr (approved):

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -14,8 +18,14 @@
  *
  * @see FeatureProvider
  */
+@Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
+    private final ExecutorService emitterExecutor = Executors.newCachedThreadPool(runnable -> {
```
chrfwow, line 24: I think this can be simplified now that the thread is no longer modified

chrfwow (commented):
Your changes look good, just make sure to fix the failing tests then this PR can be merged I believe 🚀

sideshowcoder:
I am looking at the test failures locally but the seem to be passing, when run on their own, just wanted to quickly update that I'm looking at this 👍

pom.xml
```diff
@@ -265,6 +265,8 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>3.5.2</version>
                 <configuration>
+                    <forkCount>1</forkCount>
+                    <reuseForks>false</reuseForks>
```
sideshowcoder, line 269: Ok this is obviously not a solution, but I want to know what the reason for the test failures are, are they actual issues ore is it cross test interactions? It seems we are suffering from global setup issues due to the Singleton that is OpenFeatureApi. @toddbaert any suggestion how to get around this?

aepfli: Which tests are failing? Today we merged something which increases the stability of the spectests

toddbaert: I think there's a test annotation to run some tests in isolation, which may be the best solution here. If you cannot narrow down the exact suite I'd support you adding forkCount 1 for now, and I can look into it later (or somebody else).

If you want to keep digging at it yourself, you're more than welcome, but it's not 100% related to your improvement, so I don't want to force you to do it if you are sure it's a bad test interaction.

Up to you!

liran2000: @ Execution(SAME_THREAD) annotation

sideshowcoder: Ah thanks! Yes it is always the same tests so I'm gonna annotate them and remove this global one. And then this should be done.

chrfwow (approved):
The lines not covered by tests are fine for me

pom.xml
sideshowcoder: Ok so I tried this out locally and got some fixes but still have some failing tests, with annotations it is not possible to get the same isolation as with no reuse for forks. I am gonna spent more time tomorrow on this and if I can't resolve it I would suggest to move this to a followup MR?

beeme1mr: I'm in favor of merging as is and following up with improvements in a separate PR. @toddbaert @liran2000 any concerns about that?

toddbaert: No objections.

toddbaert (approved):

toddbaert:
@sideshowcoder thanks for all your efforts on this! Well done. It will be available in our next release.

Consider joining the org: https://github.com/open-feature/community/pull/427

PR #1328, opened 2025-02-13
fix: TrackingEventDetails interface to include numeric getValue() call

jonathannorris:
## This PR
- fix: `TrackingEventDetails` interface to include numeric `Optional<Number> getValue();` function interface. Only the `Value getValue(String key);` was being exposed by extending `Structure`.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/TrackingEventDetails.java
+++ b/src/main/java/dev/openfeature/sdk/TrackingEventDetails.java
@@ -1,6 +1,14 @@
 package dev.openfeature.sdk;
 
+import java.util.Optional;
+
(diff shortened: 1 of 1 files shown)


aepfli (approved):
thank you, this change should make the live for a lot of people a lot easier and this interface method is definitely rounding up the experience. Thank you!!

beeme1mr (approved):

toddbaert (approved):

toddbaert:
Oh yes this is a nice improvement, thanks.

We are still working on getting tracking implemented internally at our org, but I'm sure we would have run into this soon.

PR #1331, opened 2025-02-15
test: Reduce usage of singelton within our tests and implementations

aepfli:
Our tests are great, but often we rely on our own Singleton for testing purposes. This can create concurrency issues or make testing really hard. By instantiating a own API object for each test we ensure that we are not messing with each other.
Furthermore we should not use `.getInstance()` within our own implementation, which is now ensured with an ArchUnit test


src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
```diff
@@ -414,8 +404,9 @@ void broken_provider() {
             text =
                     "In cases of abnormal execution, the `evaluation details` structure's `error message` field **MAY** contain a string containing additional details about the nature of the error.")
     @Test
-    void broken_provider_withDetails() {
-        FeatureProviderTestUtils.setFeatureProvider(new AlwaysBrokenWithDetailsProvider());
+    void broken_provider_withDetails() throws InterruptedException {
+        api.setProvider(new AlwaysBrokenWithDetailsProvider());
+        Thread.sleep(50);
```
justinabrahms, line 409: Why the sleep?

justinabrahms (commented):
I quite like this! The one concern I have would be people not using the singleton causing issues.

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
aepfli: I need to investigate this/optimize the test. It was flaky because I either got a flag not found or a provider not ready error, and with `setProviderAndWait` it dit not initialize properly

aepfli:
> I quite like this! The one concern I have would be people not using the singleton causing issues.

The good part is, the constructor is not public but protected, and always have been. Hence either people need to extend the API object, or have to be in the same package. This is already possible, so we are actually not changing anything regarding this behavior

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
```diff
@@ -23,7 +22,7 @@ class DeveloperExperienceTest implements HookFixtures {
 
     @Test
     void simpleBooleanFlag() {
-        OpenFeatureAPI api = OpenFeatureAPI.getInstance();
+        OpenFeatureAPI api = new OpenFeatureAPI();
```
chrfwow, line 25: Does it make sense to move this into a `@BeforeEach` method?

src/test/java/dev/openfeature/sdk/OpenFeatureAPITestUtil.java
```diff
@@ -0,0 +1,11 @@
+package dev.openfeature.sdk;
+
+public class OpenFeatureAPITestUtil {
+
+    private OpenFeatureAPITestUtil() {}
+
+    public static OpenFeatureAPI createAPI() {
+        return new OpenFeatureAPI();
+    }
+    ;
```
chrfwow, line 10: Leftover `;`?

src/test/java/dev/openfeature/sdk/DeveloperExperienceTest.java
aepfli: yes it does make sense, i did a simply replace for certain strings, and did not explore the tests more into details ;) - might not be the only test which could be improved with structure, i will take a look at it, as it seems to me like another string replace :D

src/test/java/dev/openfeature/sdk/OpenFeatureAPITestUtil.java
aepfli:
```suggestion
```

src/test/java/dev/openfeature/sdk/FlagEvaluationSpecTest.java
aepfli: found the culprit and removed it

chrfwow (approved):

toddbaert (approved):

Issue #1362, opened 2025-03-06
MutableContext / MutableStructure equality bug


glissand0:
MutableContext stores a MutableStructure instance, which uses Lombok's `@EqualsAndHashCode`. Using IntelliJ's delombok, this is the generated code

```
public boolean equals(final Object o) {
      if (o == this) {
          return true;
      }
      if (!(o instanceof MutableStructure)) {
          return false;
      }
      final MutableStructure other = (MutableStructure) o;
      if (!other.canEqual((Object) this)) {
          return false;
      }
      return true;
  }

  protected boolean canEqual(final Object other) {return other instanceof MutableStructure;}

  public int hashCode() {
      int result = 1;
      return result;
  }
```

 ie, all MutableStructure objects will be treated as equal and the below tests pass
```
    @Test
    void mutableStructureEquality() {
        MutableStructure m1 = new MutableStructure();
        m1.add("key1", "val1");
        MutableStructure m2 = new MutableStructure();
        m1.add("key2", "val2");
        assertEquals(m1, m2);
    }

    @Test
    void mutableCtxEquality() {
        final Map<String, Value> attributes = new HashMap<>();
        attributes.put("key1", new Value("val1"));
        final MutableContext ctx = new MutableContext(attributes);

        final Map<String, Value> attributes2 = new HashMap<>();
        final MutableContext ctx2 = new MutableContext(attributes2);

        assertThat(ctx).isEqualTo(ctx2);
    }
```
This is Lombok's behaviour when a class does not contain any attributes, regardless of what is in the parent class (see, eg https://medium.com/vena-engineering/the-inheritance-hashset-related-bug-with-lombok-36dbcfb04381)

This can be fixed by changing the equality annotation on MutableStructure to `@EqualsAndHashCode(callSuper = true)`

On a related note, ImmutableContext does not override `equals` at all. This could be the intention, but I think it would be useful if it did?

PR #1364, opened 2025-03-06
fix: equals and hashcode of several classes

chrfwow:
## This PR
Fixes some of the `equals` and `hashcode` methods in the sdk.

### Related Issues

Fixes [#1362](https://github.com/open-feature/java-sdk/issues/1362)

### Notes
I assumed that for the classes I modified, two objects should be equal iff their underlying data is equal. E.g. the class `Value`. Two distinct `Value` objects should be equal if they represent the same data:
```java
        Value val1 = new Value(12312312);
        Value val2 = new Value(12312312);
        assertEquals(val1, val2);
```
Please comment if you think there are more such cases in the sdk, or if I added this behavior somewhere it is not appropriate.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/AbstractStructure.java
+++ b/src/main/java/dev/openfeature/sdk/AbstractStructure.java
@@ -3,8 +3,10 @@ package dev.openfeature.sdk;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import lombok.EqualsAndHashCode;
 
 @SuppressWarnings({"PMD.BeanMembersShouldSerialize", "checkstyle:MissingJavadocType"})
+@EqualsAndHashCode
 abstract class AbstractStructure implements Structure {
 
     protected final Map<String, Value> attributes;
@@ -45,47 +47,5 @@ abstract class AbstractStructure implements Structure {
                         HashMap::new,
                         (accumulated, entry) -> accumulated.put(entry.getKey(), convertValue(entry.getValue())),
                         HashMap::putAll);
-    }
-
-    public boolean equals(final Object o) {
-        if (o == this) {
-            return true;
-        }
-        if (!(o instanceof AbstractStructure)) {
-            return false;
-        }
-        final AbstractStructure other = (AbstractStructure) o;
-        if (other.attributes == attributes) {
-            return true;
-        }
-        if (attributes == null || other.attributes == null) {
-            return false;
-        }
-        if (other.attributes.size() != attributes.size()) {
-            return false;
-        }
-
-        for (Map.Entry<String, Value> thisEntry : attributes.entrySet()) {
-            Value thisValue = thisEntry.getValue();
-            Value otherValue = other.attributes.get(thisEntry.getKey());
-            if (thisValue == null && otherValue == null) {
-                continue;
-            }
-            if (thisValue == null || otherValue == null) {
-                return false;
-            }
-            if (!thisValue.equals(otherValue)) {
-                return false;
-            }
-        }
-
-        return true;
-    }
-
-    public int hashCode() {
-        if (attributes == null) {
-            return 0;
-        }
-        return attributes.hashCode();
     }
 }
--- a/src/main/java/dev/openfeature/sdk/AbstractStructure.java
+++ b/src/main/java/dev/openfeature/sdk/AbstractStructure.java
@@ -11,7 +11,7 @@ abstract class AbstractStructure implements Structure {
 
     @Override
     public boolean isEmpty() {
-        return attributes == null || attributes.size() == 0;
+        return attributes == null || attributes.isEmpty();
     }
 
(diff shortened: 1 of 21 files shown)


src/main/java/dev/openfeature/sdk/AbstractStructure.java
```diff
@@ -46,4 +46,46 @@ public Map<String, Object> asObjectMap() {
                         (accumulated, entry) -> accumulated.put(entry.getKey(), convertValue(entry.getValue())),
                         HashMap::putAll);
     }
+
+    public boolean equals(final Object o) {
```
liran2000, line 50: why implement it instead of adding EqualsAndHashCode and adding inheritors with EqualsAndHashCode**(callSuper = true)** ?
Then it seems to call it on attributes map entries.
similar for others.

src/main/java/dev/openfeature/sdk/Value.java
```diff
@@ -316,4 +314,31 @@ public static Value objectToValue(Object object) {
             throw new TypeMismatchError("Flag value " + object + " had unexpected type " + object.getClass() + ".");
         }
     }
+
+    /**
+     * Returns true iff {@code this} is equal to {@code o}, or if both objects represent the same data.
+     * @param o the other object
+     * @return true iff both objects are equal or represent the same data
+     */
+    public boolean equals(final Object o) {
```
liran2000, line 323: Here too, similar implementation to the _EqualsAndHashCode_ by delombok.

src/main/java/dev/openfeature/sdk/AbstractStructure.java
chrfwow: Lombok's implementation has a bit more overhead, and as we can see from the issue, it is not always an implementation one would expect

liran2000: I read the issue. By solution of this, it may be suggested that Lombok's implementation is not useful here?
If their implementation works, widely used, and makes the methods redunadant, why not use it?
One thing I see misleading is that callSuper value ia false. But if you add the value as true, isn't it enough to make all the tests pass without explicit implementations?

src/test/java/dev/openfeature/sdk/ImmutableStructureTest.java
```diff
@@ -154,4 +159,74 @@ void constructorHandlesNullValue() {
...
+        Map<String, Value> attrs2 = new HashMap<>();
+        attrs2.put("test", new Value(45));
+        ImmutableStructure structure2 = new ImmutableStructure(attrs2);
+
+        assertEquals(structure1, structure2);
+    }
+
+    @Test
+    void unequalMutableStructuresAreNotEqual() {
+        MutableStructure m1 = new MutableStructure();
```
glissand0, line 191: minor comment: the MutableStructure tests are in `ImmutableStructureTest`. Should they be moved into a separate class?

src/main/java/dev/openfeature/sdk/AbstractStructure.java
glissand0: Agree with @liran2000 here, if we can configure lombok to work "as expected" then there is less code to maintain :)

glissand0 (commented):
Thanks for addressing the issue so promptly, appreciate it! 👍

glissand0:
Perhaps in a separate PR could we address the `toString` generation as well? Eg, for MutableStructure it is 
```
public String toString() {return "MutableStructure()";}
```
@ToString(callSuper = true) generates what we want

src/main/java/dev/openfeature/sdk/Value.java
chrfwow: Do we want a `Value` object to have the same hashCode as its `innerObject`?

src/main/java/dev/openfeature/sdk/AbstractStructure.java
chrfwow: Ok. Do we want different inheritors of `AbstractStructure` and classes that delegate to it to equal each other if they have the same underlying `AbstractStructure`?

glissand0: I don't think they should be equal, else what's the point of the different classes?

chrfwow: I would say it's fair to consider a `MutableContext` equal to an `ImmutableContext` if they represent the same data.

I changed the code to use `@EqualsAndHashCode(callSuper = true)` where applicable

src/main/java/dev/openfeature/sdk/Value.java
liran2000: I think it's fine to leave it as default Lombok implementation, which is kind of what it is.

liran2000 (approved):

glissand0 (approved):

PR #1392, opened 2025-03-26
fix: hooks not run in NOT_READY/FATAL

toddbaert:
Fixes a bug where our "short circuit" logic was "short circuiting" too early, leading to various errors in hooks.

Fixes: https://github.com/open-feature/java-sdk/issues/1370

### DIFF
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
@@ -178,12 +178,6 @@ public class OpenFeatureClient implements Client {
             // provider must be accessed once to maintain a consistent reference
             provider = stateManager.getProvider();
             ProviderState state = stateManager.getState();
-            if (ProviderState.NOT_READY.equals(state)) {
-                throw new ProviderNotReadyError("provider not yet initialized");
-            }
-            if (ProviderState.FATAL.equals(state)) {
-                throw new FatalError("provider is in an irrecoverable error state");
-            }
 
             mergedHooks = ObjectUtils.merge(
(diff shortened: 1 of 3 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -178,12 +178,6 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
             // provider must be accessed once to maintain a consistent reference
             provider = stateManager.getProvider();
             ProviderState state = stateManager.getState();
-            if (ProviderState.NOT_READY.equals(state)) {
-                throw new ProviderNotReadyError("provider not yet initialized");
-            }
-            if (ProviderState.FATAL.equals(state)) {
-                throw new FatalError("provider is in an irrecoverable error state");
-            }
```
toddbaert, line 186: Throwing these before we ran the hooks and setup the `afterHookContext` was the cause of the issue. We just needed to move it later.

src/test/java/dev/openfeature/sdk/HookSpecTest.java
```diff
@@ -594,6 +594,25 @@ void erroneous_flagResolution_setsAppropriateFieldsInFlagEvaluationDetails() {
         assertThat(evaluationDetails.getValue()).isTrue();
     }
 
+    @Test
+    void shortCircuit_flagResolution_runsHooksWithAllFields() {
```
toddbaert, line 598: This test fails without the fix.

aepfli (approved):

PR #1444, opened 2025-05-13
feat: add logging on provider state transitions

chrfwow:
## This PR
Adds logging on state transitions. This way errors and their recovery can be retraced better.

### Related Issues
None


src/main/java/dev/openfeature/sdk/FeatureProviderStateManager.java
```diff
@@ -24,43 +24,60 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
...
+    private void setState(ProviderState state) {
+        ProviderState oldState = this.state.getAndSet(state);
+        if (oldState != state) {
+            String providerName;
+            if (delegate.getMetadata() == null || delegate.getMetadata().getName() == null) {
+                providerName = "unknown";
+            } else {
+                providerName = delegate.getMetadata().getName();
+            }
+            log.info("Provider {} transitioned from state {} to state {}", providerName, oldState, state);
```
beeme1mr, line 73: I think this should be a debug log.

toddbaert: I understand why you'd say so, and I think I would agree most times, but the point of this change was to get better visibly into state transitions, particularly for when providers become ready again. Because this isn't in a hot code path, I think it might be worthwhile to have. We've seen internally that such a log would be very useful to confirm the provider is back in a good state.

If we make it debug, we'll never see it "after the fact" if logging it set to normal level, so I think it will make troubleshooting issues harder. I think the chances of spam here are very low in normal situations.

aepfli: I am not sure, we need this log all the time. imho I think it is important when we are going from error back to ready - but the other transitions are not as important to see. Stale can happen, we have a gracePeriod, I do not need to know it. but after an error, I need to know when got back to ready

aepfli (approved):

PR #1463, opened 2025-05-28
feat: add means of awaiting event emission, fix flaky build

### LINKED ISSUE #1449: Flaky build due to a possible race condition
Recently, the main build failed. An affected tests (`EventsTest.matchingErrorEventsMustRunImmediately`) failed because it assumes that an error event handler that is added to a provider in an error state will be called immediately and exactly once. However, this handler was called twice.

On researching that, it turned out that the test fails only under certain circumstances. In my case this was when I executed the whole test suite; running only the test mentioned did not result in an error.

I believe the cause to be a race condition in the following parts of the sdk:

```java
final TriConsumer<EventProvider, ProviderEvent, ProviderEventDetails> localOnEmit = this.onEmit;
if (localOnEmit != null) {
    emitterExecutor.submit(() -> localOnEmit.accept(this, event, details));
}
```
This code snippet shows parts of how the `EventProvider` emits events. Note that the `localOnEmit` is called only inside a lambda that is executed by the `emitterExecutor`. When this is changed to be invoked to on the current thread, the test does not fail.

The failing test adds a new error event listener to a client with a provider already in an error state.
I believe that in the time the `emitterExecuter` takes to start executing the lambda, the handler is added to the collection of handlers, and is therefore executed a second time by the `emitterExecutor`. This might be why the test does not fail when instead the current thread calls `localOnEmit.accept(this, event, details)`. 
Just moving the execution of this call to the current thread might not be enough to fix this issue, however, as we do not know the interleaving of threads in a general case.
Instead, it might be required to guard additions of new handlers and running of existing ones with a lock.

chrfwow:
## This PR

- adds the ability to await event emissions by returning a new construct
- uses this construct in tests

:warning: in rare cases, this could be a breaking change, but the events API is currently experimental, so we will not do a major version ubmp

### Related Issues

Fixes #1449

### DIFF
--- a/src/main/java/dev/openfeature/sdk/Awaitable.java
+++ b/src/main/java/dev/openfeature/sdk/Awaitable.java
@@ -4,6 +4,10 @@ package dev.openfeature.sdk;
  * A class to help with synchronization by allowing the optional awaiting of the associated action.
  */
 public class Awaitable {
+
+    /**
+     * An already-completed Awaitable. Awaiting this will return immediately.
+     */
     public static final Awaitable FINISHED = new Awaitable(true);
 
     private boolean isDone = false;


src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -76,14 +76,21 @@ public void shutdown() {
...
+        final var localOnEmit = this.onEmit;
+
+        if (localEventProviderListener == null && localOnEmit == null) {
+            return;
         }
 
-        final TriConsumer<EventProvider, ProviderEvent, ProviderEventDetails> localOnEmit = this.onEmit;
-        if (localOnEmit != null) {
-            emitterExecutor.submit(() -> localOnEmit.accept(this, event, details));
+        try (var ignored = OpenFeatureAPI.lock.readLockAutoCloseable()) {
```
chrfwow, line 87: It's not nice to access the static lock that's clearly only exposed for testing, but we need to use this very lock to protect against concurrent reads/writes to the underlying map.

I removed the call to the `emitterExecutor`, so now all listeners are invoked sequentially on the calling thread. This was needed as several tests expect the state of the provider to be set immediately after calling this method.
If we consider this a problem, I could work around this issue by returning something our tests can wait for. In this case, we would need to call both the `localOnEmit` and the `localEventProviderListener` in the `emitterExecutor`.

aepfli (approved):
Thank you! I really would love to have the same detail and knowledge as you have for concurrency. Thank you!

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -21,7 +21,6 @@
 @Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
-    private final ExecutorService emitterExecutor = Executors.newCachedThreadPool();
```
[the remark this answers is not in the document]
chrfwow: Wrong, we do need it

src/main/java/dev/openfeature/sdk/EventProvider.java
chrfwow: By moving the call from the `emitterExecutor` to the current thread, I reintroduced a race condition that was already fixed. 
Therefore I added the executor again. This means that some tests will fail that expect the state of the provider to be updated immediately after a call to `emit`. This is why I now return an `Awaitable`, for which those tests (or users) can wait.

chrfwow:
@toddbaert in theory, this is a breaking change. The state of the provider is not set synchronously when calling the `emit()` method, but only after the `emitterExecutor` runs. Users can wait for the state change to be propagated by calling `await()` on the return value of `emit()`.

aepfli (approved):

src/main/java/dev/openfeature/sdk/Awaitable.java
```diff
@@ -0,0 +1,40 @@
+package dev.openfeature.sdk;
+
+/**
+ * A class to help with synchronization.
+ */
+public class Awaitable {
```
toddbaert, line 6: I like this.

src/main/java/dev/openfeature/sdk/EventSupport.java
```diff
@@ -160,14 +156,14 @@ public void shutdown() {
...
-        private final Map<ProviderEvent, List<Consumer<EventDetails>>> handlerMap;
+        private final Map<ProviderEvent, Collection<Consumer<EventDetails>>> handlerMap;
 
         HandlerStore() {
             handlerMap = new ConcurrentHashMap<>();
-            handlerMap.put(ProviderEvent.PROVIDER_READY, new ArrayList<>());
-            handlerMap.put(ProviderEvent.PROVIDER_CONFIGURATION_CHANGED, new ArrayList<>());
-            handlerMap.put(ProviderEvent.PROVIDER_ERROR, new ArrayList<>());
-            handlerMap.put(ProviderEvent.PROVIDER_STALE, new ArrayList<>());
+            handlerMap.put(ProviderEvent.PROVIDER_READY, new ConcurrentLinkedQueue<>());
```
toddbaert, line 163: Good idea.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -93,8 +110,8 @@ public void emit(ProviderEvent event, ProviderEventDetails details) {
      *
      * @param details The details of the event
      */
-    public void emitProviderReady(ProviderEventDetails details) {
-        emit(ProviderEvent.PROVIDER_READY, details);
+    public Awaitable emitProviderReady(ProviderEventDetails details) {
```
toddbaert, line 113: This is a public class people can override to implement a provider that supports events. I like this API and I think it's handy to have.

I was trying to see if there was _any_ way this could be considered breaking. I tried adding this at the top of the class:

`private BiConsumer<ProviderEvent, ProviderEventDetails> doesThisCompile = this::emit;`

And (surprisingly to me) before and after your change, this still compiles even though `BIcConsumer` returns `void`. Apparently, return values of method references are ignored when assigned to void functional interfaces!

EDIT:

Oh - I just realized, this could break implementations that override these methods... I think that would be relatively rare but it's possible. What do you think @aepfli @chrfwow ?

chrfwow: I don't think this would be much of a problem. Providers can just return `Awaitable.FINISHED` if they do this synchronously, otherwise it's probably best to return something to get informed about the state update anyway.

src/main/java/dev/openfeature/sdk/Awaitable.java
```diff
@@ -0,0 +1,40 @@
+package dev.openfeature.sdk;
+
+/**
+ * A class to help with synchronization.
```
toddbaert:
```suggestion
 * A class to help with synchronization by allowing the optional awaiting of the associated action.
```

src/main/java/dev/openfeature/sdk/Awaitable.java
```diff
@@ -0,0 +1,40 @@
+package dev.openfeature.sdk;
+
+/**
+ * A class to help with synchronization by allowing the optional awaiting of the associated action.
+ */
+public class Awaitable {
+    public static final Awaitable FINISHED = new Awaitable(true);
```
toddbaert, line 11:
```suggestion

    /**
     * An already-completed Awaitable. Awaiting this will return immediately.
     */
    public static final Awaitable FINISHED = new Awaitable(true);
```

PR #1478, opened 2025-06-12
fix: Reduce locking and concurrency issues

chrfwow:
## This PR
Reduces locking and potential concurrency issues

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
     - uses: actions/checkout@09d2acae674a48949e3602304ab46fd20ae0c42f
     - name: Set up JDK 17
-      uses: actions/setup-java@ebb356cc4e59bcf94f518203228485f5d40e4b58
+      uses: actions/setup-java@67aec007b3fcabe15ca665bfccc1e255dd52e30d
       with:
         java-version: '17'
         distribution: 'temurin'
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -22,14 +22,14 @@ jobs:
         uses: actions/checkout@09d2acae674a48949e3602304ab46fd20ae0c42f
 
       - name: Set up JDK 11
-        uses: actions/setup-java@ebb356cc4e59bcf94f518203228485f5d40e4b58
+        uses: actions/setup-java@67aec007b3fcabe15ca665bfccc1e255dd52e30d
         with:
             java-version: ${{ matrix.build.java }}
             distribution: 'temurin'
             cache: maven
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@4c57370d0304fbff638216539f81d9163f77712a
+        uses: github/codeql-action/init@b69421388d5449cc5a5e1ca344d71926bda69e07
         with:
           languages: java
 
@@ -55,4 +55,4 @@ jobs:
           verbose: true # optional (default = false)
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@4c57370d0304fbff638216539f81d9163f77712a
+        uses: github/codeql-action/analyze@b69421388d5449cc5a5e1ca344d71926bda69e07
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@09d2acae674a48949e3602304ab46fd20ae0c42f
 
       - name: Set up JDK 17
-        uses: actions/setup-java@ebb356cc4e59bcf94f518203228485f5d40e4b58
+        uses: actions/setup-java@67aec007b3fcabe15ca665bfccc1e255dd52e30d
         with:
           java-version: '17'
           distribution: 'temurin'
--- a/.github/workflows/static-code-scanning.yaml
+++ b/.github/workflows/static-code-scanning.yaml
@@ -33,12 +33,12 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@4c57370d0304fbff638216539f81d9163f77712a
+        uses: github/codeql-action/init@b69421388d5449cc5a5e1ca344d71926bda69e07
         with:
           languages: java
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@4c57370d0304fbff638216539f81d9163f77712a
+        uses: github/codeql-action/autobuild@b69421388d5449cc5a5e1ca344d71926bda69e07
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@4c57370d0304fbff638216539f81d9163f77712a
+        uses: github/codeql-action/analyze@b69421388d5449cc5a5e1ca344d71926bda69e07
--- a/pom.xml
+++ b/pom.xml
@@ -571,7 +571,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-gpg-plugin</artifactId>
-                        <version>3.2.7</version>
+                        <version>3.2.8</version>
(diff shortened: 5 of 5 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -24,7 +26,7 @@ public class OpenFeatureAPI implements EventBus<OpenFeatureAPI> {
     private final List<Hook> apiHooks;
```
chrfwow: This `ArrayList` could also be replaced with some concurrent data structure, further reducing the usage of locks and object churn, but it might be slower in some cases

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -396,18 +396,20 @@ public OpenFeatureAPI on(ProviderEvent event, Consumer<EventDetails> handler) {
      */
     @Override
     public OpenFeatureAPI removeHandler(ProviderEvent event, Consumer<EventDetails> handler) {
-        this.eventSupport.removeGlobalHandler(event, handler);
+        try (AutoCloseableLock ignored = lock.writeLockAutoCloseable()) {
```
chrfwow, line 399: It might be possible to remove locking on the `eventSupport` entirely, now that it uses concurrent collections.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -49,8 +50,7 @@ public class OpenFeatureClient implements Client {
     private final List<Hook> clientHooks;
```
chrfwow: This `ArrayList` could also be replaced with some concurrent data structure, further reducing the usage of locks

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -216,7 +214,7 @@ public void setProvider(String domain, FeatureProvider provider) {
      * @throws OpenFeatureError if the provider fails during initialization.
      */
     public void setProviderAndWait(FeatureProvider provider) throws OpenFeatureError {
-        try (AutoCloseableLock __ = lock.writeLockAutoCloseable()) {
```
toddbaert, line 219: I think at some point checkstyle or sonar forced this, but our settings must have changed.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
toddbaert: Ya it's final unlike the `evaluationContext` is, so we wouldn't need to even use an atomic reference. 

Do you see any reason not to?

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -136,8 +137,12 @@ public OpenFeatureClient addHooks(Hook... hooks) {
      */
     @Override
     public List<Hook> getHooks() {
-        try (AutoCloseableLock __ = this.hooksLock.readLockAutoCloseable()) {
-            return this.clientHooks;
+        try (AutoCloseableLock ignored = this.hooksLock.readLockAutoCloseable()) {
+            if (this.clientHooks.isEmpty()) {
+                return Collections.emptyList();
+            } else {
+                return new ArrayList<>(this.clientHooks);
```
toddbaert: This will create some additional object churn per evaluation, I guess you're worried about exposing this mutable state?

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
chrfwow: I got a warning from my IDE, so I renamed the variables

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: I don't want to expose a reference to the list, as this would effectively remove the lock entirely. At least the `Collections.emptyList();` should not allocate anything.
I don't think it would add churn per evaluation, as I don't use the getter during the evaluation, but use the field directly.
Also consider that this and the corresponding change in `OpenFeatureAPI` is technically a breaking change

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -318,18 +313,23 @@ public void addHooks(Hook... hooks) {
...
-        }
+        return new ArrayList<>(this.apiHooks);
+    }
+
+    /**
+     * Returns a reference to the collection of {@link Hook}s.
+     *
+     * @return The collection of {@link Hook}s.
+     */
+    Collection<Hook> getMutableHooks() {
```
chrfwow, line 324: I really don't like this method, but we need some Method where we can return a `ConcurrentLinkedQueue` and don't have to convert it into a list to reduce memory churn during evaluations

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
chrfwow: Upon further review, I'm not so sure. We might get into trouble in the `runHandlersForProvider`, but it might also work just fine. 
The problem lies rather with setting/getting the provider and its state from the `providerRepository` than anything we do in here.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -136,19 +133,15 @@ public OpenFeatureClient addHooks(Hook... hooks) {
      */
     @Override
     public List<Hook> getHooks() {
-        try (AutoCloseableLock __ = this.hooksLock.readLockAutoCloseable()) {
-            return this.clientHooks;
-        }
+        return new ArrayList<>(this.clientHooks);
```
chrfwow, line 136: We can't return the `ConcurrentLinkedQueue`, as this method returns a `List`.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -157,32 +150,33 @@ public OpenFeatureClient setEvaluationContext(EvaluationContext evaluationContex
...
+            var stateManager = openfeatureApi.getFeatureProviderStateManager(this.domain);
             // provider must be accessed once to maintain a consistent reference
-            provider = stateManager.getProvider();
-            ProviderState state = stateManager.getState();
+            var provider = stateManager.getProvider();
+            var state = stateManager.getState();
 
             mergedHooks = ObjectUtils.merge(
-                    provider.getProviderHooks(), flagOptions.getHooks(), clientHooks, openfeatureApi.getHooks());
+                    provider.getProviderHooks(), flagOptions.getHooks(), clientHooks, openfeatureApi.getMutableHooks());
```
chrfwow, line 177: Directly access the `openfeatureApi`'s `ConcurrentLinkedQueue`

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: Done

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
toddbaert: You mean it's breaking in that the behavior changes, right (different references)? There's no associated compilation breakages here (or I'm missing something?).

If so, I don't think I would _really_ qualify that as breaking... it's a bit of a grey area, but IMO if we don't make specific guarantees about references, we don't necessarily need to adhere to them. I think if we considered all behavior as completely unchangeable we couldn't even improve the locking :sweat_smile: .

chrfwow: Yes, that's what I meant

toddbaert (approved):

toddbaert:
I'll merge this tomorrow unless I hear objections.

PR #1506, opened 2025-07-06
feat: Added hook data

### LINKED ISSUE #1472: Add Hook Data
## Summary

The OpenFeature specification now includes support for hook data, which allows hooks to maintain state across their execution stages (before → after/error → finally). This feature enables use cases like distributed tracing, performance monitoring, and multi-stage validation.

This issue tracks the implementation of hook data support in this SDK to achieve compliance with the latest specification.

## Action Required

SDK maintainers need to implement hook data support according to the specification defined in:
- https://github.com/open-feature/spec/pull/273 - Initial hook data specification
- https://github.com/open-feature/spec/pull/307 - Hook data updates

A reference implementation is available in the .NET SDK PR: https://github.com/open-feature/dotnet-sdk/pull/387

## Hook data specification enables stateful hook operations

The OpenFeature hook data specification provides a mechanism for different stages of the same hook to share state during a single flag evaluation. This feature supports implementing telemetry, performance monitoring, and resource management use cases. The .NET SDK's implementation in https://github.com/open-feature/dotnet-sdk/pull/387 demonstrates a successful approach that other SDKs can follow.

## Reference Links

- **Specification PRs**: 
  - https://github.com/open-feature/spec/pull/273 - Initial hook data specification
  - https://github.com/open-feature/spec/pull/307 - Hook data updates
- **Implementation Example**: https://github.com/open-feature/dotnet-sdk/pull/387
- **Documentation**:
  - [OpenFeature Hook Specification](https://openfeature.dev/specification/sections/hooks/)
  - [Hook Concepts Documentation](https://openfeature.dev/docs/reference/concepts/hooks/)

## Understanding hook data fundamentals

Hook data provides the ability for hooks to maintain state across their execution lifecycle. When a hook starts an operation in the `before` stage (like opening a telemetry span), it needs a way to access that same data in the `after` stage to complete the operation. Without hook data, developers resort to workarounds like global state, weak maps, or thread-local storage, which introduce complexity and potential bugs.

The specification defines hook data as a **mutable, per-hook, per-evaluation** data structure. This scoping means each hook instance gets its own isolated data store that persists only for the duration of a single flag evaluation. Different hooks evaluating the same flag don't share data, preventing interference and ensuring predictable behavior.

### Data structure and access patterns

Hook data **MUST** be implemented as a structure supporting arbitrary properties with string keys and values of any type. The structure must be mutable to allow stages to modify shared state. SDKs typically implement this as a map-like interface with `set(key, value)` and `get(key)` methods.

The data structure must be created before the first hook stage executes and remain available throughout all stages (before → after/error → finally). This lifecycle management is the SDK's responsibility, not the hook developer's.

### Isolation and scoping guarantees

Each hook instance maintains its own hook data - there's no sharing between different hooks. This isolation extends to:
- Different hook types running for the same evaluation
- Multiple instances of the same hook type
- Hooks registered at different levels (global, client, invocation)

The specification explicitly states that hook data is not shared between different hooks, ensuring separation of concerns.

### Integration with hook context

Hook data must be accessible through the hook context parameter passed to all hook stages. The typical pattern follows:

```pseudocode
context.hookData.set("key", value)  // In before stage
value = context.hookData.get("key") // In after stage
```

## Learning from the .NET SDK implementation

The .NET SDK's PR https://github.com/open-feature/dotnet-sdk/pull/387 provides implementation insights. Their approach enhanced the existing `HookContext<T>` class with a `Data` property, maintaining backward compatibility while adding new functionality. Lessons include:

**API Design**: The implementation uses a simple key-value store pattern that's intuitive for developers. Type safety is handled through casting when retrieving values, following .NET conventions.

**No Breaking Changes**: Existing hooks continue working without modification. The hook data feature is additive - hooks can opt-in to using it without requiring changes to hooks that don't need state sharing.

**Examples**: The implementation includes examples like timing hooks that measure execution duration, demonstrating the feature's value to developers.

### Phase 1: Core implementation

Start by implementing the basic hook data structure with these capabilities:
- String-keyed storage with any-type values
- Mutable data access through set/get operations
- Proper lifecycle management (creation before first stage, cleanup after finally)
- Thread-safe access for concurrent evaluations

### Phase 2: Integration and testing

Integrate hook data with your existing hook infrastructure:
- Add data property to hook context
- Ensure data flows through all hook stages
- Implement isolation between hook instances
- Create comprehensive test suites covering edge cases

### Phase 3: Documentation and examples

Provide guidance for hook developers:
- API documentation with type signatures
- Examples (timing, telemetry, validation)
- Best practices and anti-patterns
- Migration guide if introducing any changes

### Memory management

Hook data should be garbage collected after the evaluation completes. Consider implementing weak references or explicit cleanup to prevent memory leaks in long-running applications.

### Thread safety

Many applications evaluate flags concurrently. Ensure hook data access is thread-safe without introducing significant performance overhead. Consider using concurrent data structures appropriate for your language.

### Type safety

In statically-typed languages, consider providing type-safe access patterns while maintaining the flexibility of arbitrary value storage. Generic methods or type assertions can help balance safety with usability.

### Performance impact

Hook data should add minimal overhead to flag evaluations. Consider lazy initialization - only create the data structure when a hook actually uses it. Benchmark the implementation to ensure acceptable performance.

### OpenTelemetry integration

The canonical use case involves creating spans in the `before` stage and ending them in `after`:

```java
// Before stage
Span span = tracer.spanBuilder("feature-flag-evaluation").startSpan();
context.hookData.set("span", span);

// After stage
Span span = (Span) context.hookData.get("span");
span.end();
```

### Performance monitoring

Tracking evaluation duration is a common requirement:

```csharp
// Before stage
context.Data.Set("startTime", DateTime.UtcNow);

// After stage
var duration = DateTime.UtcNow - (DateTime)context.Data.Get("startTime");
logger.LogInformation($"Evaluation took {duration.TotalMilliseconds}ms");
```

### Multi-stage validation

Complex validation scenarios benefit from accumulating results:

```python
# Before stage
context.hook_data.set("validation_errors", [])

# Multiple validation stages can add errors
errors = context.hook_data.get("validation_errors")
if not is_valid(context):
    errors.append("Validation failed")

# Finally stage checks accumulated errors
if context.hook_data.get("validation_errors"):
    logger.warning(f"Validation issues: {errors}")
```

## Testing requirements

Comprehensive test coverage should include:

1. **Basic functionality**: Set and get operations work correctly
2. **Isolation**: Data doesn't leak between hook instances
3. **Lifecycle**: Data persists through all stages but not beyond
4. **Type handling**: Various data types can be stored and retrieved
5. **Edge cases**: Null values, missing keys, concurrent access
6. **Performance**: Minimal overhead for evaluations not using hook data

## Migration and compatibility

When implementing hook data support:

1. **Maintain backward compatibility**: Existing hooks must continue working
2. **Version appropriately**: Follow semantic versioning based on impact
3. **Provide migration guides**: Help developers adopt the new feature
4. **Consider feature flags**: Allow gradual rollout of hook data support

## Success criteria

A complete hook data implementation should:
- Pass all specification compliance tests
- Support the primary use cases (telemetry, monitoring, validation)
- Add minimal performance overhead
- Provide good developer experience
- Include comprehensive documentation
- Maintain backward compatibility

## Conclusion

Hook data provides a feature that enables observability and monitoring capabilities in OpenFeature. By following the specification requirements and learning from successful implementations like the .NET SDK, maintainers can provide consistent, reliable hook data support across all OpenFeature SDKs. The implementation should balance flexibility with safety, performance with functionality, and maintain the isolation guarantees that make hook data predictable and useful.

## Additional Resources

- https://github.com/open-feature/spec
- [Hook Specification Details](https://openfeature.dev/specification/sections/hooks/)
- [Hook Concepts Guide](https://openfeature.dev/docs/reference/concepts/hooks/)
- [.NET SDK Documentation](https://openfeature.dev/docs/reference/technologies/server/dotnet/)
- https://github.com/open-feature/dotnet-sdk/releases

mdxabu:
## This PR
Added the Hook Data

- adds this new feature

### Related Issues

Fixes #1472

@beeme1mr, Please review this if there is any problem or if I did anything wrong, ping me!

### DIFF
--- a/src/main/java/dev/openfeature/sdk/HookData.java
+++ b/src/main/java/dev/openfeature/sdk/HookData.java
@@ -45,7 +45,7 @@ public interface HookData {
     }
 
     /**
-     * Default thread-safe implementation of HookData.
+     * Default implementation of HookData.
      */
     public class DefaultHookData implements HookData {
         private final Map<String, Object> data = new HashMap<>();


src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    static HookData create() {
+        return new DefaultHookData();
+    }
+
+    /**
+     * Default thread-safe implementation of HookData.
+     */
+    public class DefaultHookData implements HookData {
+        private final Map<String, Object> data = Collections.synchronizedMap(new HashMap<>());
```
chrfwow: Do we need a thread safe implementation? If the hook data only lives for the duraion of a flag evalution, there will probably not be concurrent access to this object. IIRC, a flag evaluation happens synchronously.

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param <T> the type to cast to
+     * @param key the key to retrieve the value for
+     * @param type the class to cast to
+     * @return the value cast to the specified type, or null if not found
+     * @throws ClassCastException if the value cannot be cast to the specified type
+     */
+    <T> T get(String key, Class<T> type);
+
+    /**
+     * Default implementation using ConcurrentHashMap for thread safety.
```
chrfwow: `DefaultHookData` uses a `Collections.synchronizedMap(new HashMap<>())`, not a `ConcurrentHashMap`

src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -25,6 +25,12 @@ public class HookContext<T> {
     ClientMetadata clientMetadata;
     Metadata providerMetadata;
 
+    /**
+     * Hook data provides a way for hooks to maintain state across their execution stages.
+     * Each hook instance gets its own isolated data store.
+     */
+    HookData hookData;
```
chrfwow, line 32: With this addition, the Constructor of the `HookContext` changes, which may be a breaking change @toddbaert

chrfwow:
@mdxabu you will need to force push a commit that is signed off, otherwise the DCO check will not succeed

mdxabu:
> @mdxabu you will need to force push a commit that is signed off, otherwise the DCO check will not succeed

Can I squash those commits into one commit, and then I will sign off?

chrfwow:
> Can I squash those commits into one commit, and then I will sign off?

Yes, you could do that. If you click on the failed DCO action, you schould see a short explanation

mdxabu:
@chrfwow, I force push the commits! Now DCO action runs successfully.

chrfwow:
The build currently fails because of formatting issues. You can resolve them automatically by running the spotless plugin

src/main/java/dev/openfeature/sdk/HookContext.java
toddbaert: That's certainly a breaking change, yep. @mdxabu you can add new constructors, but you'll have to preserve any old ones. Please add unit tests to confirm this.

src/main/java/dev/openfeature/sdk/HookData.java
toddbaert: Ya I think somebody would have to really _try_ to create a concurrency issue here, because of the nature of hooks. I agree this seems un-needed.

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Hook data provides a way for hooks to maintain state across their execution stages.
+ * Each hook instance gets its own isolated data store that persists only for the duration
+ * of a single flag evaluation.
+ */
+public interface HookData {
```
toddbaert, line 11: We may be able to use some of the functionality in [Structure](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Structure.java) or it's implementations to do this. We don't _need_ all of those methods, but some wouldn't hurt, I think, and it would reduce duplication and promote consistency.

alexandraoberaigner:
@mdxabu how is it going? We do have an internal dependency on this. Can we support somehow? :)

mdxabu:
Yeah, Sure!

alexandraoberaigner:
> Yeah, Sure!

Great, let me know how I can support :)

src/main/java/dev/openfeature/sdk/HookData.java
alexandraoberaigner: @toddbaert I took a closer look and found that `Structure` enforces that all stored data must be of type `Value`, which conflicts with the [OPENFEATURE SPEC](https://openfeature.dev/specification/sections/hooks#:~:text=string%2C%20and-,values%20of%20any%20type,-.)  allowing values of any type in hookdata. This restriction makes `Structure` unsuitable for hook data. As far as I understand it, sticking with @mdxabu 's implementation aligns better with the spec. Pls let me know if I missed anything! 🙏

toddbaert: Yes, but can use lombok to [delegate only some methods](https://github.com/search?q=repo%3Aopen-feature%2Fjava-sdk%20%40delegate&type=code).

@aepfli what do you think? You have been working on improving consistency of our POJOs... what is you opinion on this?

aepfli: I am actually questioning some of my rewokr based on this findings. -> yes we do have the need for some POJO's to contain only string/numbers/booleans/etc - but i am not sure we need this kind of structure for all of them. especially like the hookcontext is not used for futher evaluation etc. it is used within the same hook, and can have any kind of data. i think that a pure object map would be sufficient in this case. and we should allow all kind of objects (structure limits this, but structure has a dedicated usage for certain operations)

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param <T> the type to cast to
+     * @param key the key to retrieve the value for
+     * @param type the class to cast to
+     * @return the value cast to the specified type, or null if not found
+     * @throws ClassCastException if the value cannot be cast to the specified type
+     */
+    <T> T get(String key, Class<T> type);
+
+    /**
+     * Default implementation using ConcurrentHashMap for thread safety.
```
alexandraoberaigner:
```suggestion
     * Default implementation is using a synchronized map. 
```
Update accordingly after [this comment](https://github.com/open-feature/java-sdk/pull/1506#discussion_r2333875139) is addressed

src/main/java/dev/openfeature/sdk/HookData.java
alexandraoberaigner: Should we use a normal HashMap instead of the synchronized Map then?

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    static HookData create() {
+        return new DefaultHookData();
+    }
+
+    /**
+     * Default thread-safe implementation of HookData.
+     */
+    public class DefaultHookData implements HookData {
+        private final Map<String, Object> data = Collections.synchronizedMap(new HashMap<>());
```
alexandraoberaigner:
```suggestion
        private final Map<String, Object> data = new HashMap<>();
```
Maybe I'm not seeing the full picture here, but if we don't need thread-safety [as stated here](https://github.com/open-feature/java-sdk/pull/1506#discussion_r2189436262), I guess a normal hash map would be sufficient?

src/main/java/dev/openfeature/sdk/HookContext.java
alexandraoberaigner: This should be done with [this test](https://github.com/open-feature/java-sdk/pull/1506/commits/13dbeeefeafcc75c15fe8e515c9ebf1675e228ad#diff-2e7c27ca29c5199dc0028ebb9d12db488f180c725b4399ce86ace63c7b79dfc9:~:text=shouldCreateHookContextWithoutHookData)

aepfli (approved):
Looks good from my side. Thank you for the great and valuable addition. I think this will make sharing data between hook steps way easier. Thank you!

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,76 @@
+package dev.openfeature.sdk;
+
+import java.util.Collections;
```
aepfli:
```suggestion
```
not needed anymore :)

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param <T> the type to cast to
+     * @param key the key to retrieve the value for
+     * @param type the class to cast to
+     * @return the value cast to the specified type, or null if not found
+     * @throws ClassCastException if the value cannot be cast to the specified type
+     */
+    <T> T get(String key, Class<T> type);
+
+    /**
+     * Default implementation is using a synchronized map.
```
aepfli:
```suggestion
     * Default implementation is using a synchronized map.
```

fixing spotless

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,75 @@
...
```
chrfwow:
```suggestion
     * Default implementation uses a HashMap.
```
Do we even need this comment? I think this should be transparent to the user

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    static HookData create() {
+        return new DefaultHookData();
+    }
+
+    /**
+     * Default thread-safe implementation of HookData.
```
chrfwow:
```suggestion
     * Default implementation of HookData.
```
Not thread safe anymore

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -87,10 +88,21 @@ private EvaluationContext callBeforeHooks(
...
         Collections.reverse(reversedHooks);
         EvaluationContext context = hookCtx.getCtx();
+
+        // Create hook data for each hook instance
+        Map<Hook, HookData> hookDataMap = new HashMap<>();
+        for (Hook hook : reversedHooks) {
+            if (hook.supportsFlagValueType(flagValueType)) {
+                hookDataMap.put(hook, HookData.create());
+            }
+        }
```
chrfwow, line 98: Why do we need this map and loop? We don't return the map, so it will be GC'ed after the method returns, and we could create the `HookData` in the loop below like this:
```HookContext contextWithHookData = hookCtx.withHookData(HookData.create());```

File: src/test/java/dev/openfeature/sdk/HookDataTest.java [hunk 12]
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                } finally {
+                    latch.countDown();
+                }
+            });
+        }
+
+        assertTrue(latch.await(10, TimeUnit.SECONDS));
+        executor.shutdown();
+    }
```
chrfwow, line 101: I'd expect this test to fail. HookData is not thread safe anymore. We can remove this test. (In the future, such tests should be done with [VmLens](https://github.com/open-feature/java-sdk/pull/1567), but that is not merged yet)

src/test/java/dev/openfeature/sdk/HookSupportTest.java
```diff
@@ -105,4 +112,33 @@ private EvaluationContext evaluationContextWithValue(String key, String value) {
         EvaluationContext baseContext = new ImmutableContext(attributes);
         return baseContext;
     }
+
+    private static class TestHook implements Hook<String> {
```
chrfwow, line 116: Seems to be unused

chrfwow (changes requested):

aepfli:
hey @mdxabu - we took a look at your implementation and it is a good foundation, but we have seen some issues, with how data was enriched, and also with the object churn (creating a lot of objects). So we gave it a shot on our own here #1587 - we used your pr as a basis, and we will keep all of your commits. we just wanted to make sure you are okay with this. Thank you for the effort and the foundation.

mdxabu:
> #1587

Yeah, I'm okay with that. I'm glad you're asking for my opinion! I like this community!

PR #1567, opened 2025-08-19
feat: Add vmlens checks

chrfwow:
Adds a [vmlens](https://vmlens.com/) check to this repo. This tool will us in the future to find and fix concurrency issues in our code base.
VMLens works by attaching a Java agent during the test run, which will intercept calls inside the `allInterleavings.hasNext()` loop. The agent goes over all possible iterations of interleavings that threads could do, and thereby discoveres race conditions.
```
        try (AllInterleavings allInterleavings = new AllInterleavings("Concurrent evaluations")) {
            while (allInterleavings.hasNext()) {
                Runner.runParallel(
                        () -> assertEquals("def", client.getStringValue("a", "a")),
                        () -> assertEquals("as", client.getStringValue("b", "b")));
            }
        }
```
Asserts inside the tests help to verify that no operation was done on an intermediate state of the application.
Note that iterating over all interleavings can take some time, so the tests should be kept short.

The agent is installed by the `vmlens-maven-plugin` as needed.


aepfli (approved):

toddbaert:
I definitely think this is cool, and useful.

I have a concern about committing binaries though; to me, these always present a vector for attack in OSS projects, since it's very easy to miss changes to them and nearly impossible to verify their contents.

Could we instead download this files if they don't exist with `dependency:get` or maybe with `maven-download-plugin` or even a script (probably not he best option as it will break on windows)?

If you think I'm being overly worried here, please feel free to object. cc @aepfli @beeme1mr @lukas-reining

chrfwow:
I agree with you, @toddbaert, having the binary in our code base is not optimal. If we choose to download the jar file during each build run, would you say we have to make sure that we check the integrity of the jar file by comparing it to some hash code, so that we can be sure the file wasn't tampered with?

.gitignore
```diff
@@ -13,3 +13,6 @@ target
 
 # used for spec compliance tooling
 java-report.json
+
+# vmlens stuff
+/vmlens-agent/vmlens/
```
aepfli: do we still need this ignore?

chrfwow: no. nice catch

chrfwow:
@toddbaert I removed the vmlens .jar files, because they are not needed. The vmlens-maven-plugin installs the agent on its own

src/main/java/dev/openfeature/sdk/Awaitable.java
```diff
@@ -10,7 +10,7 @@ public class Awaitable {
      */
     public static final Awaitable FINISHED = new Awaitable(true);
 
-    private boolean isDone = false;
+    private volatile boolean isDone = false;
```
liran2000: AtomicBoolean can be better

chrfwow: True, but not required here. We do all accesses to `isDone` inside `synchronized` blocks.

liran2000: In this case, why volatile is needed?

chrfwow: It seems not to be needed. Just some leftover of some experiments I did. Thx

PR #1587, opened 2025-09-10
feat: add hook data 

### LINKED ISSUE #1472: Add Hook Data

alexandraoberaigner:
## This PR
improves PR #1506 from @mdxabu  who started the work on adding hook data

### Related Issues

Fixes #1472

### Notes
Tests & Impl are still WIP: 
* HookData isolation might not fully work yet (not sure if the test is setup incorrectly or we need to adapt the implementation still)
* pls let me know if you see any edge cases that I didn't think of yet 🙏

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -50,7 +50,7 @@ jobs:
       run: mvn --batch-mode --update-snapshots verify
 
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v5.5.0
+      uses: codecov/codecov-action@v5.5.1
       with:
         token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
         flags: unittests # optional
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -29,7 +29,7 @@ jobs:
             cache: maven
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@0d33fd9f263def169b43518224e68932ef8cc79a
+        uses: github/codeql-action/init@aa90e97ad2ed17cde6a43e89f70138299e64f837
         with:
           languages: java
 
@@ -46,7 +46,7 @@ jobs:
 
       - if: matrix.build.java == '17'
         name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v5.5.0
+        uses: codecov/codecov-action@v5.5.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
           flags: unittests # optional
@@ -55,4 +55,4 @@ jobs:
           verbose: true # optional (default = false)
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@0d33fd9f263def169b43518224e68932ef8cc79a
+        uses: github/codeql-action/analyze@aa90e97ad2ed17cde6a43e89f70138299e64f837
--- a/.github/workflows/static-code-scanning.yaml
+++ b/.github/workflows/static-code-scanning.yaml
@@ -33,12 +33,12 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@0d33fd9f263def169b43518224e68932ef8cc79a
+        uses: github/codeql-action/init@aa90e97ad2ed17cde6a43e89f70138299e64f837
         with:
           languages: java
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@0d33fd9f263def169b43518224e68932ef8cc79a
+        uses: github/codeql-action/autobuild@aa90e97ad2ed17cde6a43e89f70138299e64f837
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@0d33fd9f263def169b43518224e68932ef8cc79a
+        uses: github/codeql-action/analyze@aa90e97ad2ed17cde6a43e89f70138299e64f837
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.38</version>
+            <version>1.18.40</version>
             <scope>provided</scope>
         </dependency>
 
@@ -60,7 +60,7 @@
             <!-- used so that lombok can generate suppressions for spotbugs. It needs to find it on the relevant classpath -->
             <groupId>com.github.spotbugs</groupId>
             <artifactId>spotbugs</artifactId>
-            <version>4.9.4</version>
+            <version>4.9.5</version>
             <scope>provided</scope>
         </dependency>
 
@@ -258,7 +258,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>3.5.3</version>
+                <version>3.5.4</version>
                 <configuration>
                     <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
@@ -277,7 +277,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
-                <version>3.5.3</version>
+                <version>3.5.4</version>
                 <configuration>
                     <argLine>
                         ${surefireArgLine}
@@ -403,7 +403,7 @@
                     <plugin>
                         <groupId>com.github.spotbugs</groupId>
                         <artifactId>spotbugs-maven-plugin</artifactId>
-                        <version>4.9.4.0</version>
+                        <version>4.9.4.2</version>
                         <configuration>
                             <excludeFilterFile>spotbugs-exclusions.xml</excludeFilterFile>
                             <plugins>
@@ -419,7 +419,7 @@
                             <dependency>
                                 <groupId>com.github.spotbugs</groupId>
                                 <artifactId>spotbugs</artifactId>
-                                <version>4.9.4</version>
+                                <version>4.9.5</version>
                             </dependency>
                         </dependencies>
                         <executions>
@@ -663,7 +663,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
-                        <version>3.5.3</version>
+                        <version>3.5.4</version>
                         <configuration>
                             <argLine>
                                 ${surefireArgLine}
@@ -679,7 +679,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-failsafe-plugin</artifactId>
-                        <version>3.5.3</version>
+                        <version>3.5.4</version>
                         <configuration>
                             <argLine>
                                 ${surefireArgLine}


src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -5,62 +5,92 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Consumer;
+import java.util.function.BiConsumer;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.Pair;
```
chrfwow: This makes the build fail: 
```
Error:  Used undeclared dependencies found:
Error:     org.apache.commons:commons-lang3:jar:3.18.0:provided
```
Can we use something else instead?

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -71,26 +101,51 @@ private <T> void executeChecked(Hook<T> hook, Consumer<Hook<T>> hookCode, String
...
+        // Create hook data for each hook instance
+        Map<Hook, HookData> hookDataMap = new HashMap<>();
         for (Hook hook : reversedHooks) {
             if (hook.supportsFlagValueType(flagValueType)) {
-                Optional<EvaluationContext> optional =
-                        Optional.ofNullable(hook.before(hookCtx, hints)).orElse(Optional.empty());
+                hookDataMap.put(hook, HookData.create());
+            }
+        }
+        */
```
chrfwow: Can be removed

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -164,29 +165,21 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
...
-            var stateManager = openfeatureApi.getFeatureProviderStateManager(this.domain);
+            final var stateManager = openfeatureApi.getFeatureProviderStateManager(this.domain);
             // provider must be accessed once to maintain a consistent reference
-            var provider = stateManager.getProvider();
-            var state = stateManager.getState();
-
+            final var provider = stateManager.getProvider();
+            final var state = stateManager.getState();
+            afterHookContext = HookContext.from(
+                    key, type, this.getMetadata(), provider.getMetadata(), mergeEvaluationContext(ctx), defaultValue);
```
chrfwow: I am not sure if we should set the `afterHookContext` here. The value assigned here is not the context for the after hooks and technically, it might change the context available to the error hooks if there is an error when merging the context

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -71,26 +101,51 @@ private <T> void executeChecked(Hook<T> hook, Consumer<Hook<T>> hookCode, String
...
 
     private EvaluationContext callBeforeHooks(
-            FlagValueType flagValueType, HookContext hookCtx, List<Hook> hooks, Map<String, Object> hints) {
+            FlagValueType flagValueType,
+            HookContext hookCtx,
+            List<Pair<Hook, HookData>> hookDataPairs,
+            Map<String, Object> hints) {
         // These traverse backwards from normal.
-        List<Hook> reversedHooks = new ArrayList<>(hooks);
+        List<Pair<Hook, HookData>> reversedHooks = new ArrayList<>(hookDataPairs);
```
chrfwow, line 124:
```suggestion
        List<Pair<Hook, HookContext>> reversedHooks = new ArrayList<>(hookDataPairs);
```
Is this possible? That way we wouldn't need to create the `HookContext` from the `HookData` every time we invoke hooks

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -5,62 +5,92 @@
...
+            List<Pair<Hook, HookData>> hookDataPairs,
             Map<String, Object> hints) {
-        executeHooks(flagValueType, hooks, "error", hook -> hook.error(hookCtx, e, hints));
+        executeHooks(flagValueType, hookDataPairs, hookCtx, "error", (hook, ctx) -> hook.error(ctx, e, hints));
+    }
+
+    public List<Pair<Hook, HookData>> getHookDataPairs(List<Hook> hooks) {
+        var pairs = new ArrayList<Pair<Hook, HookData>>();
+        for (Hook hook : hooks) {
+            pairs.add(Pair.of(hook, HookData.create()));
```
chrfwow: We might be able to wrap creating the hook pair with an if
```
if (hook.supportsFlagValueType(flagValueType)) {
// create
}
```
This way we only add hooks to the list that will be run, which should also make this check in the hook execution redundant.

src/test/java/dev/openfeature/sdk/HookDataTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        hookData.set("key", "original");
+        assertEquals("original", hookData.get("key"));
+
+        hookData.set("key", "updated");
+        assertEquals("updated", hookData.get("key"));
+    }
+
+    @Test
+    void shouldBeThreadSafe() throws InterruptedException {
```
chrfwow: Please remove this test

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: i will use a simpe list of map.Entries or create an own class for this - pair is really easy

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
aepfli: it is a flow issue, what if we have an exception before we reach the afterhookcontext, eg. in one of the before hooks - then we do not have any kind of hook context for the error hook. I know this is not ideal, most likely due to naming. what about simple renaming to just hookContext?

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: so theory crafting, but imho worth an own follow up pr.

1. We rename the hookContext to HookContextWithoutData (strange naming, maybe we find a better one)
2. We create an interface called HookContext to enure backwards compatibility with all kind of constructors etc. and static methods (like `.from`  - not sure if we can do that)
3. We add methods for getting HookData to HookContext
4. HookContextWithoutData should implement this interface, but we do not return anything for the getters for hookdata, or an Global empty immutable object (we need to discuss this)
5. We generate a HookContextWithDataDecorator. which:
    - implements hookContext
    - has two fields: hookContext and hookData
    - delegates all methods to hookContext, except hookData
6. this way we have backwards compatibility and we do not need to duplicate the hook context multiple times.

wdyt?

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: really good idea! added

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: we can do that, but i think currently we have this behaviour were the context would be `null` in some error cases

src/main/java/dev/openfeature/sdk/HookSupport.java
chrfwow: No static methods in interfaces in java, but couldn't we just make `HookContextWithDataDecorator` extend `HookContext`?

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: create a package private pair class

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
aepfli: should it be null in some errorcases? that is the question, because i do think the context is important (provider and state) - so i think it is valuable information even on error cases. i think the old implementation is more like a bug, and this behaviour feels "more" right

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: i am not a big fan of inheritance, and prefer composition, and i do think that this will be nicer with an interface, compared to the class, it adds a dependency where we rely on implementation of hookContext, rather than a simple interface contract

src/main/java/dev/openfeature/sdk/Pair.java
```diff
@@ -0,0 +1,36 @@
...
+class Pair<K, V> {
+    private final K key;
+    private final V value;
+
+    private Pair(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    public K getKey() {
```
guidobrei: Can we just rely on `Left`/`Key` and `Right`/`Value`, and not 2 names for the same concept?

src/main/java/dev/openfeature/sdk/ImmutableContext.java
```diff
@@ -20,6 +20,8 @@
 @SuppressWarnings("PMD.BeanMembersShouldSerialize")
 public final class ImmutableContext implements EvaluationContext {
 
+    public static final ImmutableContext EMPTY = new ImmutableContext();
```
chrfwow, line 23: Do we need this?

src/test/java/dev/openfeature/sdk/benchmark/AllocationBenchmark.java
```diff
@@ -54,6 +54,30 @@ public Optional<EvaluationContext> before(HookContext<Object> ctx, Map<String, O
                 return Optional.ofNullable(new ImmutableContext());
             }
         });
+        client.addHooks(new Hook<String>() {
```
chrfwow, line 57: Doesn't this have to be a `StringHook` for the filter to work?

src/main/java/dev/openfeature/sdk/HookContextWithoutData.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return resulting context for hook
+     */
+    static <T> HookContextWithoutData<T> from(
+            String key,
+            FlagValueType type,
+            ClientMetadata clientMetadata,
+            Metadata providerMetadata,
+            EvaluationContext ctx,
+            T defaultValue) {
+        return HookContextWithoutData.<T>builder()
```
guidobrei: You could get rid of the builder.

src/main/java/dev/openfeature/sdk/ImmutableContext.java
aepfli: i am using it for creation of a hookcontext without context -> so i only have one instance, and not generate multiple. i thought here is the best place for it

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,81 @@
...
+            data.put(key, value);
+        }
+
+        @Override
+        public Object get(String key) {
+            if (data == null) {
+                return null;
+            }
+            return data.get(key);
+        }
```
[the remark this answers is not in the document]
chrfwow: Will only ever be used in a single threaded context, and lazy initialization will save memory churn in case hooks do not use HookData, which will likely be the more common case

PR #1607, opened 2025-09-16
fix: flacky e2e tests

aepfli:
Seems like we do have a race condition, where we are not entering a certain state, i assume that this might be caused by a race condition, of too many events back to back and how they are processed.

Maybe we should change the implementation of our eventing to a queue, which gets filled and taken from, to ensure events are thrown in the order they are emitted?


chrfwow (approved):

chrfwow:
> Maybe we should change the implementation of our eventing to a queue, which gets filled and taken from, to ensure events are thrown in the order they are emitted?

Might be a good idea, we don't want to miss state changes

aepfli:
superseded by #1628 - the fix and solution is way nicer  - not need for this pr anymore

PR #1620, opened 2025-09-19
feat: add hook data support

alexandraoberaigner:
## This PR
this re-adds hook data support avoiding breaking changes in the binary (we introduced with v1.18.0)

### Related Issues
#1472 

###Notes

With [v1.18.1](https://github.com/open-feature/java-sdk/pull/1602) we reverted the hook data feature:
> revert hook data to resolve bytecode incompatibility (https://github.com/open-feature/java-sdk/issues/1613) ([9845601](https://github.com/open-feature/java-sdk/commit/984560196d4a34fb21c8946d1dc675cf6565e90f))


src/main/java/dev/openfeature/sdk/HookContextWithData.java
```diff
@@ -0,0 +1,14 @@
+package dev.openfeature.sdk;
+
+public class HookContextWithData<T> extends HookContext<T> {
```
aepfli: i am not sure we want to duplicate everything, i think the previous implementation with helper methods to access the object within is fine.

src/main/java/dev/openfeature/sdk/HookData.java
```diff
@@ -0,0 +1,81 @@
...
+        public <T> T get(String key, Class<T> type) {
+            Object value = get(key);
+            if (value == null) {
+                return null;
+            }
+            if (!type.isInstance(value)) {
+                throw new ClassCastException("Value for key '" + key + "' is not of type " + type.getName());
+            }
+            return type.cast(value);
+        }
```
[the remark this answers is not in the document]
chrfwow: No need for thread safety here, reducing object churn is more important

src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -1,16 +1,18 @@
 package dev.openfeature.sdk;
 
 import lombok.Builder;
+import lombok.Data;
 import lombok.NonNull;
 import lombok.Value;
 import lombok.With;
+import lombok.experimental.NonFinal;
```
chrfwow: Why do we need this import? Looks dangerous

src/test/java/dev/openfeature/sdk/benchmark/AllocationBenchmark.java
```diff
@@ -54,6 +54,30 @@ public Optional<EvaluationContext> before(HookContext<Object> ctx, Map<String, O
                 return Optional.ofNullable(new ImmutableContext());
             }
         });
+        client.addHooks(new Hook<String>() {
```
chrfwow:
```suggestion
        client.addHooks(new StringHook () {
```
I think we need to use the dedicated implementations for the filter for flag type to work

src/main/java/dev/openfeature/sdk/HookContext.java
alexandraoberaigner: I didn't see the experimental when using it 👀 We actually don't need it - it's a remainder & unused import. I changed the Value annotation to Data  because Value makes the class final

src/main/java/dev/openfeature/sdk/Pair.java
```diff
@@ -0,0 +1,28 @@
...
+        this.value = value;
+    }
+
+    public K getLeft() {
+        return key;
+    }
+
+    public V getRight() {
+        return value;
+    }
```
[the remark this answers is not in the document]
alexandraoberaigner: No strong opinion about it. @aepfli didn' we talk about this being good practice to use left/right?

src/main/java/dev/openfeature/sdk/HookContextWithData.java
alexandraoberaigner: What do you mean with duplicate everything exactly? If I make HookContext a class instead of an interface I need to call its ctor and pass the data...or did I miss something here? or are you suggesting something else than inheritance?

alexandraoberaigner: Thinking of using a decorator...do you think this is overkill?

aepfli: i think a decorator pattern is a perfect fit for this. we can even leave the hookcontext as a @Value class - and keep its immutability

alexandraoberaigner: we cannot keep @Value unfortunately since the decorator extends from HookContext...pls let me know if I missed something

src/test/java/dev/openfeature/sdk/benchmark/AllocationBenchmark.java
alexandraoberaigner: As part of this PR or in general?

src/main/java/dev/openfeature/sdk/HookContextWithData.java
aepfli: i thought when you said decorator about something like 

```java

public class HookContextWithData<T> extends HookContext<T> {
    private final HookContext<T> hookContext;

    public HookContextWithData(HookContext hookcontext, HookData hookData) {
         this.hookdata = hookdata;
         this.hookContext = hookContext;
    }

    public getFlagKey() {
        return hookContext.getFlagKey()
    }

    // ... and for the others except hookdata

    public getHookData() {
          return hookdata;
    }
}
```

this way, we do not need to duplicate the other properties of HookContext, just the hookData is special to each of them.

alexandraoberaigner: I meant the decorator pattern 🙈

alexandraoberaigner: With your example I'd need to add a parameterless ctor to HookContext - which is not feasable because of the @nonnull annotations on the members - as soon as we extend we need to call the super() ctor

aepfli: but we could add a protected constructor to the hookContext, passing null for all the values, can't we? theoretically? (lombok makes everthing hard)

alexandraoberaigner: Well, I guess we can ignore the hint... but feels wrong
image

src/test/java/dev/openfeature/sdk/benchmark/AllocationBenchmark.java
chrfwow: At least for this benchmark. If you look at the [implementation](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Hook.java#L52), the base `Hook<T>` does not do any filtering, but the [dedicated Hooks do](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/BooleanHook.java#L13)

src/main/java/dev/openfeature/sdk/Pair.java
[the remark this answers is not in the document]
alexandraoberaigner: went for being consistent and changed it

src/test/java/dev/openfeature/sdk/benchmark/AllocationBenchmark.java
alexandraoberaigner: ooh jup, good catch! Thanks

src/main/java/dev/openfeature/sdk/HookContextWithData.java
chrfwow: If a variable is annotated with `@NonNull`, it should be non-`null`.
Why can't we make `HookContextWithData` extend `HookContext` and add a `HookData` variable? I think we talked about this already, but why don't we just add a nullable and default null `HookData` variable to `HookContext`?

aepfli: but this will not work with the decorator pattern based on extending a concrete class.

src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -51,10 +63,220 @@ public static <T> HookContext<T> from(
...
+    public Metadata getProviderMetadata() {
+        return this.providerMetadata;
+    }
+
+    public HookData getHookData() {
+        return this.hookData;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
aepfli: do we need such an exhaustive equals method? - seems like a lot of code for an equals.

src/main/java/dev/openfeature/sdk/HookExecutor.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+class HookExecutor {
```
aepfli: [Q] is this renaming necessary for this task, it currently hides the changes actually affecting the logic, and makes it harder to grasp the change, than this renaming actually helps the code base.

src/main/java/dev/openfeature/sdk/DefaultHookData.java
```diff
@@ -0,0 +1,54 @@
+package dev.openfeature.sdk;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Default implementation of HookData.
+ */
+public class DefaultHookData implements HookData {
```
aepfli, line 9: [suggestion] As we do have an interface, we can hide the concrete implementation from our users. There is not benefit, for them to instantiate this or extend this class. If somebody starts using it, it is even harder for us to not break anything. hence i suggestion to make this class package private.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -164,32 +162,24 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
...
-                    mergedHooks,
-                    hints);
-
-            afterHookContext =
-                    HookContext.from(key, type, this.getMetadata(), provider.getMetadata(), mergedCtx, defaultValue);
+            var sharedHookContext =
+                    new SharedHookContext(key, type, this.getMetadata(), provider.getMetadata(), defaultValue);
+
+            var evalContext = mergeEvaluationContext(ctx);
+            hookExecutor = HookExecutor.create(mergedHooks, sharedHookContext, evalContext, hints);
```
aepfli: [issue] Static constructors for object regularly used adds a lot of complexity for testing, we are now heavily bound during the testing of the OpenFeatureClient to the hookExecutor, mocking is only possible with advanced tooling manipulating the bytecode heavily. Although i like the idea of creating one executor per evaluation, i think this adds more complexity to the code now, and maintainability is decreased. Especially our hotpath should be easy to test without dependencies. Hence for clarity i am oppose introducing static creator methods for logic related objects.

src/test/java/dev/openfeature/sdk/DefaultHookDataTest.java
```diff
@@ -0,0 +1,79 @@
+package dev.openfeature.sdk;
+
+import static org.junit.jupiter.api.Assertions.*;
```
aepfli: [nit] no star imports please, i know checkstyle is not checking our tests for this, but generally we should try to avoid it as much as possible

src/main/java/dev/openfeature/sdk/HookContext.java
alexandraoberaigner: this was created by delomboking the annotation which we don't want anymore but need to keep the interface to not break anything. Do you think we should simplify the implementation? I tend to not touching the generated code...

src/main/java/dev/openfeature/sdk/HookExecutor.java
alexandraoberaigner: done

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
alexandraoberaigner: Pls let me know if the current changes are going into the direction you had in mind :)

aepfli: it is going into the direction, but does not solve the issue, that our logic is still tightly coupled. We should keep the initialization within the constructor of our openFeatureClient, and split logic and data into two separate objects.

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -5,97 +5,98 @@
...
-            for (Hook hook : hooks) {
-                if (hook.supportsFlagValueType(flagValueType)) {
-                    executeChecked(hook, hookCode, hookMethod);
-                }
+    public void executeBeforeHooks() {
+        // These traverse backwards from normal.
+        List<Pair<Hook, HookContext>> reversedHooks = new ArrayList<>(hooks);
+        Collections.reverse(reversedHooks);
+
+        for (Pair<Hook, HookContext> hookContextPair : reversedHooks) {
```
chrfwow: Alternatively, we could iterate with a for i loop in reverse which would be slightly faster, but maybe less readable. Not a requirement for me

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -5,97 +5,98 @@
...
+            var hook = hookContextPair.getKey();
+            var hookContext = hookContextPair.getValue();
+
+            Optional<EvaluationContext> returnedEvalContext =
+                    Optional.ofNullable(hook.before(hookContext, hints)).orElse(Optional.empty());
+            if (returnedEvalContext.isPresent()) {
+                // update shared evaluation context for all hooks
+                setEvaluationContext(evaluationContext.merge(returnedEvalContext.get()));
```
chrfwow: I wonder if we should collect all hook contexts in a list and merge them only after the loop. This would mean that the context changes of a hook cannot be read by another hook. Not sure if this would be spec compliant though

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -222,9 +212,11 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
             }
             details.setErrorMessage(e.getMessage());
             enrichDetailsWithErrorDefaults(defaultValue, details);
-            hookSupport.errorHooks(type, afterHookContext, e, mergedHooks, hints);
+            hookSupport.executeErrorHooks(e);
```
chrfwow: add non null check to `hookSupport`

src/main/java/dev/openfeature/sdk/HookContext.java
aepfli: i think this has already changed - the equals are way smaller already

src/main/java/dev/openfeature/sdk/HookContext.java
```diff
@@ -51,6 +75,300 @@ public static <T> HookContext<T> from(
...
+
+    public ClientMetadata getClientMetadata() {
+        return sharedContext.getClientMetadata();
+    }
+
+    public Metadata getProviderMetadata() {
+        return sharedContext.getProviderMetadata();
+    }
+
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "Intentional exposure of hookData")
```
chrfwow, line 120: I wonder why this is needed. `HookData` is already public

src/main/java/dev/openfeature/sdk/HookSupportData.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import lombok.Getter;
+
+/**
+ * Encapsulates data for hook execution per flag evaluation
+ */
+@Getter
+class HookSupportData {
```
chrfwow, line 11: I don't see the value of separating `HookSupport` and `HookSupportData`. We need to create a new object for every flag evaluation anyway

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -164,32 +165,25 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
...
-            var stateManager = openfeatureApi.getFeatureProviderStateManager(this.domain);
+            final var stateManager = openfeatureApi.getFeatureProviderStateManager(this.domain);
             // provider must be accessed once to maintain a consistent reference
-            var provider = stateManager.getProvider();
-            var state = stateManager.getState();
+            final var provider = stateManager.getProvider();
+            final var state = stateManager.getState();
 
-            mergedHooks = ObjectUtils.merge(
+            var mergedHooks = ObjectUtils.merge(
```
chrfwow, line 177: we could already set the hooks of the `hookData(Support)` object here, in case somethin goes wrong in the context merge

src/main/java/dev/openfeature/sdk/HookSupportData.java
alexandraoberaigner: This was my solution to @aepfli's [comment here](https://github.com/open-feature/java-sdk/pull/1620#discussion_r2376653272). It should simplify testing

src/main/java/dev/openfeature/sdk/HookContext.java
alexandraoberaigner: spotbugs complained... 🤔

src/main/java/dev/openfeature/sdk/HookSupport.java
alexandraoberaigner: I couldn't find something in the spec right away, but with merging it right away we will be at least consistent with what is implemented [in dotnet](https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeature/HookRunner.cs#L71)

src/main/java/dev/openfeature/sdk/HookSupport.java
alexandraoberaigner: I'm gonna go for readability here - it's already weird that its reversed so we shouldn't hide that behind some less readable logic

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
alexandraoberaigner: pls rereview

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
alexandraoberaigner: does this still apply? & if so, can you elaborate how/why you would add the null check?

src/main/java/dev/openfeature/sdk/HookSupport.java
aepfli: what if the execution of one hook is failing, is the context computed in the mean time valuable? - if yes, we should set it each time. eg. a counter argument to in dotnet is, we do it like that in js

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
aepfli: I am torn by this appraoch. HookSupportData is not a full POJO, it hides logic, and has logic inside, we need to `initialize` it, hence it is not really a POJO. We could actually use setters for this too.

For me the logic is also a little bit flawed currently, if we throw an exception before the hookSupportData is generated, we will also create NPE's in the catch block, as the object is not initialized.

I like the idea of the HookSupportData as it reduces the amount of parameters for the methods, and makes the API of hookSupport cleaner. But we might just want to use simple package private getters and setters to manipulated the object one after another, without the need of an initialization.

src/main/java/dev/openfeature/sdk/SharedHookContext.java
```diff
@@ -0,0 +1,52 @@
...
+        this.clientMetadata = clientMetadata;
+        this.providerMetadata = providerMetadata;
+        this.defaultValue = defaultValue;
+    }
+
+    public HookContext<T> hookContextFor(EvaluationContext evaluationContext, HookData hookData) {
+        return new HookContext<>(this, evaluationContext, hookData);
+    }
+
+    @Generated
```
chrfwow: What does the `@Generated` annotation do?

chrfwow (commented):
Do we have plans for e2e tests?

src/main/java/dev/openfeature/sdk/SharedHookContext.java
alexandraoberaigner: exclude the function from codecov

chrfwow: Should we put it then on methods that are not generated? Is this a recommended thing to do by codecov?
If yes, we should add a comment explaining this, otherwise I am sure I will have forgotten the next time I check the code and remove those annotations

aepfli: We should use a different annotation, `@Generated` conveys the message, that this method was auto generated, and not manually created. Hence people might refrain from editing it, although needed. Hence i suggest to use our `@ExcludeFromGeneratedCoverageReport` to exclude it.

```suggestion
    @ExcludeFromGeneratedCoverageReport
```

alexandraoberaigner: > Should we put it then on methods that are not generated?

@chrfwow Which ones do you mean? I put it on the delomboked functions

src/main/java/dev/openfeature/sdk/HookSupportData.java
alexandraoberaigner: @chrfwow Can I resolve this conv?

chrfwow: Yes you can. I would like to see a test where this really helps, I am not convinced yet

src/main/java/dev/openfeature/sdk/SharedHookContext.java
chrfwow: >  I put it on the delomboked functions

But now they are not generated any more in the sense that lombok will (re)create them for you if you change things. Therefore they should be tested, as they are now part of our normal code base. Whether to test getters/setters is another question, but they will be covered implicitely through other tests

src/main/java/dev/openfeature/sdk/HookSupport.java
alexandraoberaigner: as discussed: we leave the behaviour as before

src/main/java/dev/openfeature/sdk/HookContext.java
alexandraoberaigner: we need it because we need to expose a mutable structure

alexandraoberaigner:
> Do we have plans for e2e tests?

as discussed: will be a follow up task to add a gherkin test

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -222,9 +217,13 @@ private <T> FlagEvaluationDetails<T> evaluateFlag(
             }
             details.setErrorMessage(e.getMessage());
             enrichDetailsWithErrorDefaults(defaultValue, details);
-            hookSupport.errorHooks(type, afterHookContext, e, mergedHooks, hints);
+            if (hookSupportData.getHooks() != null && hookSupportData.getHints() != null) {
```
aepfli: as those checks seem to be specific to the hooksupport and the execution, should we move them into the hooksupport?

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -3,99 +3,108 @@
...
-    private <T> void executeHooks(
-            FlagValueType flagValueType, List<Hook> hooks, String hookMethod, Consumer<Hook<T>> hookCode) {
-        if (hooks != null) {
-            for (Hook hook : hooks) {
-                if (hook.supportsFlagValueType(flagValueType)) {
-                    executeChecked(hook, hookCode, hookMethod);
-                }
+            Optional<EvaluationContext> returnedEvalContext = Optional.ofNullable(
+                            hook.before(hookContext, data.getHints()))
+                    .orElse(Optional.empty());
```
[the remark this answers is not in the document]
alexandraoberaigner: If no one disagrees, I'd ignore this suggestion. A variable of type optional shouldn't be null but point to .empty()

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
alexandraoberaigner: I'd say no: with the split of the data and the logic the data's responsibility is on the side of the user of hooksupport & hooksupportdata, so its not the hooksupports fault if the data wasn't initialized properly because of any error that happened. Otherwise we'd need nullchecks for every access of HookSupportData

alexandraoberaigner: EDIT: its now only null checking hooks (after gemini suggestion applied): 
`if (hookSupportData.getHooks() != null) {`

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
chrfwow: I still think we should do this, otherwise we might miss calling hooks in the error stage

src/main/java/dev/openfeature/sdk/HookSupport.java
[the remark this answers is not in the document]
alexandraoberaigner: Pushed the change by accident and saw that Sonar even complains about this change with `Ensure this "Optional" could never be null and remove this null-check.`

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -3,99 +3,125 @@
...
+     * in the given data object with a new {@link HookData} instance.
+     *
+     * @param hookSupportData   the data object to modify
+     * @param sharedContext     the shared context from which the new {@link HookContext} is created
+     */
+    public void setHookContexts(HookSupportData hookSupportData, SharedHookContext sharedContext) {
+        for (int i = 0; i < hookSupportData.hooks.size(); i++) {
+            Pair<Hook, HookContext> hookContextPair = hookSupportData.hooks.get(i);
+            HookContext curHookContext = sharedContext.hookContextFor(null, new DefaultHookData());
+            Pair<Hook, HookContext> updatedPair = Pair.of(hookContextPair.getKey(), curHookContext);
```
chrfwow: Can we use a setter here on the Pair to avoid to create Pairs twice for each Hook?

chrfwow (commented):
lgtm, however, I would love to see a test that does a full flag evaluation using HookData

alexandraoberaigner:
> lgtm, however, I would love to see a test that does a full flag evaluation using HookData

Done, please take a look again if there is still anything missing :)

src/test/java/dev/openfeature/sdk/OpenFeatureClientTest.java
```diff
@@ -104,4 +106,33 @@ void shouldNotCallEvaluationMethodsWhenProviderIsInNotReadyState() {
...
         assertThat(details.getErrorCode()).isEqualTo(ErrorCode.PROVIDER_NOT_READY);
     }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    @DisplayName("Should support usage of HookData with/without error")
+    void shouldSupportUsageOfHookData(boolean isError) {
+        FeatureProvider provider = new TestEventsProvider(5000);
+        OpenFeatureAPI api = new OpenFeatureAPI();
+        if (isError) {
```
chrfwow, line 116: This is a very hacky way of getting a provider that will fail a flag evaluation. Couldn't you just use the `AlwaysBrokenWithExceptionProvider`?

alexandraoberaigner: I could also throw an exception 🤷

src/test/java/dev/openfeature/sdk/OpenFeatureClientTest.java
```diff
@@ -104,4 +106,34 @@ void shouldNotCallEvaluationMethodsWhenProviderIsInNotReadyState() {
...
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    @DisplayName("Should support usage of HookData with/without error")
+    void shouldSupportUsageOfHookData(boolean isError) {
+        OpenFeatureAPI api = new OpenFeatureAPI();
+        FeatureProvider provider;
+        if (isError) {
+            provider = new AlwaysBrokenWithExceptionProvider();
+        } else {
+            provider = new TestEventsProvider(5000);
```
chrfwow: Does that mean that the test will wait 5 sec before doing anything?

alexandraoberaigner: Jup... I found a better test provider to use

PR #1628, opened 2025-09-23
fix: improve vmlens handling

### LINKED ISSUE #1611: improve vmlens handling and execution
Vmlens is a powerful tool but our current developer experience is diminished, as it is executed all the times and increases the basic verify step heavily.

But especially our basic verify step should be fast and provide fast feedback. I suggest to move the vmlens to an own maven profile which is not executed per default. Additionally I suggest to also execute the vmlens profile as a separate GitHub action, to distinguish between the topics, and to give faster feedback on pull requests if it is a general testing compilation issue or if it is a concurrency issue.

Wdyt @chrfwow ?

ThomasKrieger:
## This PR

Renamed the VmLensTest to VmLensCT and changed the VMLens plugin config to include only tests with postfix CT.

### Related Issues

Fixes #1611

### Notes

The VMLens test no takes something around 4s.

### Follow-up Tasks

Please let me know if you prefer a special profile, or if the time is now o.k.

### DIFF
--- a/src/test/java/dev/openfeature/sdk/e2e/steps/ProviderSteps.java
+++ b/src/test/java/dev/openfeature/sdk/e2e/steps/ProviderSteps.java
@@ -25,9 +25,7 @@ import dev.openfeature.sdk.providers.memory.Flag;
 import dev.openfeature.sdk.providers.memory.InMemoryProvider;
 import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
-import java.time.Duration;
 import java.util.Map;
-import org.awaitility.Awaitility;
 
 public class ProviderSteps {
     private final State state;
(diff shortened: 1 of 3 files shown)


chrfwow:
@ThomasKrieger the DCO check fails because the commits on this PR are not signed. Please sign them off and force push them, the check should succceed then

src/test/java/dev/openfeature/sdk/e2e/steps/ProviderSteps.java
```diff
@@ -113,19 +112,15 @@ private void setupMockProvider(ErrorCode errorCode, String errorMessage, Provide
         switch (providerState) {
             case FATAL:
             case ERROR:
-                mockProvider.emitProviderReady(details);
-                mockProvider.emitProviderError(details);
+                mockProvider.emitProviderReady(details).await();
```
chrfwow, line 115: nice

PR #1634, opened 2025-09-25
fix: deployment failure because no tests were run by VMLens

ThomasKrieger:
fix: for #1611 and following build error

## This PR

moved to a new VMLens version which can handle no tests

### How to test
 mvn -DskipTests clean verify should run

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -32,7 +32,7 @@ jobs:
             cache: maven
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@e4b85ab654a244f420e4d9a9ca6305cf2d3fdc90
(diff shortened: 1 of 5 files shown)


chrfwow:
@ThomasKrieger why were no vmlens tests run, and don't we want the build to fail if no tests were run (as this could indicate a problem with out test setup)?

ThomasKrieger:
@chrfwow this was during the release.yml github step deploy line 63:
          mvn --batch-mode \
            --settings release/m2-settings.xml -DskipTests clean deploy
            
Which led to the build failure for the last pipeline.
I will check what is happening if you leave out -DskipTests and no tests are run

chrfwow:
I'll approve this as it is only a library update. Please re-request a review if you do change something in the github actions

ThomasKrieger:
@chrfwow I added https://github.com/vmlens/vmlens/issues/45 to fix this in VMLens

PR #1704, opened 2025-11-04
feat: use descriptive thread names

### LINKED ISSUE #1633: Use Descriptive Thread Names
Sometimes we are relying on threads, and create our owns. For debugging purposes it is really hard, if those threads do not have names. We should give our threads dedicated names.

jarebudev:
## This PR
Enables threads (daemon and non-daemon) to be named as required to aid debugging.

### Related Issues

Resolves #1633


toddbaert:
@jarebudev I removed the `!` from the title - because I don't think there's any breaking API changes here, right?

jarebudev:
> @jarebudev I removed the `!` from the title - because I don't think there's any breaking API changes here, right?

My mistake! Yes - no breaking API changes.

PR #1717, opened 2025-11-12
feat: Layered context to reduce memory churn

chrfwow:
Introduces layered context to reduce memory churn

Part of #1718

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -21,12 +21,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e
+    - uses: actions/checkout@c2d88d3ecc89a9ef08eebf45d9637801dcee7eb5
       with:
         fetch-depth: 0
         submodules: recursive
     - name: Set up JDK 17
-      uses: actions/setup-java@66b945764b75604b3cfd644c3ada5232cf6c90c6
+      uses: actions/setup-java@4e7e684fbb6e33f88ecb2cf1e6b3797739cf499b
       with:
         java-version: '17'
         distribution: 'temurin'
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -19,20 +19,20 @@ jobs:
     runs-on: ${{ matrix.os}}
     steps:
       - name: Check out the code
-        uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e
+        uses: actions/checkout@c2d88d3ecc89a9ef08eebf45d9637801dcee7eb5
         with:
           fetch-depth: 0
           submodules: recursive
 
       - name: Set up JDK ${{ matrix.build.java }}
-        uses: actions/setup-java@66b945764b75604b3cfd644c3ada5232cf6c90c6
+        uses: actions/setup-java@4e7e684fbb6e33f88ecb2cf1e6b3797739cf499b
         with:
             java-version: ${{ matrix.build.java }}
             distribution: 'temurin'
             cache: maven
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@ffa63f0dac7bf42e00061d972fa89c8c81288540
+        uses: github/codeql-action/init@0e52774aee3a99ebade1d021ab7e132c6f8058bf
         with:
           languages: java
 
@@ -58,4 +58,4 @@ jobs:
           verbose: true # optional (default = false)
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@ffa63f0dac7bf42e00061d972fa89c8c81288540
+        uses: github/codeql-action/analyze@0e52774aee3a99ebade1d021ab7e132c6f8058bf
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,13 +37,13 @@ jobs:
 
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e
+        uses: actions/checkout@c2d88d3ecc89a9ef08eebf45d9637801dcee7eb5
         with:
           fetch-depth: 0
           submodules: recursive
 
       - name: Set up JDK 17
-        uses: actions/setup-java@66b945764b75604b3cfd644c3ada5232cf6c90c6
+        uses: actions/setup-java@4e7e684fbb6e33f88ecb2cf1e6b3797739cf499b
         with:
           java-version: '17'
           distribution: 'temurin'
--- a/.github/workflows/static-code-scanning.yaml
+++ b/.github/workflows/static-code-scanning.yaml
@@ -29,16 +29,16 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e
+        uses: actions/checkout@c2d88d3ecc89a9ef08eebf45d9637801dcee7eb5
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@ffa63f0dac7bf42e00061d972fa89c8c81288540
+        uses: github/codeql-action/init@0e52774aee3a99ebade1d021ab7e132c6f8058bf
         with:
           languages: java
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@ffa63f0dac7bf42e00061d972fa89c8c81288540
+        uses: github/codeql-action/autobuild@0e52774aee3a99ebade1d021ab7e132c6f8058bf
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@ffa63f0dac7bf42e00061d972fa89c8c81288540
+        uses: github/codeql-action/analyze@0e52774aee3a99ebade1d021ab7e132c6f8058bf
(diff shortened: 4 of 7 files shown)


src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * LayeredEvaluationContext implements EvaluationContext by layering multiple contexts:
+ * API-level, Transaction-level, Client-level, Invocation-level, and Hook-level.
+ * The contexts are checked in that order for values, with Hook-level having the highest precedence.
+ */
+public class LayeredEvaluationContext implements EvaluationContext {
```
chrfwow, line 15: Alternatively, I could also implement this as a list of `EvaluationContext`s, with each list entry representing a context level (index 0 = api, index 1 = transaction, ...) This could maybe make the code cleaner (especially the merging, as I would just need to add another list entry; I don't think merge will/should be called often on this context though), but also less performant. Wdyt?

chrfwow:
Benchmark results:
(I ran `AllocationBenchmark#main` with the IntelliJ Profiler. This is not a proper JMH benchmark, so the values are not representative per evaluation, but should be helpful to compare both versions)

Baseline:
Duration: 5267 ms
Allocations: 16.27 GB
Flamegraph of allocations:
image



This PR:
Duration: 1649 ms
Allocations: 7.75 GB
Flamegraph of allocations:
image

src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -77,8 +62,10 @@ public void executeBeforeHooks(HookSupportData data) {
                             hook.before(hookContext, data.getHints()))
                     .orElse(Optional.empty());
             if (returnedEvalContext.isPresent()) {
```
aepfli, line 64: As we are using Optionals already, we should use all the available syntax sugar, and switch to `.ifPresent` and chain as much as possible

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
aepfli: so theoretically before hooks can also add information for the evaluation context - and this should also be somehow merged into this. would it make sense to simple add all the evaluationContext returned to the layer, and iterate over them as a list? 

Question if it is worth it - at least in theory it would be one less object

src/main/java/dev/openfeature/sdk/HookSupport.java
chrfwow: That would be nice, but `Optional`s in Java don't work well enough to get this with type safety. I would need some castings and redundant `isPresent` checks to get this to work

aepfli: but do we need the Optional wrapping on line 61? we want to reduce object churn etc. but we are generating a new Optional all the time - so i am wondering if we really need the Optional or if a null check would be sufficient here

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+        return Collections.unmodifiableMap(map);
+    }
+
+    @Override
+    public Map<String, Object> asObjectMap() {
+        var keySet = ensureKeySet();
+        var keys = keySet.size();
+        if (keys == 0) {
+            return new HashMap<>(0);
```
toddbaert, line 223:
```suggestion
            return Collections.emptyMap();
```

toddbaert: I think we can avoid this allocation, unless for some reason returning a singleton object is a bad idea.

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+        return getFromContext(apiContext, key);
+    }
+
+    @Override
+    public Map<String, Value> asMap() {
+        var keySet = ensureKeySet();
+        var keys = keySet.size();
+        if (keys == 0) {
+            return new HashMap<>(0);
```
toddbaert, line 180:
```suggestion
            return Collections.emptyMap();
```

Same as https://github.com/open-feature/java-sdk/pull/1717/files#r2543987740

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,235 @@
...
+        if (clientContext != null) {
+            keys.addAll(clientContext.keySet());
+        }
+        if (transactionContext != null) {
+            keys.addAll(transactionContext.keySet());
+        }
+        if (apiContext != null) {
+            keys.addAll(apiContext.keySet());
+        }
+        this.keySet = keys;
```
toddbaert, line 130: Should we make use of the `Collections.emptySet()` singleton while we have no keys?

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
toddbaert: On second thought, this may be a bad idea since people might expect to be able to add to this map, though l expect it's not that common 🤔

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
toddbaert: https://github.com/open-feature/java-sdk/pull/1717#discussion_r2544305406

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
chrfwow: We have a `asUnmodifiableMap` method, which implies that `asMap` returns something modifieable. I think this would lead to a hard to catch bug, if the map returned here sometimes is modifieable, and sometimes not.

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,235 @@
...
+        if (keys == 0) {
+            return new HashMap<>(0);
+        }
+        var map = new HashMap<String, Object>(keys);
+
+        for (String key : keySet) {
+            map.put(key, convertValue(getValue(key)));
+        }
+        return map;
+    }
```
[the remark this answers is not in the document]
chrfwow: That really helped a lot

src/main/java/dev/openfeature/sdk/HookSupport.java
chrfwow: I tried to change as little as possible of unrelated code so that we can scientifically see the benefit of the new context. Other optimizations will come in a later PR

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
chrfwow: I don't think I understand your comment, we only know that we have no keys at the end of this method

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
toddbaert: Agreed.

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
toddbaert: I was saying perhaps we could more lazily create a key set if there actually is any keys, and otherwise just return the empty set and avoid allocating something, but it's a micro-optimization.

chrfwow: But how do we know if we have any keys without generating the keyset? And after we generated it, we might as well return it, as we have allocated it anyway

toddbaert: It's the allocation of the `HashSet` on 107 that I think we can save, possibly, which is a heap object... I think we can do all the null checks and then only create this object if any of them are non null, otherwise just return the singleton, right? Maybe I'm missing something.

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
```diff
@@ -0,0 +1,254 @@
...
+        var otherTargetingKey = overridingContext.getTargetingKey();
+        if (otherTargetingKey != null) {
+            merged.targetingKey = otherTargetingKey;
+        }
+        return merged;
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return (hookContexts == null || hookContexts.isEmpty())
```
jarebudev: If the `hookContexts` collection contains at least one `EvaluationContext`, then a call to `isEmpty` will return false.  

But we do not check inside the `hookContexts` collection that all the `EvaluationContext`s that it contains are all empty.

should these also be checked for emptiness?

e.g. `(hookContexts == null || hookContexts.isEmpty() || hookContexts.stream().allMatch(ec -> ec.isEmpty()))`

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
jarebudev: could  the `isEmpty()` method be used to determine whether to create the `HashSet`, or if true, return the singleton?

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
chrfwow: Nice catch!

toddbaert:
I'll merge this tomorrow unless I hear objections.

Issue #1718, opened 2025-11-12
Layered Evaluation Context


chrfwow:
As suggested by @guidobrei , an even better solution to the context memory allocation problem (see #1588) would be to have a layered evalution context, that has a reference to api, transaction, client, invocation and hook contexts. 
When this layered context is queried for a value, it will search through the contexts as per the specified order, and does not need to merge any contexts, thereby greatly reducing memory churn.

An as of yet untested implementation can be found in this draft PR: https://github.com/open-feature/java-sdk/pull/1717
According the IntelliJ Profiler running on `AllocationBenchmark#main`, the overall memory footprint of flag evaluations is approximately halfed (take these numbers with caution, I am not sure if this imlpementation hasn't broken anything)

Note: while browsing through the code, I discovered a few more potential opportunities to reduce our memory footprint which should be tackled separately, as this problem is the main driver of memory churn

PR #1734, opened 2025-11-20
feat: Cache hooks per Flag Evaluation Type

chrfwow:
Hooks are now stored in a map to allow efficient lookup of hooks supporting certain FlagValueTypes. This and other changes makes merging the hooks cheaper.

### Related Issues
Part of #1718 

### Follow Up Tasks

Implement efficient Hook Storage in all providers (optional, but will speed things up)


chrfwow:
- [ ] Make sure the additions of default methods to interfaces is not a byte code incompatibility

chrfwow:
Benchmark results:
(I ran AllocationBenchmark#main with the IntelliJ Profiler. This is not a proper JMH benchmark, so the values are not representative per evaluation, but should be helpful to compare both versions)

Baseline:
Duration: 5267 ms
Allocations: 16.834 GB
Flamegraph of allocations:
image


This PR:
Duration: 5023 ms (this is an insignificant change, the deviations on those tests are quite large)
Allocations: 14.224 GB
Flamegraph of allocations:
image

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -313,16 +326,21 @@ public void addHooks(Hook... hooks) {
      * @return A list of {@link Hook}s.
      */
     public List<Hook> getHooks() {
-        return new ArrayList<>(this.apiHooks);
+        // Hooks can be duplicated if they support multiple FlagValueTypes
```
aepfli: I am not sure, the order of hooks is important, i know that you are now using other methods, to retrieve the needed hooks, but can this implementation now be a problem, because we are not keeping the order of the hooks. Now we merge multiple hook source after each other.

src/main/java/dev/openfeature/sdk/OpenFeatureClient.java
```diff
@@ -125,7 +130,16 @@ public void track(String trackingEventName, EvaluationContext context, TrackingE
      */
     @Override
     public OpenFeatureClient addHooks(Hook... hooks) {
-        this.clientHooks.addAll(Arrays.asList(hooks));
+        var types = FlagValueType.values();
```
aepfli: [suggestion] seems like this block of code repeats itself, should this maybe be a static helper function?

aepfli (approved):
i am torn a little bit and not 100% convinced, that this will make things easier. maybe we should think about v2 and rethink our hooksupport api overall.

But i also do not have big remarks to add

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
chrfwow: I could also store the hooks in a list in addition to the map, this way we would have both efficient hook lookup and ordered hooks in this method, but would waste a bit more memory. But I am not sure if returning unordered hooks here really is a problem

aepfli (approved):

aepfli (approved):

PR #873, opened 2024-03-28
fix: shutdown method blocks until task executor shutdown completes

jarebudev:
## This PR
is to ensure the shutdown method waits until the task executor (which is a daemon thread) has either completed shutdown or the timeout period elapses.

### Related Issues

Fixes #683

### Notes
I set the period to wait for completion to 3 seconds

### Follow-up Tasks
n/a

### How to test
n/a

PR #1744, opened 2025-11-26
fix: allow for providers to safely shutdown

nicklasl:
## Summary

Improves the shutdown behaviour of the `ProviderRepository` to ensure safe termination of the task executor. The previous implementation called `shutdown()` on the executor but did not wait for termination, potentially leading to abrupt shutdown, aborted provider shutdown sequences or hanging threads.

## Changes

  - Added graceful shutdown handling in `ProviderRepository.shutdown()` with a 3-second timeout
  - Added shutdown state tracking to prevent race conditions
  - Handles `RejectedExecutionException` when tasks are submitted to a shut down executor

## Implementation Details

  The shutdown process now:
1. Calls `shutdown()` on the task executor to prevent new tasks
2. Waits up to 3 seconds for running tasks to complete via `awaitTermination()`
3. If tasks don't complete within the timeout, calls `shutdownNow()` to force termination
4. Properly handles `InterruptedException` by calling `shutdownNow()` and restoring the interrupt flag
5. Prevents new providers from being registered during shutdown via `AtomicBoolean` flag
6. Handles `RejectedExecutionException` by running provider shutdown synchronously when executor is closed

## Test Coverage

Added comprehensive test coverage for shutdown behavior:
  [x] Successful shutdown when executor terminates within timeout
  [x] Forced shutdown when executor does not terminate within timeout
  [x] Graceful handling of interruption during shutdown
  [x] Protection against indefinite hangs during shutdown
  [x] Concurrent shutdown calls (idempotent behavior)
  [x] Shutdown during provider initialization
  [x] Provider replacement during shutdown
  [x] Prevention of new provider registration after shutdown starts

## Related Issues

https://github.com/open-feature/java-sdk/issues/1745

This change aligns with the broader effort to improve shutdown behaviour across the SDK, similar to PR #873 which addresses the same issue on Eventing.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -330,7 +330,7 @@ class ProviderRepository {
         managersToShutdown.forEach(this::shutdownProvider);
         taskExecutor.shutdown();
         try {
-            if (!taskExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
+            if (!taskExecutor.awaitTermination(EventSupport.SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
                 log.warn("Task executor did not terminate before the timeout period had elapsed");
                 taskExecutor.shutdownNow();
             }
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -234,9 +234,11 @@ class ProviderRepository {
     }
 
     private void shutDownOld(FeatureProviderStateManager oldManager, Consumer<FeatureProvider> afterShutdown) {
-        if (oldManager != null && !isStateManagerRegistered(oldManager)) {
-            shutdownProvider(oldManager);
-            afterShutdown.accept(oldManager.getProvider());
+        synchronized (registerStateManagerLock) {
+            if (oldManager != null && !isStateManagerRegistered(oldManager)) {
+                shutdownProvider(oldManager);
+                afterShutdown.accept(oldManager.getProvider());
+            }
         }
     }
 
@@ -289,20 +291,42 @@ class ProviderRepository {
      * including the default feature provider.
      */
     public void shutdown() {
-        List<FeatureProviderStateManager> managersToShutdown;
+        List<FeatureProviderStateManager> managersToShutdown = prepareShutdown();
+        if (managersToShutdown != null) {
+            completeShutdown(managersToShutdown);
+        }
+    }
 
+    /**
+     * Prepares the repository for shutdown by marking it as shutting down and
+     * collecting all managers that need to be shut down.
+     *
+     * <p>After this call, any attempt to set a provider will throw IllegalStateException.
+     *
+     * @return list of managers to shut down, or null if shutdown was already initiated
+     */
+    List<FeatureProviderStateManager> prepareShutdown() {
         synchronized (registerStateManagerLock) {
             if (isShuttingDown.getAndSet(true)) {
-                return;
+                return null;
             }
 
-            managersToShutdown = Stream.concat(
+            List<FeatureProviderStateManager> managersToShutdown = Stream.concat(
                             Stream.of(this.defaultStateManger.get()), this.stateManagers.values().stream())
                     .distinct()
                     .collect(Collectors.toList());
             this.stateManagers.clear();
+            return managersToShutdown;
         }
+    }
 
+    /**
+     * Completes the shutdown by shutting down all providers and waiting for
+     * pending tasks to complete.
+     *
+     * @param managersToShutdown the managers to shut down (from prepareShutdown)
+     */
+    void completeShutdown(List<FeatureProviderStateManager> managersToShutdown) {
         managersToShutdown.forEach(this::shutdownProvider);
         taskExecutor.shutdown();
         try {
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
@@ -277,5 +278,14 @@ class ProviderRepository {
                 .forEach(this::shutdownProvider);
         this.stateManagers.clear();
         taskExecutor.shutdown();
+        try {
+            if (!taskExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
+                log.warn("Task executor did not terminate before the timeout period had elapsed");
+                taskExecutor.shutdownNow();
+            }
+        } catch (InterruptedException e) {
+            taskExecutor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
     }
 }
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -11,6 +11,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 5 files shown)


aepfli (approved):

nicklasl:
I added a commit to additionally prevent race conditions during shutdown:
  - Added `AtomicBoolean isShuttingDown` flag to track repository shutdown state
  - `setProvider()` now throws `IllegalStateException` if called during shutdown
  - Handles `RejectedExecutionException` when shutdown tasks try to submit to closed executor
  - Ensures idempotent shutdown (multiple concurrent calls are safe)

src/test/java/dev/openfeature/sdk/ProviderRepositoryTest.java
```diff
@@ -289,6 +291,191 @@ void shouldRunLambdasOnError() throws Exception {
...
+            }
+
+            @Test
+            @DisplayName("should handle interruption during shutdown gracefully")
+            void shouldHandleInterruptionDuringShutdownGracefully() throws Exception {
+                FeatureProvider provider = createMockedProvider();
+                setFeatureProvider(provider);
+
+                Thread shutdownThread = new Thread(() -> {
+                    providerRepository.shutdown();
```
chrfwow: How long does the Mocked Provider take to shutdown? Are we sure we interrupt during this time?
Maybe this is a case for a [VMLens](https://github.com/open-feature/java-sdk/blob/main/src/test/java/dev/openfeature/sdk/vmlens/VmLensCT.java) test

src/test/java/dev/openfeature/sdk/ProviderRepositoryTest.java
```diff
@@ -289,6 +291,191 @@ void shouldRunLambdasOnError() throws Exception {
...
+            @Test
+            @DisplayName("should handle concurrent shutdown calls gracefully")
+            void shouldHandleConcurrentShutdownCallsGracefully() {
+                FeatureProvider provider = createMockedProvider();
+                setFeatureProvider(provider);
+
+                assertThatCode(() -> {
+                            Thread t1 = new Thread(() -> providerRepository.shutdown());
+                            Thread t2 = new Thread(() -> providerRepository.shutdown());
+                            Thread t3 = new Thread(() -> providerRepository.shutdown());
```
chrfwow: We cannot be sure that any of these threads execute in parallel. This should be a [VMLens](https://github.com/open-feature/java-sdk/blob/main/src/test/java/dev/openfeature/sdk/vmlens/VmLensCT.java) test

chrfwow (commented):
The implementation looks nice, but the tests may have some race conditions

src/test/java/dev/openfeature/sdk/ProviderRepositoryTest.java
nicklasl: Thanks, I'll look into making them VMLens tests instead, just need to find some time :)

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -272,10 +290,23 @@ private void shutdownProvider(FeatureProvider provider) {
...
     public void shutdown() {
+        if (isShuttingDown.getAndSet(true)) {
+            return;
+        }
+
         Stream.concat(Stream.of(this.defaultStateManger.get()), this.stateManagers.values().stream())
                 .distinct()
                 .forEach(this::shutdownProvider);
         this.stateManagers.clear();
         taskExecutor.shutdown();
```
chrfwow, line 331: I wonder if we have to wrap this with `synchronized (registerStateManagerLock) {`, otherwise we might lose additions to `this.stateManagers`

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
```diff
@@ -0,0 +1,189 @@
...
+    private FeatureProvider createMockedProvider(String name, AtomicInteger shutdownCounter) {
+        FeatureProvider provider = mock(FeatureProvider.class);
+        when(provider.getMetadata()).thenReturn(() -> name);
+        doAnswer(invocation -> {
+            shutdownCounter.incrementAndGet();
+            return null;
+        }).when(provider).shutdown();
+        try {
+            doAnswer(invocation -> null).when(provider).initialize(any());
+        } catch (Exception e) {
```
chrfwow: Why do we need this try/catch?

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
```diff
@@ -0,0 +1,189 @@
...
+                Thread t2 = new Thread(api::shutdown);
+                Thread t3 = new Thread(api::shutdown);
+
+                t1.start();
+                t2.start();
+                t3.start();
+
+                t1.join();
+                t2.join();
+                t3.join();
```
chrfwow:
```suggestion
                Runner.runParalell(api::shutdown,api::shutdown,api::shutdown);
```
Idk if this is the exact right syntax, but VmLens provides something like this, which handles starting and stopping the threads

nicklasl:
@chrfwow I've been working on the shutdown behavior fix for ProviderRepository a bit on the side and ran into a VMLens issue I could use help with.

I wrote a CT test (ProviderRepositoryCT) to verify concurrent shutdown behavior, but VMLens crashes when ThreadPoolExecutor.shutdown() is called inside AllInterleavings:

```
  NullPointerException: Cannot invoke "TLinkedList.iterator()" because "tasks" is null
    at com.vmlens...ThreadPoolMap.joinAll()
    at java.util.concurrent.ThreadPoolExecutor.shutdown()
    at dev.openfeature.sdk.EventSupport.shutdown()
```

Claude tells me that VMLens instruments `ThreadPoolExecutor` and its internal state gets corrupted when `shutdown()` runs during test execution. The existing `VmLensCT` avoids this by calling `shutdown()` in `@AfterEach` (outside `AllInterleavings`), but that defeats the purpose of testing concurrent shutdown.

I've commented out the tests for now but I'm wondering if you've encountered this before? Any ideas for testing concurrent shutdown with VMLens, or should we rely on the regular unit tests?

chrfwow:
@nicklasl I opened an issue for the NPE in VmLens: https://github.com/vmlens/vmlens/issues/58
If it is not too much work, I would keep the old tests that you had before VmLens, and update to the current version as soon as the issue is fixed. Speaking from experience, VmLens bugs are typically fixed quickly.

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
nicklasl: yeah, these are old mistakes.

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
nicklasl: we don't, fixed.

chrfwow:
@nicklasl the bug in the VmLens Library should now be fixed. Could you please re-enable the VMLens tests?

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
```diff
@@ -0,0 +1,137 @@
...
+        try (AllInterleavings allInterleavings =
+                new AllInterleavings("Concurrent API shutdown - provider called once")) {
+            while (allInterleavings.hasNext()) {
+                // Fresh state for each interleaving
+                AtomicInteger shutdownCount = new AtomicInteger(0);
+                FeatureProvider provider = createMockedProvider("test-provider", shutdownCount);
+                OpenFeatureAPI api = OpenFeatureAPITestUtil.createAPI();
+
+                // Set provider and wait for initialization to complete
+                api.setProviderAndWait(provider);
```
chrfwow, line 50: Would it make sense to copy this test but to call `setProvider` (Without wait) instead in the copy?

chrfwow (approved):
Approving, but please consider my comment

src/test/java/dev/openfeature/sdk/vmlens/ProviderRepositoryCT.java
nicklasl: Good point. I added an additional test with this 👍

toddbaert:
I'll do a final review on this and merge if I don't find any serious issues.

toddbaert:
I added one minor change to use an existing constant.

victornoel:
@nicklasl it's not clear to me if this change prevents the following problem:

- calling setProvider will trigger a shutdown then a load of the provider
- but the shutdown happens asynchronously: because it's in its own thread and we don't wait for the future to finish before continuing
- so the provider may be initialized again BEFORE the shutdown happens

This is a problem, e.g., for Unleash, where they will log an error about duplicate instances or even fail if some settings are set (which would be a think I would want to do).

nicklasl:
Hmm. I don't think this PR would handle that case. This was about the provider repository completely shutting down while your case is about replacing the provider, right?

I don't know it well enough but does the spec specify the ordering here?

victornoel:
@nicklasl ok, I guess I will first try with 1.20.2 to see if I can reproduce my problem and then I will open an issue if not :)

victornoel:
@nicklasl indeed, this change does not fix my problem, though I think it solved a few side effects I had before that would create more errors due to the concurrent execution so it's still useful :) I created https://github.com/open-feature/java-sdk/issues/1934

PR #1749, opened 2025-11-27
test: Test for breaking changes

chrfwow:
A crude POC to add a test that will fail when a breaking change is introduced

DO NOT MERGE!!!!!!!


src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -112,7 +112,7 @@ public Awaitable emit(final ProviderEvent event, final ProviderEventDetails deta
      *
      * @param details The details of the event
      */
-    public Awaitable emitProviderReady(ProviderEventDetails details) {
+    /*public*/ Awaitable emitProviderReady(ProviderEventDetails details) {
```
chrfwow, line 115:
```suggestion
    public Awaitable emitProviderReady(ProviderEventDetails details) {
```
Revert. This is my breaking change which I want to catch with this test

PR #1756, opened 2025-11-29
fix: equivalent EvaluationContext .equals() not returning true

### LINKED ISSUE #1754: Cannot compare (`.equals`) various context implementaations
A user reported an issue where they cannot compare instances of `EvaluationContext` as expected. This came about with our recent introduction of the [LayeredEvaluationContext](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java).

> I guess we cannot rely on evaluation context equality anymore but have to e.g. rely on its asMap equality.

This isn't strictly a bug, since I don't think we ever specified this behavior, but I think it should be easy to implement and nice-to-have.

I think we should override `.hashcode` etc to enable this sort of thing, and add tests to prevent regressions.

Vishalup29:
## Description

This PR adds consistent `equals` and `hashCode` implementations to the SDK’s
`EvaluationContext` implementations (`ImmutableContext` and `LayeredEvaluationContext`).

A user reported in Issue #1754 that comparing context instances did not behave as expected,
especially after the introduction of `LayeredEvaluationContext`. Because the SDK does not
guarantee object identity, equality should be defined based on the resolved attribute map
returned by `asMap()`.

This PR implements that behavior and adds regression tests to ensure we maintain consistent
equality semantics going forward.

## Changes

- Added `equals(Object)` and `hashCode()` to:
  - `ImmutableContext`
  - `LayeredEvaluationContext`
- Equality is now defined in terms of their `asMap()` representations.
- Added unit tests validating:
  - Equality for two identical `ImmutableContext` instances.
  - Equality for two identical `LayeredEvaluationContext` instances.
  - Map-based equality between `ImmutableContext` and `LayeredEvaluationContext`.

## Related Issue

Fixes #1754

## Notes

- Equality is intentionally based on `asMap()` rather than structural depth or source layering.
- This aligns the context behavior with user expectations and simplifies future comparisons.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
+++ b/src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
@@ -266,7 +266,7 @@ public class LayeredEvaluationContext implements EvaluationContext {
             return this.getResolvedMap().equals(((LayeredEvaluationContext) that).getResolvedMap());
         }
 
-        return this.getResolvedMap().equals(that.asMap());
+        return this.getResolvedMap().equals(that.asUnmodifiableMap());
     }
 
     @Override
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -114,7 +114,7 @@ public final class ImmutableContext implements EvaluationContext {
             return false;
         }
         EvaluationContext that = (EvaluationContext) o;
-        return this.asMap().equals(that.asMap());
+        return this.asUnmodifiableMap().equals(that.asUnmodifiableMap());
     }
 
     /**
@@ -130,7 +130,7 @@ public final class ImmutableContext implements EvaluationContext {
             synchronized (this) {
                 result = cachedHashCode;
                 if (result == null) {
-                    result = asMap().hashCode();
+                    result = asUnmodifiableMap().hashCode();
                     cachedHashCode = result;
                 }
             }


toddbaert:
@Vishalup29 Gemini's comments here are accurate, I think. Lazily building and caching this map is probably a good idea, instead of doing it again with every call... Though I think there's memory trade offs here as well...

The comment about the Lombok impl is also right, I think that needs to be removed. 

cc @chrfwow because I'd like your insights here as well. What do you think? I think either way we'll generate a heap object doing 'asmap' during 'hashcode', so we might as well cache it in case we need it again, right? I think this is optimal as long as we do it lazily?

chrfwow:
Yes, we should definetly cache the map when we create it. I have to say I don't really like that the first invocation of `equals` or `hashcode` will be quite expensive, but if we want to provide this assurance, I don't see a way around it.
We should also keep in mind that the `asMap` implementation of `ImmutableContext` (`ImmutableStructure`) is quite expensive, so maybe we can cache this as well

toddbaert:
> Yes, we should definetly cache the map when we create it. I have to say I don't really like that the first invocation of `equals` or `hashcode` will be quite expensive, but if we want to provide this assurance, I don't see a way around it

@chrfwow 

Yes, I feel exactly the same way. It will be expensive, and allocate memory once, but I think it's still worth having. At least the use case we know of it was just used in test assertions - specifically to validate/assert in a test that a context was set... perhaps this will be the most common use-case.

Vishalup29:
Thanks for the detailed feedback!

- Updated `ImmutableContext` to cache its hash code lazily (and removed the Lombok `@EqualsAndHashCode` annotation). 
- Refactored `LayeredEvaluationContext` to lazily cache the resolved attribute map via a new `getResolvedMap()` helper, with invalidation in `putHookContext. asMap()`, `equals`, and `hashCode` now all use this helper.
- Kept equality semantics based on `asMap()` as discussed and expanded tests to cover mixed-context and layered equality including `hashCode`.

cc: @toddbaert @chrfwow Thanks again for your review. Really appreciate it

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
chrfwow: Please use this helper function instead: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/HashMapUtils.java#L8

chrfwow: I wonder if we should create the object map from our reference map

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -145,10 +148,31 @@ void mergeShouldObtainKeysFromOverridingContextWhenExistingContextIsEmpty() {
...
+        ImmutableContext a = new ImmutableContext(null, map1);
+        ImmutableContext b = new ImmutableContext(null, map2);
+
+        assertEquals(a, b);
+        assertEquals(a.hashCode(), b.hashCode());
+
+        // hashCode should be stable across multiple invocations (caching behavior)
+        int first = a.hashCode();
+        int second = a.hashCode();
+        assertEquals(first, second);
```
chrfwow: I think this should be its own test

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
```diff
@@ -145,10 +148,31 @@ void mergeShouldObtainKeysFromOverridingContextWhenExistingContextIsEmpty() {
...
         EvaluationContext overriding = new ImmutableContext(attributes);
         EvaluationContext merge = ctx.merge(overriding);
-        assertEquals(new java.util.HashSet<>(java.util.Arrays.asList("key1", "key2")), merge.keySet());
+        assertEquals(new HashSet<>(Arrays.asList("key1", "key2")), merge.keySet());
     }
 
-    @DisplayName("Two different MutableContext objects with the different contents are not considered equal")
+    @DisplayName("Two ImmutableContext objects with identical attributes are considered equal")
+    @Test
+    void testImmutableContextEquality() {
```
chrfwow, line 156: We have a very similar test in line 190

chrfwow (commented):
Looks very good! I just have some minor comments

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
Vishalup29: moved it to a new test

src/test/java/dev/openfeature/sdk/ImmutableContextTest.java
Vishalup29: added a new test instead

src/main/java/dev/openfeature/sdk/LayeredEvaluationContext.java
Vishalup29: done

Vishalup29: done

Vishalup29:
@chrfwow Addressed all your comments. 
Thanks again for the clear review comments — really appreciate the help getting this contribution into good shape. 🙂

Vishalup29:
thanks, can this be merged?
cc: @toddbaert

PR #1760, opened 2025-12-02
fix: possible StackOverflow on recursive contexts

chrfwow:
## This PR
The new `LayeredEvaluationContext` can result in a `StackOverflowException` when a before hook returns the passed hook context. This would result in the `LayeredEvaluationContext` having a reference to itself in the hook context layer. As soon as `toMap`, `toObjectMap` or similar is called, an exception will be thrown.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -26,7 +26,7 @@ jobs:
         fetch-depth: 0
         submodules: recursive
     - name: Set up JDK 17
-      uses: actions/setup-java@4e7e684fbb6e33f88ecb2cf1e6b3797739cf499b
+      uses: actions/setup-java@f2beeb24e141e01a676f977032f5a29d81c9e27e
       with:
         java-version: '17'
         distribution: 'temurin'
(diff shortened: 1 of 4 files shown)


src/main/java/dev/openfeature/sdk/HookSupport.java
```diff
@@ -63,7 +63,8 @@ public void executeBeforeHooks(HookSupportData data) {
                     .orElse(Optional.empty());
             if (returnedEvalContext.isPresent()) {
                 var returnedContext = returnedEvalContext.get();
-                if (!returnedContext.isEmpty()) {
+                // yes, we want to check for reference equality here, this prevents recursive layered contexts
+                if (returnedContext != hookContext.getCtx() && !returnedContext.isEmpty()) {
```
alexandraoberaigner, line 67:
```suggestion
                if (!returnedContext.isEmpty() && returnedContext != hookContext.getCtx()) {
```
[Q] would java optimize this anyway or could this be a slight improvement checking for empty first?

chrfwow: I don't think Java would touch the evaluation order here, but I would have to check to make sure.

I assumed that the getter + equality check will be faster than the `isEmpty()`, which has to access the underlying Map, but I didn't benchmark it.

toddbaert: I suspect that as well.

PR #1765, opened 2025-12-08
feat: add multi-provider to SDK

Vishalup29:
## Description

Move the existing multi-provider implementation from the contrib module into the core Java SDK, mark it as experimental, and deprecate the contrib implementation.

## Changes

- Move `MultiProvider` (and related strategy/utility classes) from `java-sdk-contrib/providers/multiprovider` into the core SDK under `dev.openfeature.sdk.providers.multiprovider`.
- Mark the SDK multi-provider as **experimental** using inline Javadoc and README documentation.
- Add SDK tests for the multi-provider based on the existing contrib tests.
- Update the Java SDK `README.md` to document the multi-provider feature and show a basic usage example.
- Deprecate the contrib multi-provider implementation and README, pointing users to the SDK implementation.
- Adjust exception usage to match the SDK API (use `new FlagNotFoundError()` rather than a message constructor).

## Related Issues

Fixes open-feature#1486

## Notes

- In the SDK, `FlagNotFoundError` only exposes a no-arg constructor, so the ported multi-provider now throws `new FlagNotFoundError()` where the contrib version used a message.
- Metadata handling has been aligned with the SDK’s existing patterns and dependency constraints to avoid introducing unnecessary dependencies.

### DIFF
--- a/src/main/java/dev/openfeature/sdk/multiprovider/FirstMatchStrategy.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/FirstMatchStrategy.java
@@ -1,0 +1,63 @@
+package dev.openfeature.sdk.multiprovider;
+
+import static dev.openfeature.sdk.ErrorCode.FLAG_NOT_FOUND;
+
+import dev.openfeature.sdk.ErrorCode;
+import dev.openfeature.sdk.EvaluationContext;
+import dev.openfeature.sdk.FeatureProvider;
+import dev.openfeature.sdk.ProviderEvaluation;
+import dev.openfeature.sdk.exceptions.FlagNotFoundError;
+import java.util.Map;
+import java.util.function.Function;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * First match strategy.
+ *
+ * <p>Return the first result returned by a provider.
+ * <ul>
+ *   <li>Skip providers that indicate they had no value due to {@code FLAG_NOT_FOUND}.</li>
+ *   <li>On any other error code, return that error result.</li>
+ *   <li>If a provider throws {@link FlagNotFoundError}, it is treated like {@code FLAG_NOT_FOUND}.</li>
+ *   <li>If all providers report {@code FLAG_NOT_FOUND}, return a {@code FLAG_NOT_FOUND} error.</li>
+ * </ul>
+ * As soon as a non-{@code FLAG_NOT_FOUND} result is returned by a provider (success or other error),
+ * the rest of the operation short-circuits and does not call the remaining providers.
+ */
+@Slf4j
+@NoArgsConstructor
+public class FirstMatchStrategy implements Strategy {
+
+    @Override
+    public <T> ProviderEvaluation<T> evaluate(
+            Map<String, FeatureProvider> providers,
+            String key,
+            T defaultValue,
+            EvaluationContext ctx,
+            Function<FeatureProvider, ProviderEvaluation<T>> providerFunction) {
+        for (FeatureProvider provider : providers.values()) {
+            try {
+                ProviderEvaluation<T> res = providerFunction.apply(provider);
+                ErrorCode errorCode = res.getErrorCode();
+                if (errorCode == null) {
+                    // Successful evaluation
+                    return res;
+                }
+                if (!FLAG_NOT_FOUND.equals(errorCode)) {
+                    // Any non-FLAG_NOT_FOUND error bubbles up
+                    return res;
+                }
+                // else FLAG_NOT_FOUND: skip to next provider
+            } catch (FlagNotFoundError ignored) {
+                // do not log in hot path, just skip
+            }
+        }
+
+        // All providers either threw or returned FLAG_NOT_FOUND
+        return ProviderEvaluation.<T>builder()
+                .errorMessage("Flag not found in any provider")
+                .errorCode(FLAG_NOT_FOUND)
+                .build();
+    }
+}
--- a/src/main/java/dev/openfeature/sdk/multiprovider/FirstSuccessfulStrategy.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/FirstSuccessfulStrategy.java
@@ -1,0 +1,48 @@
+package dev.openfeature.sdk.multiprovider;
+
+import dev.openfeature.sdk.ErrorCode;
+import dev.openfeature.sdk.EvaluationContext;
+import dev.openfeature.sdk.FeatureProvider;
+import dev.openfeature.sdk.ProviderEvaluation;
+import java.util.Map;
+import java.util.function.Function;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * First Successful Strategy.
+ *
+ * <p>Similar to “First Match”, except that errors from evaluated providers do not halt execution.
+ * Instead, it returns the first successful result from a provider. If no provider successfully
+ * responds, it returns a {@code GENERAL} error result.
+ */
+@Slf4j
+@NoArgsConstructor
+public class FirstSuccessfulStrategy implements Strategy {
+
+    @Override
+    public <T> ProviderEvaluation<T> evaluate(
+            Map<String, FeatureProvider> providers,
+            String key,
+            T defaultValue,
+            EvaluationContext ctx,
+            Function<FeatureProvider, ProviderEvaluation<T>> providerFunction) {
+        for (FeatureProvider provider : providers.values()) {
+            try {
+                ProviderEvaluation<T> res = providerFunction.apply(provider);
+                if (res.getErrorCode() == null) {
+                    // First successful result (no error code)
+                    return res;
+                }
+            } catch (Exception ignored) {
+                // swallow and continue; errors from individual providers
+                // are not fatal for this strategy
+            }
+        }
+
+        return ProviderEvaluation.<T>builder()
+                .errorMessage("No provider successfully responded")
+                .errorCode(ErrorCode.GENERAL)
+                .build();
+    }
+}
--- a/src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
@@ -1,0 +1,179 @@
+package dev.openfeature.sdk.multiprovider;
+
+import dev.openfeature.sdk.EvaluationContext;
+import dev.openfeature.sdk.EventProvider;
+import dev.openfeature.sdk.FeatureProvider;
+import dev.openfeature.sdk.Metadata;
+import dev.openfeature.sdk.ProviderEvaluation;
+import dev.openfeature.sdk.Value;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * <b>Experimental:</b> Provider implementation for multi-provider.
+ *
+ * <p>This provider delegates flag evaluations to multiple underlying providers using a configurable
+ * {@link Strategy}. It also exposes combined metadata containing the original metadata of each
+ * underlying provider.
+ */
+@Slf4j
+public class MultiProvider extends EventProvider {
+
+    @Getter
+    private static final String NAME = "multiprovider";
+
+    // Use CPU count as upper bound for init threads.
+    public static final int INIT_THREADS_COUNT = Runtime.getRuntime().availableProcessors();
+
+    private final Map<String, FeatureProvider> providers;
+    private final Strategy strategy;
+    private MultiProviderMetadata metadata;
+
+    /**
+     * Constructs a MultiProvider with the given list of FeatureProviders, by default uses
+     * {@link FirstMatchStrategy}.
+     *
+     * @param providers the list of FeatureProviders to initialize the MultiProvider with
+     */
+    public MultiProvider(List<FeatureProvider> providers) {
+        this(providers, new FirstMatchStrategy());
+    }
+
+    /**
+     * Constructs a MultiProvider with the given list of FeatureProviders and a strategy.
+     *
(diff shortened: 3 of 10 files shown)


Vishalup29:
@chrfwow @toddbaert Can you pls review when you get a chance?

aepfli:
Hi, thanks for your contribution and interest in the SDK! We genuinely appreciate your work on multi-provider support.

Since there's already a well-developed PR (#1500 ) addressing the same feature, having multiple PRs for the same goal creates extra effort for reviewers and can slow things down. Next time, we suggest building on the existing work - especially when significant development and review have already occurred.

A few differences to keep in mind:
- PR #1500 offers comprehensive tests, standardized exception handling, and has removed unnecessary dependencies.
- It includes extensive review feedback covering several improvements and edge cases.

Your contributions are always welcome, but i think basing the work on the previous PR would be a good approach here. Especially to also credit the previous contributor of his base work.

Thank you for your effort, can you maybe incorporate the already existing feedback from the other pull request, or re-base your efforts based on the other one. This would be highly appreciated.

Vishalup29:
@aepfli Rebased and PR on top of the changes in the PR (https://github.com/open-feature/java-sdk/pull/1500 ). it now includes all the suggestions based on the discussions on that PR. Pls take a look when you get a chance

src/test/java/dev/openfeature/sdk/multiprovider/BaseStrategyTest.java
```diff
@@ -0,0 +1,214 @@
+package dev.openfeature.sdk.multiProvider;
```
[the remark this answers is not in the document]
aepfli: seems like a valid reasoning, and we should rename the pacakge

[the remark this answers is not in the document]
Vishalup29: addressed

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @param providers the list of FeatureProviders to initialize the MultiProvider with
+     * @param strategy  the strategy (if {@code null}, {@link FirstMatchStrategy} is used)
+     */
+    public MultiProvider(List<FeatureProvider> providers, Strategy strategy) {
+        this.providers = buildProviders(providers);
+        if (strategy != null) {
+            this.strategy = strategy;
+        } else {
+            this.strategy = new FirstMatchStrategy();
```
guidobrei: Should we assume that `null` means the user wants to have a `FirstMatchStrategy`?

I'd rather prefer a `NullReferenceException` here and only use the `FirstMatchStrategy` in the single parameter `MultiProvider(List<FeatureProvider> providers)` constructor.
WDYT?

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+    }
+
+    protected static Map<String, FeatureProvider> buildProviders(List<FeatureProvider> providers) {
+        Map<String, FeatureProvider> providersMap = new LinkedHashMap<>(providers.size());
+        for (FeatureProvider provider : providers) {
+            FeatureProvider prevProvider =
+                    providersMap.put(provider.getMetadata().getName(), provider);
+            if (prevProvider != null) {
+                log.warn("duplicated provider name: {}", provider.getMetadata().getName());
```
guidobrei: Why do we want to prevent this? @aepfli, @chrfwow do we treat the provider name as unique identifier in the SDK?

And if, we should throw because it would be a clear misconfiguration.

Scenario where I could see the same provider twice is e.g., using flagd providers that should read from multiple flag configuration files. According to the spec, the [provider name](https://openfeature.dev/specification/sections/providers#requirement-211) identifies the implementation, not the instance.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * <b>Experimental:</b> Provider implementation for multi-provider.
+ *
+ * <p>This provider delegates flag evaluations to multiple underlying providers using a configurable
+ * {@link Strategy}. It also exposes combined metadata containing the original metadata of each
+ * underlying provider.
+ */
+@Slf4j
+public class MultiProvider extends EventProvider {
```
guidobrei, line 33: The `MultiProvider` does not emit any [lifecycle events](https://openfeature.dev/specification/sections/events). If we don't want to support this in the first iteration it is sufficient to implement `FeatureProvider` instead of extending `EventProvider`.

Although we should take care of this in a follow up asap, because we effectively loose the Event features when using `MultiProvider`.

src/main/java/dev/openfeature/sdk/multiprovider/FirstMatchStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+                    // Successful evaluation
+                    return res;
+                }
+                if (!FLAG_NOT_FOUND.equals(errorCode)) {
+                    // Any non-FLAG_NOT_FOUND error bubbles up
+                    return res;
+                }
+                // else FLAG_NOT_FOUND: skip to next provider
+            } catch (FlagNotFoundError e) {
+                log.debug(
```
guidobrei: We usually do not log in hot code paths like flag evaluations.

src/main/java/dev/openfeature/sdk/multiprovider/FirstSuccessfulStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+            Function<FeatureProvider, ProviderEvaluation<T>> providerFunction) {
+        for (FeatureProvider provider : providers.values()) {
+            try {
+                ProviderEvaluation<T> res = providerFunction.apply(provider);
+                if (res.getErrorCode() == null) {
+                    // First successful result (no error code)
+                    return res;
+                }
+            } catch (Exception e) {
+                log.debug(
```
guidobrei: We usually do not log in hot code paths but can see the value of having a debug message here.

guidobrei (commented):
I've gone over the Implementation once more and added some concerns to be discussed.
It does not block this PR which is about moving the provider from contrib to the SDK and afaik the issues currently exist in the contrib implementation as well. But maybe for the valid points identified we could craft some new issues.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: Thanks!! Good point, using null as a signal for the default strategy is confusing.
I’ve updated the constructors so that:

MultiProvider(List<FeatureProvider> providers) uses FirstMatchStrategy by default

MultiProvider(List<FeatureProvider> providers, Strategy strategy) now requires a non-null strategy and throws if it’s null
This keeps the simple ctor ergonomic while making the 2-arg ctor semantics explicit.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: That’s a really good question, and I agree it needs a clear spec-level answer.
For this PR I’ve kept the contrib behaviour (log a warning and let the last provider win) to avoid changing semantics while we move the implementation into the SDK.
I can open a follow-up issue to clarify whether provider names should be treated as unique identifiers or implementation identifiers, and we can update the behaviour there once we agree on the desired contract.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: Agreed that the current MultiProvider doesn’t emit events yet.
For now I’ve kept it extending EventProvider to avoid changing the surface again during the move, and to keep the door open for proper event propagation in a follow-up.
I’ll open a follow-up issue to track “Emit provider lifecycle/configuration events from MultiProvider” so we can design the right event semantics without blocking this migration PR.

src/main/java/dev/openfeature/sdk/multiprovider/FirstMatchStrategy.java
Vishalup29: Makes sense – logging inside the hot evaluation path isn’t aligned with how the rest of the SDK behaves.
I’ve removed the debug logging from both strategy implementations and now just skip on FLAG_NOT_FOUND / exceptions, relying on the individual providers (or hooks/logging hook) for diagnostics.

src/main/java/dev/openfeature/sdk/multiprovider/FirstSuccessfulStrategy.java
Vishalup29: Makes sense – logging inside the hot evaluation path isn’t aligned with how the rest of the SDK behaves.
I’ve removed the debug logging from both strategy implementations and now just skip on FLAG_NOT_FOUND / exceptions, relying on the individual providers (or hooks/logging hook) for diagnostics.

Vishalup29:
@chrfwow @toddbaert  @aepfli @guidobrei can this be approved/merged?

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
chrfwow: I don't think we want the names to be unique, but maybe it doesn't make sense to add the same provider twice to the multiprovider, and if we add telemetry which provider responded, this would make it unclear. But I believe an info log would be enough

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Getter
+    private static final String NAME = "multiprovider";
+
+    public static final int INIT_THREADS_COUNT = 8;
```
chrfwow: This seems a bit like an arbitrary number, maybe use `Runtime.getRuntime().availableProcessors();` instead, if it doesn't have some weird edge cases

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,183 @@
...
+     * Initialize the provider.
+     *
+     * @param evaluationContext evaluation context
+     * @throws Exception on error (e.g. wrapped {@link java.util.concurrent.ExecutionException}
+     *                   from a failing provider)
+     */
+    @Override
+    public void initialize(EvaluationContext evaluationContext) throws Exception {
+        var metadataBuilder = MultiProviderMetadata.builder().name(NAME);
+        HashMap<String, Metadata> providersMetadata = new HashMap<>();
```
chrfwow, line 88: Considering https://github.com/open-feature/java-sdk/pull/1765/changes#r2618228793, this should maybe be a KeyValuePair List

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -0,0 +1,183 @@
...
+            for (FeatureProvider provider : providers.values()) {
+                try {
+                    provider.shutdown();
+                } catch (Exception shutdownEx) {
+                    log.error(
+                            "error shutting down provider {} after failed initialize",
+                            provider.getMetadata().getName(),
+                            shutdownEx);
+                }
+            }
```
chrfwow: We could call `shutdown` here

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: Switched INIT_THREADS_COUNT to Runtime.getRuntime().availableProcessors() and still cap it with providers.size() when creating the thread pool.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: For now I kept originalMetadata as Map<String, Metadata> to match how buildProviders currently treats provider names as keys.
I agree a List<KeyValuePair<String, Metadata>> could better represent duplicate names; happy to open a follow-up issue to discuss that representation and adjust the metadata type there if we decide to support duplicates fully.

Vishalup29:
@chrfwow @toddbaert @aepfli @guidobrei can this be approved/merged?

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
chrfwow: Have you created the issue? I think this is something we should not forget

chrfwow:
@Vishalup29 you need to run the spotless plugin once more to fix the code format issues

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
Vishalup29: fixed all the issues. CI green

toddbaert: I think @chrfwow meant to open a new issue for the commend here. 

I've done this: https://github.com/open-feature/java-sdk/issues/1792

toddbaert (approved):
I've rebased, squashed, and added @suvaidkhan, the author of #1500 , as a co-author here, since some of their work went into this.

I'm approving, but I would love one more approval from @aepfli or @guidobrei

Vishalup29:
@chrfwow @toddbaert @aepfli @guidobrei - what is the process to get a PR merged in this repo? This PR has undergone multiple rounds of review and has also been approved multiple times. Who can merge this?

aepfli:
I am sorry, I have been out of service since the middle of December, and I will be back next Monday. - first sick and afterwards holiday season. 

Sorry for the delays on my side. But this time of the year is usually more focused on time off the computer or other side projects, rather than my usual dev life ;)

toddbaert:
> @chrfwow @toddbaert @aepfli @guidobrei - what is the process to get a PR merged in this repo? This PR has undergone multiple rounds of review and has also been approved multiple times. Who can merge this?

Hey @Vishalup29 ... I can merge it, but this took a while for a few reasons:

- Christmas means fewer people are working, so things are slower than usual.
- This is a 1000 line PR, and it's going into the main SDK, so thorough review is required.
- There's so-far unresolved comments from previous reviewers, so I would prefer to see them resolved by those people. If they can't soon, I will re-review myself once more, and merge.

As you can see, 5 days ago I made some changes on here myself, so don't worry, your efforts are not being ignored. We are just being thorough. :)

toddbaert:
@Vishalup29 I did a final re-review. Things look good. Thanks so much for your contribution and patience. Sorry that we were slow on this one. Happy new year. :christmas_tree:

Vishalup29:
Thanks @toddbaert and a very Happy New Year!

PR #1807, opened 2026-01-16
fix: Allow empty string as a valid targeting key

leoromanovsky:
## This PR

The OpenFeature spec defines targeting key as "a string" with no restriction on empty strings. The `TARGETING_KEY_MISSING` error code is for when a key is "not provided" (null), not when it's an empty string.

Previously, MutableContext rejected empty strings by checking `!targetingKey.trim().isEmpty()`, treating them the same as null. This made it impossible to distinguish between "not provided" and "explicitly set to empty string".

Changes:
- Remove isEmpty check from constructor and setTargetingKey()
- Empty string is now stored and returned by getTargetingKey()
- Null still results in no targeting key being set
- Updated tests to reflect new behavior
- Added explicit tests for empty string targeting key

### Related Issues

Fixes #1234523


src/test/java/dev/openfeature/sdk/MutableContextTest.java
```diff
@@ -59,6 +60,41 @@ void missingTargetingKeyShould() {
         assertEquals(null, ctx.getTargetingKey());
     }
 
+    @DisplayName("setTargetingKey with null should not set targeting key")
```
chrfwow, line 63: Please also add the tests for nullability to the `ImmutableContext`

leoromanovsky:
Thanks for the reviews team. @chrfwow or @aepfli please merge at your convenience.

PR #1883, opened 2026-03-06
Provider hook evaluation order

jonathannorris:
Addresses multi-provider gaps outlined in #1882 by implementing event aggregation, per-provider hook execution, tracking forwarding, and a new comparison strategy.

---
<p><a href="https://cursor.com/agents/bc-0eff3bb6-abdb-41a3-8afb-fca20965a722"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0eff3bb6-abdb-41a3-8afb-fca20965a722"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</p>

PR #1897, opened 2026-03-16
fix: multiprovider parity gaps with JS SDK reference (#1882)

### LINKED ISSUE #1882: [Multi-provider] Gaps identified relative to js-sdk reference implementation
## Context

We conducted a cross-SDK comparison of all MultiProvider implementations using the js-sdk as the reference. The Java MultiProvider was recently moved into the core SDK in #1765, and while functional for basic use cases, we identified several gaps relative to the reference implementation. Some of these were already noted during the original PR review.

### 1. Child provider event aggregation and status tracking (High)

The MultiProvider extends `EventProvider` but does not listen to or forward events from child providers. If a child provider emits `PROVIDER_ERROR`, `PROVIDER_STALE`, or `PROVIDER_CONFIGURATION_CHANGED` at runtime, those events are not surfaced. This was called out in the #1765 review by @guidobrei:
> "we effectively lose the Event features when using MultiProvider"

**Expected behavior:**
- Listen to each child provider's events
- Maintain a per-provider status map
- Compute an aggregate status using "worst-wins" precedence: `FATAL > NOT_READY > ERROR > STALE > READY`
- Emit the corresponding event when the aggregate status changes
- Always forward `PROVIDER_CONFIGURATION_CHANGED` events (pass-through)

**Reference:** js-sdk `status-tracker.ts`, dotnet-sdk `HandleProviderEventAsync` / `DetermineAggregateStatus`

### 2. Per-provider hook execution during evaluation (High)

The strategy calls provider evaluation methods directly (e.g. `provider.getBooleanEvaluation(...)`), bypassing the SDK's hook pipeline. If a child provider defines hooks via `getProviderHooks()`, those hooks are not executed.

**Expected behavior:**
- Before evaluating a child provider, run its `before` hooks with an isolated copy of the hook context
- On success: run `after` hooks
- On error: run `error` hooks
- Always: run `finally` hooks
- Hook context must be isolated per-provider to prevent cross-provider mutation

**Reference:** js-sdk `hook-executor.ts`, go-sdk `isolation.go`, dotnet-sdk `ProviderExtensions.EvaluateAsync`

### 3. Tracking event forwarding (High)

`track()` is not overridden. The default no-op implementation means tracking events are not forwarded to child providers.

**Expected behavior:**
- Iterate over child providers and forward `track()` calls
- The strategy should control which providers receive tracking (e.g. skip `NOT_READY` / `FATAL` providers)
- Errors from individual `track()` calls should be caught and logged, not propagated

**Reference:** js-sdk `multi-provider.ts` `track()`, dotnet-sdk `MultiProvider.cs` `Track()`

### 4. ComparisonStrategy (Medium)

Only `FirstMatchStrategy` and `FirstSuccessfulStrategy` exist. There is no `ComparisonStrategy` for evaluating all providers and comparing results (useful for migration validation and consistency checks).

**Expected behavior:**
- Evaluate all providers (ideally in parallel)
- If all providers agree on the value, return it
- If providers disagree, call an optional `onMismatch` callback and return the designated fallback provider's result
- If any provider errors, collect and report all errors
- Constructor accepts a `fallbackProvider` and optional `onMismatch` callback

**Reference:** js-sdk `comparison-strategy.ts`, go-sdk `comparison_strategy.go`, dotnet-sdk `ComparisonStrategy.cs`

### 5. Duplicate provider name handling (Medium)

When two child providers share the same `metadata.name`, the later provider silently overwrites the earlier one in the internal `LinkedHashMap`. The first provider is effectively lost. Related to #1792.

**Expected behavior:**
- If an explicit name conflicts with an existing name, throw (misconfiguration)
- If multiple providers share the same metadata-derived name, auto-deduplicate with a numeric suffix (`name-1`, `name-2`, etc.)
- All providers should be preserved

**Reference:** js-sdk `registerProviders()`, dotnet-sdk `RegisterProviders()`

## Spec Reference

https://openfeature.dev/specification/appendix-a/#multi-provider

jonathannorris:
## Summary

- Fixes multiprovider parity gaps identified in cross-SDK comparison against the js-sdk reference
- Decouples `OpenFeatureClient` from `MultiProvider` for hook context capture, uses a provider-level `before` hook with a `ThreadLocal`, matching the JS SDK's `WeakMap` pattern
- Adds provider state tracking, child provider event observation, and `track()` delegation to `MultiProvider`
- Adds `ComparisonStrategy`, parallel evaluation across all providers with configurable fallback and mismatch callback
- Adds test coverage for concurrent evaluation, provider hook lifecycle ordering, event/state aggregation, and `track()` forwarding

## Implementation

| Change | Details |
|--------|---------|
| **Context capture** | `MultiProvider.getProviderHooks()` returns a `before` hook that captures `ClientMetadata` and a defensive copy of the hook hints into a `ThreadLocal`; a `finallyAfter` hook cleans it up. Removes the `OpenFeatureClient` → `MultiProvider` import dependency. |
| **Provider state tracking** | Per-child `ProviderState` map with severity-based aggregate state (`FATAL > NOT_READY > ERROR > STALE > READY`). Registers/deregisters `EventProvider` observers to react to child provider state changes. `PROVIDER_CONFIGURATION_CHANGED` is always forwarded. |
| **Per-provider hooks** | `MultiProviderHookExecutor` runs each child provider's own hooks around its evaluation: `before` in registration order, `after`/`error`/`finallyAfter` in reverse, with an isolated context copy per provider. |
| **`ComparisonStrategy`** | Evaluates all providers in parallel via `ForkJoinPool.commonPool()` (or a caller-supplied executor). Returns the fallback provider's result on agreement; invokes the optional `onMismatch` callback on disagreement. On failure returns a `MultiProviderEvaluation` carrying per-provider `ProviderError` details, consistent with `FirstMatchStrategy` / `FirstSuccessfulStrategy`. |
| **`track()`** | Forwards to all child providers, skipping `NOT_READY` / `FATAL` ones. Per-provider errors are logged, not propagated. |
| **Provider deduplication** | `buildProviders` now appends `-1`, `-2` suffixes instead of silently dropping duplicates. |
| **`EventProvider`** | Adds `addEventObserver` / `removeEventObserver` for composite provider patterns. |

### Notes

- Rebased on current main. This picks up `initialize(ctx, domain)` from #1982 (forwarded to child providers) and `OpenFeatureAPI.createIsolated()` from #1928 (used by the new tests).
- Gap 5 in #1882 also asked that an explicit name conflict throw. Java's `MultiProvider` only accepts `List<FeatureProvider>` with no explicit-naming API, so there's nothing to conflict with; only the metadata-derived auto-dedup applies here.

### Related Issues

Fixes #1882
Replaces #1883

### DIFF
--- a/src/main/java/dev/openfeature/sdk/EventProvider.java
+++ b/src/main/java/dev/openfeature/sdk/EventProvider.java
@@ -3,10 +3,13 @@ package dev.openfeature.sdk;
 import dev.openfeature.sdk.internal.AutoCloseableReentrantReadWriteLock;
 import dev.openfeature.sdk.internal.ConfigurableThreadFactory;
 import dev.openfeature.sdk.internal.TriConsumer;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -24,6 +27,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public abstract class EventProvider implements FeatureProvider {
     private EventProviderListener eventProviderListener;
+    private final List<BiConsumer<ProviderEvent, ProviderEventDetails>> eventObservers = new CopyOnWriteArrayList<>();
     private final ExecutorService emitterExecutor =
             Executors.newCachedThreadPool(new ConfigurableThreadFactory("openfeature-event-emitter-thread", true));
 
@@ -71,6 +75,31 @@ public abstract class EventProvider implements FeatureProvider {
     }
 
     /**
+     * Add a provider event observer.
+     *
+     * <p>Observers are invoked whenever this provider emits an event and are intended for advanced
+     * provider composition scenarios.
+     *
+     * @param observer observer callback
+     */
+    public void addEventObserver(BiConsumer<ProviderEvent, ProviderEventDetails> observer) {
+        if (observer != null) {
+            eventObservers.add(observer);
+        }
+    }
+
+    /**
+     * Remove a previously registered provider event observer.
+     *
+     * @param observer observer callback
+     */
+    public void removeEventObserver(BiConsumer<ProviderEvent, ProviderEventDetails> observer) {
+        if (observer != null) {
+            eventObservers.remove(observer);
+        }
+    }
+
+    /**
      * Stop the event emitter executor and block until either termination has completed
      * or timeout period has elapsed.
      */
@@ -97,8 +126,9 @@ public abstract class EventProvider implements FeatureProvider {
     public Awaitable emit(final ProviderEvent event, final ProviderEventDetails details) {
         final var localEventProviderListener = this.eventProviderListener;
         final var localAttachment = this.attachment.get();
+        final var localEventObservers = this.eventObservers;
 
-        if (localEventProviderListener == null && localAttachment == null) {
+        if (localEventProviderListener == null && localAttachment == null && localEventObservers.isEmpty()) {
             return Awaitable.FINISHED;
         }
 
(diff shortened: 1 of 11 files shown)


jonathannorris:
Split this up into #2003, #2004, #2005

PR #1901, opened 2026-03-18
fix: collect and propagate per-provider errors in multi-provider strategies

### LINKED ISSUE #1900: [Multi-provider] FirstSuccessfulStrategy silently discards per-provider errors, returns generic message on all-fail
## Context

Related to #1882 (Gaps identified relative to js-sdk reference implementation).

When using `FirstSuccessfulStrategy` with `MultiProvider`, if all providers fail, the user receives only a generic error:

```
errorCode: GENERAL
errorMessage: "No provider successfully responded"
```

There is no information about *which* providers failed, *why* they failed, or what error codes each provider returned. Exceptions thrown by individual providers are caught and silently swallowed. Error-coded results from providers are discarded without any record.

This makes debugging multi-provider setups very difficult, especially in production where the root cause may differ across providers (e.g., one provider has a connection timeout, another has a parse error, a third returns flag-not-found).

`FirstMatchStrategy` has the same issue in its all-`FLAG_NOT_FOUND` fallback path, though it is less impactful since it only skips `FLAG_NOT_FOUND` errors.

### `FirstSuccessfulStrategy`

[`FirstSuccessfulStrategy.java:30-41`](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/multiprovider/FirstSuccessfulStrategy.java#L30-L41):

```java
    try {
        ProviderEvaluation<T> res = providerFunction.apply(provider);
        if (res.getErrorCode() == null) {
            return res;
        }
        // error-coded result is silently discarded here
    } catch (Exception ignored) {
        // thrown exception is silently swallowed here
    }
}

return ProviderEvaluation.<T>builder()
        .errorMessage("No provider successfully responded")
        .errorCode(ErrorCode.GENERAL)
        .build();
```

Both paths (error-coded results and thrown exceptions) are discarded with no tracking. The final fallback result carries no per-provider information.

### `FirstMatchStrategy`

[`FirstMatchStrategy.java:39-61`](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/multiprovider/FirstMatchStrategy.java#L39-L61) -- similarly discards individual `FLAG_NOT_FOUND` errors without collecting them:

```java
return ProviderEvaluation.<T>builder()
        .errorMessage("Flag not found in any provider")
        .errorCode(FLAG_NOT_FOUND)
        .build();
```

## Expected Behavior (js-sdk reference)

The js-sdk reference implementation handles this correctly:

1. **Every provider error is captured** -- both thrown exceptions (stored as `{ thrownError }`) and error-coded results are tracked in a `resolutions[]` array during iteration.

2. **Errors are aggregated on all-fail** -- When all providers fail, [`collectProviderErrors()`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/provider/multi-provider/strategies/base-evaluation-strategy.ts#L130-L145) iterates all resolutions and builds an array of `{ providerName, error }` for each failed provider.

3. **User receives an `AggregateError`** -- The [`MultiProvider` throws an `AggregateError`](https://github.com/open-feature/js-sdk/blob/main/packages/server/src/provider/multi-provider/multi-provider.ts#L194-L195) that includes:
   - `.message` -- contains the first error's provider name and message for convenience
   - `.originalErrors[]` -- array of `{ source: providerName, error }` for **every** failed provider

Example of what a js-sdk user sees when 3 providers fail:

```
AggregateError: Provider errors occurred: ProviderA: Connection timeout
  originalErrors: [
    { source: "ProviderA", error: ConnectionTimeoutError },
    { source: "ProviderB", error: ParseError },
    { source: "ProviderC", error: FlagNotFoundError }
  ]
```

Versus what a java-sdk user sees for the same scenario:

```
errorCode: GENERAL
errorMessage: "No provider successfully responded"
```

## Proposed Solution

1. **Collect per-provider errors during iteration** -- Both `FirstSuccessfulStrategy` and `FirstMatchStrategy` should accumulate error details (provider name, error code, error message, original exception) as they iterate through providers.

2. **Expose per-provider errors in the result** -- Add a `List<ProviderError>` field to `ProviderEvaluation` (and propagate through to `FlagEvaluationDetails`) so that on an all-fail result, the user can programmatically inspect each provider's error. A `ProviderError` type would carry:
   - `providerName` (String)
   - `errorCode` (ErrorCode)
   - `errorMessage` (String)
   - `exception` (Exception, nullable)

3. **Build a descriptive aggregate error message** -- The `errorMessage` on the final `ProviderEvaluation` should summarize the per-provider errors, e.g.:
   ```
   No provider successfully responded. Provider errors: [provider1: PARSE_ERROR (parse failed), provider2: GENERAL (connection timeout)]
   ```

4. **Ensure the `Strategy` interface contract is backwards-compatible** -- The `Strategy` interface itself doesn't change. The `ProviderEvaluation` field would be additive (defaults to an empty list).

## References

- js-sdk `FirstSuccessfulStrategy`: [`first-successful-strategy.ts`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/provider/multi-provider/strategies/first-successful-strategy.ts)
- js-sdk `BaseEvaluationStrategy.collectProviderErrors()`: [`base-evaluation-strategy.ts#L130-L145`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/provider/multi-provider/strategies/base-evaluation-strategy.ts#L130-L145)
- js-sdk `AggregateError` / `constructAggregateError`: [`errors.ts`](https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/provider/multi-provider/errors.ts)
- js-sdk `MultiProvider` error propagation: [`multi-provider.ts#L194-L202`](https://github.com/open-feature/js-sdk/blob/main/packages/server/src/provider/multi-provider/multi-provider.ts#L194-L202)
- Related tracking issue: #1882
- Multi-provider spec: https://openfeature.dev/specification/appendix-a/#multi-provider

suthar26:
## This PR

FirstSuccessfulStrategy and FirstMatchStrategy now collect per-provider error details (provider name, error code, message, exception) as they iterate through providers. When all providers fail, the returned ProviderEvaluation includes:

- A descriptive errorMessage listing each provider's failure
- A List<ProviderError> with structured per-provider error details

This aligns the Java SDK with the js-sdk reference implementation, which uses AggregateError with originalErrors[] to expose per-provider failure information to callers.

Changes:
- Add ProviderError class for per-provider error info
- Add providerErrors field to ProviderEvaluation and FlagEvaluationDetails
- Update FirstSuccessfulStrategy to collect errors from all failed providers
- Update FirstMatchStrategy to collect FLAG_NOT_FOUND errors
- Update Strategy interface javadoc
- Add comprehensive tests for error propagation

### Related Issues
Closes open-feature/java-sdk#1900

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ProviderEvaluation.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderEvaluation.java
@@ -4,6 +4,7 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
 
 /**
  * Contains information about how the a flag was evaluated, including the resolved value.
@@ -11,7 +12,7 @@ import lombok.NoArgsConstructor;
  * @param <T> the type of the flag being evaluated.
  */
 @Data
-@Builder
+@SuperBuilder
 @NoArgsConstructor
 @AllArgsConstructor
 public class ProviderEvaluation<T> implements BaseEvaluation<T> {
--- a/src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
@@ -7,12 +7,13 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * A {@link ProviderEvaluation} subtype returned by multi-provider strategies when all providers
- * fail to produce a successful result.
+ * A {@link ProviderEvaluation} subtype returned by multi-provider strategies that carries
+ * per-provider error details.
  *
- * <p>This type carries per-provider error details so that callers can inspect why each individual
- * provider failed. It is only returned in error scenarios; successful evaluations always return
- * a plain {@link ProviderEvaluation}.
+ * <p>This type can represent both successful and failed evaluations. When a strategy exhausts
+ * all providers without a successful result, the per-provider errors describe why each provider
+ * failed. Custom strategies may also use this type for successful results to surface information
+ * about providers that were skipped or failed before the successful one.
  *
  * <p>Usage:
  * <pre>{@code
--- a/src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
@@ -1,0 +1,122 @@
+package dev.openfeature.sdk.multiprovider;
+
+import dev.openfeature.sdk.ErrorCode;
+import dev.openfeature.sdk.ImmutableMetadata;
+import dev.openfeature.sdk.ProviderEvaluation;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A {@link ProviderEvaluation} subtype returned by multi-provider strategies when all providers
+ * fail to produce a successful result.
+ *
+ * <p>This type carries per-provider error details so that callers can inspect why each individual
+ * provider failed. It is only returned in error scenarios; successful evaluations always return
+ * a plain {@link ProviderEvaluation}.
+ *
+ * <p>Usage:
+ * <pre>{@code
+ * ProviderEvaluation<String> result = strategy.evaluate(...);
+ * if (result instanceof MultiProviderEvaluation<String> multiResult) {
+ *     for (ProviderError error : multiResult.getProviderErrors()) {
+ *         log.warn("Provider {} failed: {} - {}",
+ *             error.getProviderName(), error.getErrorCode(), error.getErrorMessage());
+ *     }
+ * }
+ * }</pre>
+ *
+ * @param <T> the type of the flag being evaluated
+ */
+public class MultiProviderEvaluation<T> extends ProviderEvaluation<T> {
+
+    private final List<ProviderError> providerErrors;
+
+    private MultiProviderEvaluation(
+            T value,
+            String variant,
+            String reason,
+            ErrorCode errorCode,
+            String errorMessage,
+            ImmutableMetadata flagMetadata,
+            List<ProviderError> providerErrors) {
+        super(value, variant, reason, errorCode, errorMessage, flagMetadata);
+        this.providerErrors =
+                providerErrors != null ? Collections.unmodifiableList(providerErrors) : Collections.emptyList();
+    }
+
+    /**
+     * Returns the per-provider error details.
+     *
+     * <p>Each entry describes why a specific provider failed during multi-provider evaluation.
+     *
+     * @return an unmodifiable list of per-provider errors, never {@code null}
+     */
+    public List<ProviderError> getProviderErrors() {
+        return providerErrors;
+    }
+
+    /**
+     * Create a new builder for {@link MultiProviderEvaluation}.
+     *
+     * @param <T> the flag value type
+     * @return a new builder
+     */
+    public static <T> Builder<T> multiProviderBuilder() {
+        return new Builder<>();
+    }
+
+    /**
+     * Builder for {@link MultiProviderEvaluation}.
+     *
+     * @param <T> the flag value type
+     */
+    public static class Builder<T> {
+        private T value;
+        private String variant;
+        private String reason;
+        private ErrorCode errorCode;
+        private String errorMessage;
+        private ImmutableMetadata flagMetadata;
+        private List<ProviderError> providerErrors;
+
+        public Builder<T> value(T value) {
+            this.value = value;
+            return this;
+        }
+
+        public Builder<T> variant(String variant) {
+            this.variant = variant;
+            return this;
+        }
+
+        public Builder<T> reason(String reason) {
+            this.reason = reason;
+            return this;
+        }
+
+        public Builder<T> errorCode(ErrorCode errorCode) {
+            this.errorCode = errorCode;
+            return this;
+        }
+
+        public Builder<T> errorMessage(String errorMessage) {
+            this.errorMessage = errorMessage;
+            return this;
+        }
+
+        public Builder<T> flagMetadata(ImmutableMetadata flagMetadata) {
+            this.flagMetadata = flagMetadata;
+            return this;
+        }
+
+        public Builder<T> providerErrors(List<ProviderError> providerErrors) {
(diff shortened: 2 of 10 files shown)


src/main/java/dev/openfeature/sdk/multiprovider/ProviderError.java
```diff
@@ -0,0 +1,70 @@
...
+ * Represents an error from a single provider during multi-provider evaluation.
+ *
+ * <p>Captures the provider name, error code, error message, and optionally the original exception
+ * that occurred during flag evaluation. This allows callers to inspect per-provider error details
+ * when a multi-provider strategy exhausts all providers without a successful result.
+ */
+@Data
+@Builder
+@AllArgsConstructor
+public class ProviderError {
```
chrfwow, line 21: The error code will always be `not found`, right? So should we maybe convert this to a `NotFound` class, or maybe even just save the names of the providers that did not find the flag?

chrfwow (commented):
Is the plan to eventually also track different error codes than `Flag not found`? Because otherwise we could simplify this a bit

src/main/java/dev/openfeature/sdk/multiprovider/ProviderError.java
suthar26: ProviderError is used across the strategies and we need the full error codes in the other strategies, 
if you believe it is a better approach to simplify FirstMatchStrategy to just get the name of the providers I can implement that.

chrfwow: Ok, when we use it in the other strategies it's fine to be consistent

chrfwow (approved):

toddbaert (commented):
Nit: the hand-rolled builder here re-declares all of `ProviderEvaluation`'s fields, so if the parent ever gains a new field this will silently fall out of sync. Lombok's `@SuperBuilder` handles inheritance nicely; you'd just need to add `@SuperBuilder` to both `ProviderEvaluation` and `MultiProviderEvaluation` and it takes care of the rest. Alternatively, if changing `ProviderEvaluation` is off the table, a comment flagging the coupling would be nice.

toddbaert (commented):
Small thing: `ProviderEvaluation`'s Lombok builder defaults `flagMetadata` to a non-null `ImmutableMetadata` via `@Builder.Default`, but the `MultiProviderEvaluation` builder doesn't set that default. Neither strategy sets it, so `getFlagMetadata()` will return `null` on these results; could cause NPEs downstream. Initializing it in the builder field (e.g. `private ImmutableMetadata flagMetadata = ImmutableMetadata.builder().build()`) should do the trick. This also goes away if you switch to `@SuperBuilder` since it'd inherit the default.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
```diff
@@ -0,0 +1,123 @@
...
+        private String reason;
+        private ErrorCode errorCode;
+        private String errorMessage;
+        private ImmutableMetadata flagMetadata;
+        private List<ProviderError> providerErrors;
+
+        public Builder<T> value(T value) {
+            this.value = value;
+            return this;
+        }
```
toddbaert: Small thing: `ProviderEvaluation`'s Lombok builder defaults `flagMetadata` to a non-null `ImmutableMetadata` via `@Builder.Default`, but this builder doesn't set that default. Neither strategy sets it, so `getFlagMetadata()` will return `null` on these results; could cause NPEs downstream. Initializing it here (e.g. `private ImmutableMetadata flagMetadata = ImmutableMetadata.builder().build()`) should do the trick. This also goes away if you switch to `@SuperBuilder` since it'd inherit the default.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
```diff
@@ -0,0 +1,123 @@
...
+     * Builder for {@link MultiProviderEvaluation}.
+     *
+     * @param <T> the flag value type
+     */
+    public static class Builder<T> {
+        private T value;
+        private String variant;
+        private String reason;
+        private ErrorCode errorCode;
+        private String errorMessage;
```
toddbaert: Nit: this builder re-declares all of `ProviderEvaluation`'s fields, so if the parent ever gains a new field this will silently fall out of sync. Lombok's `@SuperBuilder` handles inheritance nicely; you'd just need to add `@SuperBuilder` to both `ProviderEvaluation` and `MultiProviderEvaluation` and it takes care of the rest. Alternatively, if changing `ProviderEvaluation` is off the table, a comment flagging the coupling would be nice.

toddbaert (commented):
I agree with the spirit of this PR, but I think that there's a few places the use of lombok can prevent some bugs and inconsistencies. I'm not a huge fan of lombok TBH, but we are already "in on it" so I think we should stay consistent.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
suthar26: addressed this

src/main/java/dev/openfeature/sdk/multiprovider/MultiProviderEvaluation.java
suthar26: addressed this using `@SuperBuilder`

toddbaert (approved):

src/main/java/dev/openfeature/sdk/ProviderEvaluation.java
```diff
@@ -4,14 +4,15 @@
...
+import lombok.experimental.SuperBuilder;
 
 /**
  * Contains information about how the a flag was evaluated, including the resolved value.
  *
  * @param <T> the type of the flag being evaluated.
  */
 @Data
-@Builder
+@SuperBuilder
```
kinyoklion, line 15: @toddbaert It is minor, but this created a compilation failure in the LD provider. It adds 2 more type parameters for ProviderEvaluation, so if you explicitly type it anywhere, then the build fails.

Example: https://github.com/launchdarkly/openfeature-java-server/actions/runs/28385445702/job/84098789250#step:3:275

PR #1928, opened 2026-04-16
feat: support isolated API instances

### LINKED ISSUE #1925: Support isolated API instances
The spec now supports creating independent, non-singleton instances of the OpenFeature API. Each instance has its own providers, context, hooks, and events; totally separate from the global singleton and from each other. This is useful for things like micro-frontends, DI frameworks, and testing.

### What needs to happen

- Add a factory function (or equivalent) that creates new, independent API instances
- Each instance should conform to the same API contract as the existing singleton
- Ideally the factory lives in its own module/package for discoverability (recommended, not required)
- A single provider instance shouldn't be bound to more than one API instance at a time

### Links

- Parent issue: https://github.com/open-feature/spec/issues/374
- Spec section: https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances

marcozabel:
## This PR
This PR introduces support for creating isolated OpenFeature API instances, each with their own providers, hooks, context, and event handling - enabling multi-tenant or side-by-side usage without shared global state.

### Related Issues

Fixes https://github.com/open-feature/java-sdk/issues/1925


src/main/java/dev/openfeature/sdk/isolated/OpenFeatureAPIFactory.java
```diff
@@ -0,0 +1,48 @@
...
+     * <pre>{@code
+     * OpenFeatureAPI api = OpenFeatureAPIFactory.createAPI();
+     * api.setProvider(new MyProvider());
+     * Client client = api.getClient();
+     * }</pre>
+     *
+     * @return a new API instance
+     * @see OpenFeatureAPI#createIsolated()
+     */
+    public static OpenFeatureAPI createAPI() {
```
chrfwow, line 45: I don't fully see the point of this method/class. `OpenFeatureAPI.createIsolated()` has the exact same visibility level and does the same thing. I think this will only lead to confusion

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -91,7 +99,7 @@ public Awaitable emit(final ProviderEvent event, final ProviderEventDetails deta
         // These calls need to be executed on a different thread to prevent deadlocks when the provider initialization
         // relies on a ready event to be emitted
         emitterExecutor.submit(() -> {
-            try (var ignored = OpenFeatureAPI.lock.readLockAutoCloseable()) {
+            try (var ignored = localLock != null ? localLock.readLockAutoCloseable() : null) {
```
chrfwow, line 102: Why is the lock optional? IIRC, we needed that lock to avoid race conditions when emitting events

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -22,8 +22,8 @@
 @Slf4j
 @SuppressWarnings("PMD.UnusedLocalVariable")
 public class OpenFeatureAPI implements EventBus<OpenFeatureAPI> {
-    // package-private multi-read/single-write lock
-    static AutoCloseableReentrantReadWriteLock lock = new AutoCloseableReentrantReadWriteLock();
+    // package-private multi-read/single-write lock (instance-level for isolation)
+    AutoCloseableReentrantReadWriteLock lock = new AutoCloseableReentrantReadWriteLock();
```
chrfwow, line 26: Can this lock be final?

src/test/java/dev/openfeature/sdk/LockingSingeltonTest.java
```diff
@@ -33,7 +33,7 @@ void beforeEach() {
         client = (OpenFeatureClient) api.getClient("LockingTest");
 
         apiLock = setupLock(apiLock, mockInnerReadLock(), mockInnerWriteLock());
-        OpenFeatureAPI.lock = apiLock;
+        api.lock = apiLock;
```
chrfwow, line 36: Instead of setting the lock like this, can we also add a package private constructor to the api that takes a lock?

src/test/java/dev/openfeature/sdk/isolated/IsolatedAPITest.java
```diff
@@ -0,0 +1,204 @@
...
+
+        // Handlers are dispatched asynchronously; use a latch to await api1's handler.
+        api1.onProviderReady(details -> api1HandlerLatch.countDown());
+        api2.onProviderReady(details -> api2HandlerCalled.set(true));
+
+        // setting a provider on api1 should only trigger api1's handler
+        api1.setProviderAndWait(new NoOpProvider());
+
+        assertThat(api1HandlerLatch.await(1, TimeUnit.SECONDS)).isTrue();
+        assertThat(api2HandlerCalled.get()).isFalse();
```
chrfwow, line 104: I am not sure this test really tests anything, the lambda for api2 might still get called some time after this tests finished. Unless we can read the queue of the `emitterExecutor` in the `EventProvider`, we cannot be sure that the lambda isn't scheduled for execution.

We could wait before this line for a few hundred ms. While this would slow down the test suite and still would not be a real proof, it should catch most errors

src/main/java/dev/openfeature/sdk/isolated/OpenFeatureAPIFactory.java
```diff
@@ -0,0 +1,52 @@
...
+ * <p><strong>Spec references:</strong>
+ * <ul>
+ *   <li>Requirement 1.8.1 &mdash; factory function for isolated instances</li>
+ *   <li>Requirement 1.8.3 &mdash; factory in a distinct package/module</li>
+ * </ul>
+ *
+ * @see <a href="https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances">
+ *     Spec &sect;1.8 &mdash; Isolated API Instances</a>
+ */
+public final class OpenFeatureAPIFactory {
```
toddbaert: Section 1.8 is marked experimental in the spec. Might be worth adding an `@apiNote` to flag that, e.g.:

```java
* @apiNote This API is experimental and subject to change.
```

Could go on the class-level Javadoc or on `createAPI()` itself (or both).

src/main/java/dev/openfeature/sdk/isolated/OpenFeatureAPIFactory.java
toddbaert: Agree

src/test/java/dev/openfeature/sdk/isolated/IsolatedAPISingeltonTest.java
```diff
@@ -0,0 +1,50 @@
...
+        singleton.setTransactionContextPropagator(new NoOpTransactionContextPropagator());
+    }
+
+    /**
+     * Requirement 1.8.1 — isolated instances do not share state with
+     * the global singleton.
+     */
+    @Test
+    @DisplayName("isolated instance does not interfere with singleton")
+    void isolatedInstanceDoesNotInterfereWithSingleton() {
```
toddbaert, line 34: This tests that mutating an isolated instance doesn't affect the singleton, which is great. Worth adding the reverse direction too; e.g. set a provider/hook/context on the singleton and assert a previously created isolated instance is unaffected.

leakonvalinka (approved):

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -29,29 +30,46 @@ void setEventProviderListener(EventProviderListener eventProviderListener) {
...
+            TriConsumer<EventProvider, ProviderEvent, ProviderEventDetails> onEmit,
+            AutoCloseableReentrantReadWriteLock lock) {
+        Attachment existing = this.attachment;
+        if (existing != null && existing.onEmit != onEmit) {
             // if we are trying to attach this provider to a different onEmit, something has gone wrong
             throw new IllegalStateException("Provider " + this.getMetadata().getName() + " is already attached.");
-        } else {
-            this.onEmit = onEmit;
         }
+        this.attachment = new Attachment(onEmit, lock);
```
chrfwow: `this.attachment` is not guaranteed to be `null` here any more. If we want to ensure that, we need an `AtomicReference` and a cas operation

src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -9,28 +9,63 @@
...
+    /**
+     * Creates a new, independent {@link OpenFeatureAPI} instance with fully isolated state
+     * (providers, hooks, evaluation context, event handlers, transaction context propagators).
+     *
+     * <p>For typical usage, prefer the global singleton via {@link #getInstance()}.
+     *
+     * <p><strong>Note:</strong> Isolated API instances (per spec section 1.8) are experimental and
+     * subject to change.
+     */
+    public OpenFeatureAPI() {
```
toddbaert: I see the `OpenFeatureAPIFactory` was removed in favor of a public constructor; can you share the reasoning? The factory-in-a-distinct-package approach felt closer to spec 1.8.3 (factory function in a distinct module/package/namespace for discoverability), and intentionally less discoverable than the singleton.

IIUC, a public constructor on `OpenFeatureAPI` itself sits right next to `getInstance()` in IntelliSense, which works against the "intentionally less discoverable" goal. Was there a specific issue with the factory class that prompted the removal? Happy to be convinced otherwise.

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: +1 to this; we flagged the same race in our earlier review pass. `volatile` gives visibility but not atomicity. An `AtomicReference` with `compareAndSet` would be better I think, or just guarding `attach()` with a synchronized block since it's not on a hot path.

src/main/java/dev/openfeature/sdk/EventProvider.java
```diff
@@ -57,19 +58,25 @@ private static final class Attachment {
...
+        while (true) {
+            Attachment existing = this.attachment.get();
+            if (existing != null && existing.onEmit != onEmit) {
+                // if we are trying to attach this provider to a different onEmit, something has gone wrong
+                throw new IllegalStateException(
+                        "Provider " + this.getMetadata().getName() + " is already attached.");
+            }
+            if (this.attachment.compareAndSet(existing, newAttachment)) {
+                return;
+            }
```
chrfwow: Couldn't we simplify all of this with 
```
if (!this.attachment.compareAndSet(null, newAttachment)) {
   throw new IllegalStateException();
}
```
we expect the existing attachement to be `null`, every other case is exceptional, and we wouldn't retry anyway

toddbaert:
Hey @marcozabel; pushed two minor changes:

- configured maven-javadoc-plugin to recognize `@apiNote` (fixes the failing Java 17 CI; build-time only, no runtime changes).
- removed `OpenFeatureAPIFactory` per @chrfwow's point on https://github.com/open-feature/java-sdk/pull/1928#discussion_r3098063982; the sub-package factory doesn't actually achieve the "less discoverable" goal because Java's package-access rules force us to expose public surface either way. Spec 1.8.3 is a SHOULD, so we're still compliant.

New public API surface is now just:

```java
public class OpenFeatureAPI {
    public static OpenFeatureAPI createIsolated();   // factory; spec 1.8.1
}
```

src/main/java/dev/openfeature/sdk/EventProvider.java
toddbaert: I made this change... Also dropped `doesNotThrowWhenOnEmitSame` which failed with this change, BUT that test guarded a code branch that's  not reachable in practice (`attach()` is internal, and the internal caller passes a fresh method-reference each time).

toddbaert (approved):

toddbaert:
I'll merge this Friday unless I hear objections.

chrfwow (approved):

toddbaert:
@chrfwow I see you approved, so just a quick FYI - I updated this to use a weakmap to avoid some memory leakage as well as changing the warning to a throw like other impls: https://github.com/open-feature/java-sdk/pull/1928/commits/80e2a96dc068d92c998a17b836cd415efbb11c01

chrfwow (approved):

PR #1936, opened 2026-05-07
fix: shutdown now clears hooks, evaluationContext and transactionContextPropagator as per spec 1.6.2

### LINKED ISSUE #1926: shutdown() doesn't clear hooks or evaluation context
The spec (requirement 1.6.2) says shutdown MUST reset all API state, removing hooks, event handlers, and providers. There's also a pending clarification to explicitly include evaluation context and transaction context propagators (open-feature/spec#375).

Right now, `OpenFeatureAPI.shutdown()` resets providers and event handlers, but leaves `apiHooks`, `evaluationContext`, and the transaction context propagator intact. Those should be cleared too.

Found this while doing a cross-SDK audit for Python 1.0 readiness.

jarebudev:
## This PR

Ensures that Spec 1.6.2 is implemented correctly. shutdown will now also clear the hooks, evaluationContext and transactionContextPropagator.

### Related Issues

Fixes #1926

### Notes
N/A

### Follow-up Tasks
N/A

### How to test
 Unit test added to cover this change

### DIFF
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
@@ -356,7 +356,7 @@ public class OpenFeatureAPI implements EventBus<OpenFeatureAPI> {
                 providerRepository = new ProviderRepository(this);
                 eventSupport = new EventSupport();
                 clearHooks();
-                setEvaluationContext(new ImmutableContext());
+                setEvaluationContext(ImmutableContext.EMPTY);
                 setTransactionContextPropagator(new NoOpTransactionContextPropagator());
             }
         }
--- a/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
+++ b/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
@@ -355,6 +355,9 @@ public class OpenFeatureAPI implements EventBus<OpenFeatureAPI> {
             try (AutoCloseableLock ignored = lock.writeLockAutoCloseable()) {
                 providerRepository = new ProviderRepository(this);
                 eventSupport = new EventSupport();
+                clearHooks();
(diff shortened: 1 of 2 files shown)


src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java
```diff
@@ -355,6 +355,9 @@ public void shutdown() {
             try (AutoCloseableLock ignored = lock.writeLockAutoCloseable()) {
                 providerRepository = new ProviderRepository(this);
                 eventSupport = new EventSupport();
+                clearHooks();
+                setEvaluationContext(new ImmutableContext());
```
chrfwow: We could use `ImmutableContext.EMPTY` here

src/test/java/dev/openfeature/sdk/ShutdownBehaviorSpecTest.java
```diff
@@ -131,16 +133,28 @@ void mustShutdownAllProvidersOnShuttingDownApi() {
...
+            api.setTransactionContextPropagator(transactionContextPropagator);
 
             api.shutdown();
 
-            NoOpProvider p2 = new NoOpProvider();
-            api.setProvider(p2);
+            assertNotEquals(provider, api.getProvider());
+            assertTrue(api.getHooks().isEmpty());
+            assertNotEquals(evaluationContext, api.getEvaluationContext());
+            assertNotEquals(transactionContextPropagator, api.getTransactionContextPropagator());
```
chrfwow, line 158: Maybe we should also assert that we have a `NoOpTransactionContextPropagator`, an empty context, and the correct provider

chrfwow (approved):
lgtm. I don't understand where the license issue comes from

beeme1mr (approved):
FYI, I've resolved the false positive licensing issue.

PR #1972, opened 2026-06-19
perf: avoid ImmutableStructure allocation for empty ImmutableContext

tobias-ibounig-dt:
## This PR

- Add `ImmutableStructure.EMPTY` singleton (package-private)
- `ImmutableContext()` and `ImmutableContext(String)` pass `Collections.emptyMap()` instead of `new HashMap<>()`
- `ImmutableContext(String targetingKey, Map attributes)` reuses `ImmutableStructure.EMPTY` when `targetingKey` is null and attributes are empty
- Tests to guard the constructor branch logic

### Related Issues
None

### Notes

Before-hook implementations commonly return `new ImmutableContext()` to signal no context change. Each call previously allocated a throwaway `HashMap` as the constructor argument, then a second `HashMap` inside `ImmutableStructure`. Passing `Collections.emptyMap()` eliminates the first allocation; reusing `ImmutableStructure.EMPTY` eliminates both the `ImmutableStructure` and its backing map entirely.

The sharing should be fine: `ImmutableStructure` never mutates `attributes` — `getValue` clones values, `asMap`/`keySet` return copies.

The `ConstructorBranches` tests guard the `targetingKey != null` branch so that a future refactor cannot accidentally return `EMPTY` for a context that carries a targeting key.

| Metric | `main` | This PR | Delta |
|--------|--------|---------|-------|
| `run:+totalAllocatedInstances` | 2,145,282 | 1,825,005 | −320,277 (−14.9%) |
| `run:+totalAllocatedBytes` | 102,105,728 | 93,695,320 | −8,410,408 (−8.2%) |

:warning: seems I might have used another branch as baseline instead of main. Result proportions should be right, but not the exact values.

### Follow-up Tasks
- More PRs with memory improvements
- Update `benchmark.txt` after all are applied

### DIFF
--- a/src/main/java/dev/openfeature/sdk/ImmutableContext.java
+++ b/src/main/java/dev/openfeature/sdk/ImmutableContext.java
@@ -2,7 +2,6 @@ package dev.openfeature.sdk;
 
 import dev.openfeature.sdk.internal.ExcludeFromGeneratedCoverageReport;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
 import lombok.ToString;
@@ -32,7 +31,7 @@ public final class ImmutableContext implements EvaluationContext {
      * provided.
      */
     public ImmutableContext() {
-        this(new HashMap<>());
+        this(Collections.emptyMap());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


toddbaert:
> ⚠️ seems I might have used another branch as baseline instead of main. Result proportions should be right, but not the exact values.

haha ya, I first looked at the raw numbers and I was shocked - but still, these improvements are valid! Thanks as always.

toddbaert (approved):

PR #921, opened 2024-05-03
fix: removed javax.nullable annotations

DBlanchard88:
removed all @javax.nullable annotations

Fixes #867

PR #1982, opened 2026-07-02
feat: supply bound domain to provider initialization

### LINKED ISSUE #1981: Supply the bound domain to provider initialization
The spec now supplies the bound `domain` to a provider's `initialize` function, and adds an opt-in `domain-scoped` provider declaration that the API enforces by binding such a provider to at most one `domain`. This lets stateful providers (e.g. an OFREP static-context provider with a persistent cache) scope their state to a single, unambiguous `domain`.

### What needs to happen

- Pass the bound `domain` (if any) to the provider's `initialize` when it's registered via the provider mutator
- Extend the provider `initialize` to accept the optional `domain` (non-breaking, e.g. an overload or default method)
- Add a way for a provider to declare itself `domain-scoped`
- Enforce that a `domain-scoped` provider is bound to at most one `domain`, rejecting any attempt to bind an already-bound instance to another `domain`
- Associated testing and docs

### Links

- Parent issue: https://github.com/open-feature/spec/issues/403
- Spec sections: [1.1 Flag evaluation](https://openfeature.dev/specification/sections/flag-evaluation), [2.4 Initialization](https://openfeature.dev/specification/sections/providers)

jonathannorris:
## Summary

- Add optional `domain` parameter to provider `initialize` and pass the bound domain from the provider mutator (spec 1.1.2.2, 2.4.1)
- Add `isDomainScoped()` provider declaration and reject binding a domain-scoped instance to more than one domain (spec 2.4.3, 1.1.8.1)
- Forward `domain` through `MultiProvider` `initialize` to each child provider
- Add spec conformance tests, including backward compatibility for providers that only override single-arg `initialize`

## Motivation

Unblocks OFREP static-context providers that need the bound `domain` at init time to scope persisted cache keys per [open-feature/spec#393](https://github.com/open-feature/spec/pull/393) and [protocol ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md).

Parallel implementation in js-sdk: [open-feature/js-sdk#1433](https://github.com/open-feature/js-sdk/pull/1433).

### Notes

- Non-breaking: `domain` is an optional second parameter via a default method on `FeatureProvider`
- Non-domain-scoped providers retain existing behavior (single instance can back multiple domains; `initialize` runs once)
- Legacy providers that only implement `initialize(context)` remain compatible; the extra `domain` argument is ignored if unused
- Default provider is initialized with a `null` domain; named clients receive the bound domain string

### Related Issues

Fixes #1981

Relates to: [open-feature/spec#393](https://github.com/open-feature/spec/pull/393), [open-feature/js-sdk#1433](https://github.com/open-feature/js-sdk/pull/1433)

## Test plan

- [x] `InitializeBehaviorSpecTest` verifies domain is passed at init (2.4.1)
- [x] `DomainScopedProviderSpecTest` covers 1.1.8.1 and 2.4.3
- [x] `mvn test` passes after rebase onto `main`

### DIFF
--- a/src/main/java/dev/openfeature/sdk/FeatureProvider.java
+++ b/src/main/java/dev/openfeature/sdk/FeatureProvider.java
@@ -1,5 +1,6 @@
 package dev.openfeature.sdk;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -183,6 +184,38 @@ public interface FeatureProvider {
      */
     default void initialize(EvaluationContext evaluationContext) throws Exception {
         // Intentionally left blank
+    }
+
+    /**
+     * This method is called before a provider is used to evaluate flags, with the
+     * bound domain supplied when the provider is registered to a named client.
+     *
+     * <p>
+     * The default provider is initialized with a {@code null} domain. Providers that
+     * maintain per-domain state (for example a persistent cache) should override this
+     * method and declare themselves {@linkplain #isDomainScoped() domain-scoped}.
+     * </p>
+     *
+     * @param evaluationContext the global evaluation context
+     * @param domain            the bound domain, or {@code null} for the default provider
+     */
+    default void initialize(EvaluationContext evaluationContext, @Nullable String domain) throws Exception {
+        initialize(evaluationContext);
+    }
+
+    /**
+     * Returns whether this provider maintains state specific to a single domain that
+     * cannot be shared across domains.
+     *
+     * <p>
+     * Domain-scoped providers may only be bound to one domain within a single API
+     * instance.
+     * </p>
+     *
+     * @return {@code true} if this provider is domain-scoped
+     */
+    default boolean isDomainScoped() {
+        return false;
     }
 
     /**
--- a/src/main/java/dev/openfeature/sdk/FeatureProvider.java
+++ b/src/main/java/dev/openfeature/sdk/FeatureProvider.java
@@ -2,7 +2,6 @@ package dev.openfeature.sdk;
 
 import java.util.ArrayList;
 import java.util.List;
-import javax.annotation.Nullable;
 
 /**
  * The interface implemented by upstream flag providers to resolve flags for
@@ -199,7 +198,7 @@ public interface FeatureProvider {
      * @param evaluationContext the global evaluation context
      * @param domain            the bound domain, or {@code null} for the default provider
      */
-    default void initialize(EvaluationContext evaluationContext, @Nullable String domain) throws Exception {
+    default void initialize(EvaluationContext evaluationContext, String domain) throws Exception {
         initialize(evaluationContext);
     }
 
--- a/src/main/java/dev/openfeature/sdk/FeatureProvider.java
+++ b/src/main/java/dev/openfeature/sdk/FeatureProvider.java
@@ -1,8 +1,8 @@
 package dev.openfeature.sdk;
 
-import edu.umd.cs.findbugs.annotations.Nullable;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /**
  * The interface implemented by upstream flag providers to resolve flags for
--- a/src/main/java/dev/openfeature/sdk/ProviderRepository.java
+++ b/src/main/java/dev/openfeature/sdk/ProviderRepository.java
@@ -17,7 +17,6 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
@@ -155,7 +154,7 @@ class ProviderRepository {
     }
 
     private void prepareAndInitializeProvider(
-            @Nullable String domain,
+            String domain,
             FeatureProvider newProvider,
             Consumer<FeatureProvider> afterSet,
             Consumer<FeatureProvider> afterInit,
@@ -196,28 +195,22 @@ class ProviderRepository {
         }
     }
 
-    private void validateDomainScopedBinding(@Nullable String domain, FeatureProvider newProvider) {
+    private void validateDomainScopedBinding(String domain, FeatureProvider newProvider) {
         if (!newProvider.isDomainScoped()) {
             return;
         }
 
-        boolean currentlyDefault = isDefaultProviderInstance(newProvider);
-        List<String> boundNamedDomains = getBoundDomainsForProviderInstance(newProvider);
-
-        if (!currentlyDefault && boundNamedDomains.isEmpty()) {
+        // a re-set to the identical binding is always allowed (it's a no-op)
+        boolean alreadyBoundHere = domain == null
+                ? isDefaultProviderInstance(newProvider)
+                : getBoundDomainsForProviderInstance(newProvider).contains(domain);
+        if (alreadyBoundHere) {
             return;
         }
 
-        if (domain == null) {
-            if (!currentlyDefault) {
-                throw new IllegalArgumentException("Domain-scoped provider cannot be bound to more than one domain");
-            }
-            return;
-        }
-        if (boundNamedDomains.contains(domain)) {
-            return;
-        }
-        if (!boundNamedDomains.isEmpty() || currentlyDefault) {
+        // any other existing binding means this instance would span more than one domain
+        if (isDefaultProviderInstance(newProvider)
+                || !getBoundDomainsForProviderInstance(newProvider).isEmpty()) {
             throw new IllegalArgumentException("Domain-scoped provider cannot be bound to more than one domain");
         }
     }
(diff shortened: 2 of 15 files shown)


src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -185,15 +187,39 @@ private void prepareAndInitializeProvider(
...
+    private void validateDomainScopedBinding(
+            @Nullable String domain, FeatureProvider newProvider, @Nullable FeatureProviderStateManager existing) {
+        if (!newProvider.isDomainScoped()) {
+            return;
+        }
+        if (existing == null) {
+            return;
+        }
+        boolean currentlyDefault = isDefaultProvider(newProvider);
+        List<String> boundNamedDomains = getDomainsForProvider(newProvider);
```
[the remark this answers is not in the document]
jonathannorris: Agreed — spec 1.1.8.1 is about the provider instance, which in Java means reference identity. Updated `validateDomainScopedBinding` to check registration with `==` on the bound delegate. `getExistingStateManagerForProvider` now also uses identity matching for domain-scoped providers so equal-but-distinct instances are not merged. Non-domain-scoped sharing still uses `equals` via `hasSameProvider`.

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -98,7 +112,7 @@ public void initialize(EvaluationContext evaluationContext) throws Exception {
             Collection<Callable<Void>> tasks = new ArrayList<>(providers.size());
             for (FeatureProvider provider : providers.values()) {
                 tasks.add(() -> {
-                    provider.initialize(evaluationContext);
+                    provider.initialize(evaluationContext, domain);
                     return null;
```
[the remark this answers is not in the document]
jonathannorris: The PR description was stale — the implementation and tests forward `(context, domain)` to each child. Updated the summary bullet to match. MultiProvider does not enforce `isDomainScoped()` itself; that guard runs on direct `setProvider` binding in `ProviderRepository`.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
```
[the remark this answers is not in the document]
jonathannorris: `javax.annotation.Nullable` is already available at compile time via the existing `spotbugs-annotations` → `jsr305` (provided) dependency chain. CI passes on Java 11/17 without adding a new direct dependency.

jonathannorris:
Addressing the remaining CodeRabbit nitpicks from the review body:

- **`@AfterEach` teardown** (`DomainScopedProviderSpecTest`): fixed in 68fa5da.
- **DOMAIN_B `initialize` assertion**: skipping. A shared non-domain-scoped provider reuses the same state manager on the second bind; `initialize` runs once on first bind only (same behavior as js-sdk).
- **`singleArgOverrideIsOnlyInvokedOnce` rename**: skipping. The `@DisplayName` already clarifies intent; the method name is slightly misleading but low risk.
- **`InitializeBehaviorSpecTest` dedup**: skipping. The duplicate bodies are intentional so each `@Specification` requirement has its own test for traceability.

All three inline review threads (Copilot identity check, MultiProvider PR description, `javax.annotation.Nullable`) are resolved.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
```diff
@@ -41,6 +42,38 @@ default void initialize(EvaluationContext evaluationContext) throws Exception {
...
+     * <p>
+     * The default provider is initialized with a {@code null} domain. Providers that
+     * maintain per-domain state (for example a persistent cache) should override this
+     * method and declare themselves {@linkplain #isDomainScoped() domain-scoped}.
+     * </p>
+     *
+     * @param evaluationContext the global evaluation context
+     * @param domain            the bound domain, or {@code null} for the default provider
+     */
+    default void initialize(EvaluationContext evaluationContext, @Nullable String domain) throws Exception {
```
toddbaert: Same as JS, we could consider using an object here to future-proof (something like `InitializationContext`, comparable to `HookContext`). IMO not a blocker but worth thinking about before we commit to the two-arg overload.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
```diff
@@ -185,37 +187,82 @@ private void prepareAndInitializeProvider(
...
         } else {
             taskExecutor.submit(() -> {
                 // initialization happens in a different thread if we're not waiting for it
-                initializeProvider(newStateManager, afterInit, afterShutdown, afterError, oldStateManager);
+                initializeProvider(domain, newStateManager, afterInit, afterShutdown, afterError, oldStateManager);
             });
         }
     }
 
+    private void validateDomainScopedBinding(@Nullable String domain, FeatureProvider newProvider) {
```
toddbaert: Again, same as JS, might be better to extract this and implement it in one check. Something like a single upfront predicate that inspects `defaultStateManger` and `stateManagers` for this exact instance and rejects if any existing binding differs from the incoming one. More lines but keeps the rule in one place. Totally optional.

src/test/java/dev/openfeature/sdk/DomainScopedProviderSpecTest.java
```diff
@@ -0,0 +1,250 @@
...
+    @DisplayName("rejects binding a domain-scoped provider to a second named domain")
+    void rejectsBindingDomainScopedProviderToSecondNamedDomain() {
+        DomainScopedTestProvider provider = new DomainScopedTestProvider();
+
+        api.setProvider(DOMAIN_A, provider);
+
+        assertThatThrownBy(() -> api.setProvider(DOMAIN_B, provider))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Domain-scoped provider cannot be bound to more than one domain");
+    }
```
toddbaert, line 55: The three rejection tests (`rejectsBindingDomainScopedProviderToSecondNamedDomain`, `rejectsBindingDomainScopedNamedProviderAsDefault`, `rejectsBindingDomainScopedDefaultProviderToNamedDomain`) only verify the exception is thrown. IMO we should also assert `initialize` is not re-run after the rejection.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
toddbaert: We removed `javax.annotation.*` in main sources a while back (see #921), and there are no such imports anywhere in `src/main/java/` on `main` today (there's one in `/test` I think we should even remove). The runtime generally ignores the annotation if it's not available on classpath, but there are still ways to get `ClassNotFound`s if no jsr305 impl is found.

jsr305 is abandoned, so I'd rather not have `FeatureProvider` reference it at all. The Javadoc at line 56 already documents that `domain` can be null, which is what we do elsewhere in `src/main`.

I'd suggest dropping the annotation here (and the internal ones in `ProviderRepository` / `FeatureProviderStateManager` / `MultiProvider` as well).

src/main/java/dev/openfeature/sdk/multiprovider/MultiProvider.java
```diff
@@ -84,6 +85,19 @@ protected static Map<String, FeatureProvider> buildProviders(List<FeatureProvide
      */
     @Override
     public void initialize(EvaluationContext evaluationContext) throws Exception {
+        initialize(evaluationContext, null);
+    }
+
+    /**
+     * Initialize the provider with the bound domain, if any.
```
chrfwow: These docs belong in the interface

src/main/java/dev/openfeature/sdk/FeatureProvider.java
chrfwow: Are we sure that the addition of this default method is not a breaking change on a byte code level?

src/main/java/dev/openfeature/sdk/ProviderRepository.java
chrfwow: Is the functionality of this method already covered by a VmLens test? If not, we should add one to guard against registering two providers to the same domain concurrently

src/main/java/dev/openfeature/sdk/FeatureProvider.java
toddbaert: Good point.

toddbaert: Rebased against our recent additions for byecode/sourcecode change detection, and passes, so I guess no problem with this.

src/main/java/dev/openfeature/sdk/ProviderRepository.java
toddbaert: Added `concurrentRegistrationToSameDomain_exactlyOneBoundAndLoserShutDown` - please take a look - I think it covers this.

src/main/java/dev/openfeature/sdk/FeatureProvider.java
toddbaert: I removed these annotations.

chrfwow (approved):

toddbaert (approved):

PR #2005, opened 2026-08-06
feat(multiprovider): execute child provider hooks during evaluation

jonathannorris:
## Summary

- Hooks registered on a child provider now run around that child's evaluation, so a wrapped provider's own hooks observe the evaluations they take part in.
- Adds `MultiProviderHookExecutor`, which runs the before/after/error/finally stages per child, mirroring the JS SDK's `HookExecutor`.
- `before` hooks run in registration order; `after`, `error`, and `finallyAfter` run in reverse, per spec.
- `MultiProvider` exposes a provider-level hook that captures the `ClientMetadata` and hook hints from the SDK lifecycle so they can be passed into child hook contexts. This mirrors the JS SDK's `WeakMap` approach; here it's a `ThreadLocal` snapshot captured before the strategy runs, so parallel strategies work correctly.

## Related PRs

This is one of three independent PRs that together close the multi-provider gaps identified in #1882. They branch off `main` separately and can be reviewed and merged in any order. Together they replace #1897.

- #2003 `ComparisonStrategy`
- #2004 child provider event aggregation, state, and tracking
- #2005 child provider hook execution (this PR)

Relates to #1882

### DIFF
--- a/src/main/java/dev/openfeature/sdk/multiprovider/HookExecutionContext.java
+++ b/src/main/java/dev/openfeature/sdk/multiprovider/HookExecutionContext.java
@@ -1,0 +1,15 @@
+package dev.openfeature.sdk.multiprovider;
+
+import dev.openfeature.sdk.ClientMetadata;
+import java.util.Map;
+
+/** Captures hook lifecycle context (client metadata and hints) for per-provider hook execution. */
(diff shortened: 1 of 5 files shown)


toddbaert:
I'll take this over.

